PR #1, opened 2017-04-13
Initial 128-bit SIMD proposal

stoklund:
[Rendered](https://github.com/stoklund/wasm-simd/blob/proposal/proposals/simd/Overview.md).


proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1324 @@
...
+
+Floating-point operations that need a *rounding mode* take a `RoundingMode`
+operand which provides the rounding mode to use for the operation.
+
+* `TiesToEven`: Round to nearest, ties towards even.
+* `TowardPositive`: Round towards positive infinity.
+* `TowardNegative`: Round towards negative infinity.
+* `TowardZero`: Round towards zero.
+
+## Default NaN value
```
jfbastien: To confirm: the intent here is to precisely match [the latitude which WebAssembly provides](https://github.com/WebAssembly/design/blob/master/Semantics.md#floating-point-operators) w.r.t. NaNs, correct? I find it odd to repeat it here, with different wording.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1324 @@
...
+```
+
+## Subnormal flushing
+
+An implementation is allowed to flush subnormals in arithmetic floating-point
+operations. This means that any subnormal operand is treated as 0, and any
+subnormal result is rounded to 0.
+
+Note that this differs from WebAssembly scalar floating-point semantics which
+require correct subnormal handling.
```
jfbastien: I would rather not allow this here, and propose it separately from this PR.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1324 @@
...
+* `i8x16.equal(a: v128, b: v128) -> b8x16`
+* `i16x8.equal(a: v128, b: v128) -> b16x8`
+* `i32x4.equal(a: v128, b: v128) -> b32x4`
+* `i64x2.equal(a: v128, b: v128) -> b64x2`
+* `f32x4.equal(a: v128, b: v128) -> b32x4`
+* `f64x2.equal(a: v128, b: v128) -> b64x2`
+
+Integer equality is independent of the signed/unsigned interpretation. Floating
+point equality follows IEEE semantics, so a NaN lane compares not equal with
+anything, including itself:
```
jfbastien: If mentioning NaN, it's worth also mentioning `-0.0` and `0.0`.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1324 @@
...
+vectors.
+
+The memory operations work on an abstract `Buffer` instance which can be
+addressed by a `ByteOffset` type. Unaligned memory operations are allowed, but
+they may be slower than aligned operations.
+
+This specification does not address bounds checking and trap handling for memory
+operations. It is assumed that the range `addr .. addr+15` are valid offsets in
+the buffer, and that computing `addr+15` does not overflow the `ByteOffset`
+type. Bounds checking should be handled by the embedding specification.
```
jfbastien: WebAssembly simply states: "Out of bounds accesses trap. If the access is a store, if any of the accessed bytes are out of bounds, none of the bytes are modified."

We should do the same here.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1324 @@
...
+        if x == 0 and y == 0 and signbit(x) != signbit(y):
+            return +0.0
+        if x > y:
+            return x
+        else:
+            return y
+    return S.lanewise_binary(max, a, b)
+```
+
+### NaN-suppressing minimum
```
jfbastien: I would rather introduce `minNum` and `maxNum` for scalar at the same time or earlier than for SIMD.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1324 @@
...
+### Floating point to integer
+* `s32x4.fromFloat(a: v128) -> (result: v128, fail: boolean)`
+* `s64x2.fromFloat(a: v128) -> (result: v128, fail: boolean)`
+* `u32x4.fromFloat(a: v128) -> (result: v128, fail: boolean)`
+* `u64x2.fromFloat(a: v128) -> (result: v128, fail: boolean)`
+
+Lane-wise conversion from floating point to integer using the IEEE
+`convertToIntegerTowardZero` function. If any lane is a NaN or the rounded
+integer value is outside the range of the destination type, return `fail = true`
+and an unspecified `result`.
```
jfbastien: WebAssembly doesn't currently support multi-return values. This should trap to match scalars.

Also, rename to truncate so it matches scalar.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1324 @@
...
+    result = S.New()
+    for i in range(S.Lanes):
+        if s[i] < S.lanes:
+            result[i] = a[s[i]]
+        else:
+            result[i] = b[s[i] - S.lanes]
+    return result
+```
+
+## Integer arithmetic
```
jfbastien: This section is missing integer division.

proposals/simd/Overview.md
```diff
@@ -1 +1,127 @@
...
+* `boolean`: SIMD operations with a boolean argument will accept a WebAssembly
+  `i32` instead and treat zero as false and non-zero values as true. SIMD
+  operations that return a boolean will return an `i32` with the value 0 or 1.
+
+* `LaneIdx2` through `LaneIdx32`: All lane indexes are encoded as `varuint7`
+  immediate operands. Dynamic lane indexes are not used anywhere. An
+  out-of-range lane index is a validation error.
+
+* `RoundingMode`: Rounding modes are encoded as `varuint7` immediate operands.
+  An out-of-range rounding mode is a validation error.
```
stoklund: I don't think we need rounding mode support in the first version. I was anticipating their inclusion in the scalar spec.

We could represent this as an immediate field that must be `RoundTiesToEven` in the first version, or we can add duplicate opcodes with rounding mode support in the future. It's only 14 opcodes that take a rounding mode, so I guess duplicating them is not too bad.

proposals/simd/webassembly-opcodes.md
```diff
@@ -0,0 +1,279 @@
...
+| `i8x16.replaceLane(a: v128, i: LaneIdx16, x: i32) -> v128` | [i8x16.replaceLane](portable-simd.md#replace-lane-value) |
+| `i8x16.add(a: v128, b: v128) -> v128`                      | [i8x16.add](portable-simd.md#integer-addition) |
+| `i8x16.sub(a: v128, b: v128) -> v128`                      | [i8x16.sub](portable-simd.md#integer-subtraction) |
+| `i8x16.mul(a: v128, b: v128) -> v128`                      | [i8x16.mul](portable-simd.md#integer-multiplication) |
+| `i8x16.neg(a: v128) -> v128`                               | [i8x16.neg](portable-simd.md#integer-negation) |
+| `i8x16.addSaturate_s(a: v128, b: v128) -> v128`            | [s8x16.addSaturate](portable-simd.md#saturating-integer-addition) |
+| `i8x16.addSaturate_u(a: v128, b: v128) -> v128`            | [u8x16.addSaturate](portable-simd.md#saturating-integer-addition) |
+| `i8x16.subSaturate_s(a: v128, b: v128) -> v128`            | [s8x16.subSaturate](portable-simd.md#saturating-integer-subtraction) |
+| `i8x16.subSaturate_u(a: v128, b: v128) -> v128`            | [u8x16.subSaturate](portable-simd.md#saturating-integer-subtraction) |
```
stoklund: These are very useful operations both for fixed-point DSP and image processing. They all exist as individual SIMD instructions in ARM, Intel, and MIPS ISAs.

These instructions are not commonly available in scalar versions in ISAs. In scalar code you're usually computing with 32-bit registers anyway, so saturating results to 8 or 16 bits is easier. In SIMD code, you only get 8 or 16 bits per lane, so these instructions are more important there.

LLVM's auto-vectorizer can pattern-match these operations in scalar code and generate the vector instructions.

proposals/simd/portable-simd.md
stoklund: Yes, that is correct.

proposals/simd/portable-simd.md
stoklund: That's an intentional omission. As it turns out, no ISAs provide vectorized integer division instructions.

proposals/simd/portable-simd.md
stoklund: I agree, and you won't find these instruction in the table of proposed opcodes. I'll make a note that I omitted them.

proposals/simd/portable-simd.md
stoklund: Agreed. I did just that in the *Floating point conversions* section of the overview document.

proposals/simd/portable-simd.md
stoklund: Filed #2.

proposals/simd/webassembly-opcodes.md
```diff
@@ -0,0 +1,279 @@
...
+| `f32x4.gt(a: v128, b: v128) -> b32x4`                         | [f32x4.greaterThan](portable-simd.md#greater-than) |
+| `f32x4.ge(a: v128, b: v128) -> b32x4`                         | [f32x4.greaterThanOrEqual](portable-simd.md#greater-than-or-equal) |
+| `f32x4.abs(a: v128) -> v128`                                  | [f32x4.abs](portable-simd.md#absolute-value) |
+| `f32x4.min(a: v128, b: v128) -> v128`                         | [f32x4.min](portable-simd.md#nan-propagating-minimum) |
+| `f32x4.max(a: v128, b: v128) -> v128`                         | [f32x4.max](portable-simd.md#nan-propagating-maximum) |
+| `f32x4.div(a: v128, b: v128, rmode: RoundingMode) -> v128`    | [f32x4.div](portable-simd.md#division) |
+| `f32x4.sqrt(a: v128, rmode: RoundingMode) -> v128`            | [f32x4.sqrt](portable-simd.md#square-root) |
+| `f32x4.reciprocalApproximation(a: v128) -> v128`              | [f32x4.reciprocalApproximation](portable-simd.md#reciprocal-approximation) |
+| `f32x4.reciprocalSqrtApproximation(a: v128) -> v128`          | [f32x4.reciprocalSqrtApproximation](portable-simd.md#reciprocal-square-root-approximation) |
```
stoklund: Filed #3.

proposals/simd/Overview.md
jfbastien: I would suggest starting this PR without rounding at all, mimic scalar, and filing a separate issue proposing how to encode it for both SIMD and scalar.

proposals/simd/webassembly-opcodes.md
jfbastien: I agree saturation is useful. I'm asking if the first version needs this though. Not necessarily the first ship of SIMD, simply the first proposal.

I ask because by adding saturation here you're making the portable performance case that needs to be built harder, in my opinion.

proposals/simd/portable-simd.md
jfbastien: Ah apologies, I though I'd seen it in one of the files in your PR, and it was missing here. This is good then!

proposals/simd/webassembly-opcodes.md
dtig: Could you elaborate on what exactly makes portable performance harder here? I would argue that as all the saturate operations can be directly mapped to individual instructions on Arm, Intel and MIPS ISAs, this is an easier performance win than some of the other operations.

jfbastien: You'll need to justify this SIMD proposal on the merits of one more kind of benchmarks, across all those ISAs. SIMD can stand on its own with just a few FP and int benchmarks, but if you want to throw in saturation have at it. I'm just stating where I think the bar is.

proposals/simd/Overview.md
```diff
@@ -1 +1,127 @@
...
+1. [A specification of portable SIMD operations](portable-simd.md) that came
+   out of the SIMD.js work.
+2. [A table of proposed WebAssembly operations](webassembly-opcodes.md) with
+   links to the portable specification.
+3. This document which describes the mapping between WebAssembly and the
+   portable specification.
+
+Referencing a common specification of the portable SIMD semantics reduces the
+work required to support both SIMD.js and WebAssembly SIMD in the same
+implementation.
```
arunetm: I agree that this is not too relevant to be included as a goal.

proposals/simd/Overview.md
```diff
@@ -1 +1,127 @@
...
+* `b32x4`: A vector of 4 boolean lanes.
+* `b64x2`: A vector of 2 boolean lanes.
+
+The 128 bits in a `v128` value are interpreted differently by different
+operations. They can represent vectors of integers or IEEE floating point
+numbers.
+
+The four boolean vector types do not have a prescribed representation in
+memory; they can't be loaded or stored. This allows implementations to choose
+the most efficient representation, whether as a predicate vector or some
```
arunetm: Can you help to understand why the bool  vector load/stores disallowed?

rossberg-chromium:
Just a quick question: @jfbastien recently suggested distinguishing an int and a float SIMD type. Do you guys still intend to explore that option in follow-ups?

jfbastien:
> Just a quick question: @jfbastien recently suggested distinguishing an int and a float SIMD type. Do you guys still intend to explore that option in follow-ups?

I'd be interested in seeing if it makes a difference for producers or consumers to do this. I'm not convinced it's the right thing! @pizlonator thinks it'll cause way too many coercions (bloating the binary) and won't be helpful to the consumer anyways because the consumer already needs to reason about the SIMD types.

My thinking was that having a compiler like LLVM express things like lane crossing would mean browsers can be lazy about it and just trust the input. No need to think about PD versus PS mov / bitwise ops, or shuffle type, just trust the input blindly.

stoklund:
@rossberg-chromium, in my opinion we don't need separate types, but possibly a few extra instructions.

To give some background, Intel's SIMD instruction sets have multiple versions of a few instructions. For example, `pxor`, `xorps`, and `xorpd` are architecturally identical implementations of `v128.xor`, but current micro-architectures will issue `pxor` to the integer execution stack and the two others to the floating point stack. There has never been a micro-architectural difference between the float-flavored `xorps` and the double-flavored `xorpd` as far as I know.

There is a 1-cycle bypass delay when an instruction in the integer stack depends on a result computed in the floating point stack or vice versa. This additional latency only matters if the dependency is on the critical path. As soon as there's a few cycles between the instructions, operands are read from the register file and not from pipeline bypasses. Then there is no longer any difference.

Other ISAs don't make this distinction, it is an Intel-only thing.

If we want to let WebAssembly producers to give hints to the code generator's fiddling with these micro-architectural details, we could do it by providing float-flavored versions of the logical, load/store, and shuffle/swizzle operations. These new operations would be identical to the existing ones except for giving this hint to the code generator. I don't think a new type is required.

I would prefer that we don't do this in the initial proposal, and only add the new instructions if they have demonstrable performance benefits. I am expecting improvements to real-world benchmarks to be in the noise.

Without the hints, a simple SSE code generator would apply a basic heuristic like looking at the instructions that computed the inputs and choose `xorps` if they are floating-point instructions, `pxor` otherwise. LLVM's algorithm is a bit more involved, but a little goes a long way.

stoklund:
> My thinking was that having a compiler like LLVM express things like lane crossing would mean browsers can be lazy about it and just trust the input. No need to think about PD versus PS mov / bitwise ops, or shuffle type, just trust the input blindly.

Annoyingly, the Intel shuffle instructions don’t have the simple symmetry of the logical and load/store variants. For example, the integer-flavored `pshufd` implements `v32x4.swizzle` but doesn’t have a floating-point counterpart. The float-flavored `shufps` implements some of the `v32x4.shuffle` patterns but doesn’t have an integer counterpart.

The specific SSE version you’re targeting can also affect the results. With SSE 4.1 support, `i32x4.mul` becomes a `pmulld` (int-like) instruction, but on earlier SSE versions it expands into an instruction sequence that ends in `shufps` (float-like), at least in the current Firefox implementation.

AndrewScheidecker:
There are a few operators remaining that use mixed case to denote word boundaries (`addSaturate`, `subSaturate`, `extractLanes`, and `replaceLanes`). IMO those should use _ to match the rest of the wasm names.

Will globals of type `v128` and `b*` be allowed? If so, those types should have a const operator that can be used as an initializer expression.

What is the natural alignment of the `v128.load(1|2|3)` operators?

Can we get vector versions of the float `ceil`, `floor`, `trunc`, `nearest` operators?

How about the integer `rotl`, `rotr`, `clz`, and `ctz` operators? Don't know how good the hardware support is for those, though.

stoklund:
Thanks, @AndrewScheidecker 

I don't have strong opinions on the operator names. I translated names for the operations with existing WebAssembly counterparts and left the rest identical to the SIMD.js spec. I imagined we would have a bikeshedding fest after we've decided which operations go in. Plain changing to snake case makes sense.

> Will globals of type v128 and b* be allowed? If so, those types should have a const operator that can be used as an initializer expression.

I don't think there's a reason to disallow SIMD globals, so we should include them for completeness. I could imagine a use for global constants at least.

> What is the natural alignment of the v128.load(1|2|3) operators?

I think 32 bits makes sense as the natural alignment for these operators.

> Can we get vector versions of the float ceil, floor, trunc, nearest operators?

I am not sure why they didn't make it into the SIMD.js spec. Let me check if they are universally available.

> How about the integer rotl, rotr, clz, and ctz operators? Don't know how good the hardware support is for those, though.

Rotates, probably. I'm not sure if `clz` and `ctz` have wide hardware support, and they could be expensive to emulate, particularly the 16-bit and 8-bit variants. I'll look into it.

proposals/simd/webassembly-opcodes.md
```diff
@@ -0,0 +1,275 @@
...
+| `i32x4.gt_u(a: v128, b: v128) -> b32x4` | [u32x4.greaterThan](portable-simd.md#greater-than) |
+| `i32x4.ge_s(a: v128, b: v128) -> b32x4` | [s32x4.greaterThanOrEqual](portable-simd.md#greater-than-or-equal) |
+| `i32x4.ge_u(a: v128, b: v128) -> b32x4` | [u32x4.greaterThanOrEqual](portable-simd.md#greater-than-or-equal) |
+| `i32x4.trunc_s/f32x4(a: v128) -> v128` | [s32x4.fromFloat](portable-simd.md#floating-point-to-integer) |
+| `i32x4.trunc_u/f32x4(a: v128) -> v128` | [u32x4.fromFloat](portable-simd.md#floating-point-to-integer) |
+
+## `f32x4` operations
+| WebAssembly | Portable SIMD |
+|:------------|:--------------|
+| `f32x4.build(x: f32[4]) -> v128` | [f32x4.build](portable-simd.md#build-vector-from-individual-lanes) |
```
sunfishcode: I assume there should also be a `v128.const` instruction somewhere​ which takes a 128-bit immediate.

Then, it's worth asking whether these `build` opcodes are needed in wasm. LLVM doesn't have corresponding operators, for example, and they don't tend to map to single instructions in implementations.

proposals/simd/webassembly-opcodes.md
```diff
@@ -0,0 +1,275 @@
...
+| `b8x16.or(a: b8x16, b: b8x16) -> b8x16` | [b8x16.or](portable-simd.md#logical-or) |
+| `b8x16.xor(a: b8x16, b: b8x16) -> b8x16` | [b8x16.xor](portable-simd.md#logical-xor) |
+| `b8x16.not(a: b8x16) -> b8x16` | [b8x16.not](portable-simd.md#logical-not) |
+| `b8x16.anyTrue(a: b8x16) -> i32` | [b8x16.anyTrue](portable-simd.md#any-lane-true) |
+| `b8x16.allTrue(a: b8x16) -> i32` | [b8x16.allTrue](portable-simd.md#all-lanes-true) |
+
+## `b16x8` operations
+| WebAssembly | Portable SIMD |
+|:------------|:--------------|
+| `b16x8.build(x: i32[8]) -> b16x8` | [b16x8.build](portable-simd.md#build-vector-from-individual-lanes) |
```
sunfishcode: Does a `b16x8.const` also make sense?

proposals/simd/webassembly-opcodes.md
```diff
@@ -0,0 +1,275 @@
...
+
+## `v32x4` operations
+| WebAssembly | Portable SIMD |
+|:------------|:--------------|
+| `v32x4.select(s: b32x4, t: v128, f: v128) -> v128` | [v32x4.select](portable-simd.md#lane-wise-select) |
+| `v32x4.swizzle(a: v128, s: LaneIdx4[4]) -> v128` | [v32x4.swizzle](portable-simd.md#swizzle-lanes) |
+| `v32x4.shuffle(a: v128, b: v128, s: LaneIdx8[4]) -> v128` | [v32x4.shuffle](portable-simd.md#shuffle-lanes) |
+| `v32x4.load1(addr, offset) -> v128` | [v32x4.load1](portable-simd.md#partial-load) |
+| `v32x4.load2(addr, offset) -> v128` | [v32x4.load2](portable-simd.md#partial-load) |
+| `v32x4.load3(addr, offset) -> v128` | [v32x4.load3](portable-simd.md#partial-load) |
```
sunfishcode: Partial loads and stores were [removed](https://github.com/kripken/emscripten/issues/4645#issuecomment-256101137) from asm.js since the non-power-of-2 size is awkward in some implementations.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1320 @@
...
+    assert isinan(x) or isnan(y)
+    if not isnan(x):
+        return canonicalize_nan(y)
+    if not isnan(y)
+        return canonicalize_nan(x)
+    # Both x and y are NaNs: pick one to propagate.
+    if unspecified_choice():
+        return canonicalize_nan(x)
+    else:
+        return canonicalize_nan(y)
```
stoklund: That's what you get for specifying things twice. I think the next step for this proposal is to roll up this portable spec and the wasm specifics into a single wasm-only document. Then I can simply refer to existing semantics instead of trying to do a stand-alone spec.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1320 @@
...
+Return a new vector with lanes identical to `a`, except for lane `i` which has
+the value `x`.
+
+```python
+def S.replaceLane(a, i, x):
+    result = S.New()
+    for j in range(S.Lanes):
+        result[j] = a[j]
+    result[i] = x
+    return result
```
stoklund: My intention was for these indices to be immediate operands. Do you think we need dynamic lane indices for these operations? It's hard to generate good code for that.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1320 @@
...
+```python
+def S.shuffle(a, b, s):
+    result = S.New()
+    for i in range(S.Lanes):
+        if s[i] < S.lanes:
+            result[i] = a[s[i]]
+        else:
+            result[i] = b[s[i] - S.lanes]
+    return result
+```
```
stoklund: I agree. Lane indices are supposed to be immediate operands everywhere.

proposals/simd/webassembly-opcodes.md
stoklund: We talked about this for asm.js too and ended up keeping them. I'll file an issue.

proposals/simd/webassembly-opcodes.md
stoklund: Yes, I think we should add const instructions for all the new types.

proposals/simd/portable-simd.md
jfbastien: I would rather have immediate operands, unless we can show performance advantages in dynamic ones.

proposals/simd/portable-simd.md
```diff
@@ -0,0 +1,1320 @@
...
+            return -0.0
+        if x < y:
+            return x
+        else:
+            return y
+    return S.lanewise_binary(minNum, a, b)
+```
+
+Note that this function behaves differently than the IEEE 754 `minNum` function
+when one of the operands is a signaling NaN.
```
arunetm: Seems like it was agreed upon in Simd.js to remove minNum and maxNum due to the compliance issue and not enough use cases. https://github.com/tc39/ecmascript_simd/issues/341
Does it make sense to remove it from portable_simd as well?
And the replacement from IEEE754-2018 can be introduced to wasm first if it seems fit when it lands.

proposals/simd/webassembly-opcodes.md
arunetm: Does bools have uses beyond holding/manipulating results of logical and relational operations?
Wondering if there is a need to have support for building bools, since we intend to have bool consts.

AndrewScheidecker:
FYI I have a prototype for this proposal in [WAVM](https://github.com/AndrewScheidecker/WAVM), including a [port of the Blake2b hash function](https://github.com/AndrewScheidecker/WAVM/blob/master/Test/Blake2b/blake2b.wast).

stoklund:
@jfbastien, would you object to my merging this PR in its current state? I think I've addressed everything brought up in reviews either by changing the text or by filing issues for individual topics that need more discussion.

jfbastien:
Yeah I think as is you've got a good starting point, received good feedback and definite interest from multiple folks.

Issue #33, opened 2018-08-07
Are shuffle's lane indices dynamic? 


gnzlbg:
Reading the spec:

> Shuffle lanes
> 
> `v8x16.shuffle(a: v128, b: v128, s: LaneIdx32[16]) -> v128`

it appears that the lane indices argument for the `shuffle` intrinsic is a non-immediate mode argument.

If my memory doesn't fail me: 

* LLVM's `shufflevector` requires compile-time indices. 

* Most architectures do not support shuffling vector lanes of arbitrary width with dynamic indices:
   * x86/x86_64 SSSE3's pshufb (and AVX-512 variants) support shuffling _bytes_ with non-immediate mode indices
   * ARM 32/64 vtbl/vtbx family of instructions support shuffling _bytes_ with non-immediate mode indices

That is, if the user is shuffling a `v8x16`, we can use those instructions directly to shuffle with dynamic indices without issues.

If the user attempts to shuffle a `v16x8`, `v32x4`, etc. what's the best machine code that we can generate for dynamic indices?

If I remember correctly, generating the appropriate sequence of shuffle bytes instructions requires knowing the lane indices during machine code generation. If these aren't known, the only thing a code generator could do AFAICT is fall back to scalar code, that is, copy the vectors to memory, do a scalar shuffle there e.g. by copying into a third vector, and copy the result of this third vector back to a vector register. This sounds like a huge performance cliff to me.

What am I missing? How can we generate efficient vector shuffles from dynamic lane indices?

gnzlbg:
Duh...

What I was missing is that `LaneIdx32` is an immediate mode operand. 

Honestly, I think the spec should be a bit more idiot proof here, by:

* renaming `LaneIdx32` to `ImmLaneIdx32` (for consistency with `ImmByte`) 
* renaming the `s` argument in `v128.shuffle` to `sImm`, `imm`, or similar

gnzlbg:
So it appears that `LaneIdx` are not always immediate mode operands, which would mean that my original interpretation of the spec is correct, and that the vector shuffles accept dynamic run-time indices. 

So I am re-opening this for clarification.

gnzlbg:
@billbudge how does V8 implement these?

gnzlbg:
Duh again. So yeah, the lane indices are dynamic, but that's ok because the `shuffle` only operates on `v8x16` and one can generate efficient code for that on pretty much every architecture.

billbudge:
[...]

PR #34, opened 2018-08-07
Rename LaneIdx{2,4,8,16,32} to ImmLaneIdx{2,4,8,16,32}

gnzlbg:
This makes it clear everywhere that `ImmLaneIdx` is an immediate operand and also makes the nomenclature for immediate operands consistent (both `ImmByte` and `ImmLaneIdx`  start with the prefix `Imm`) .

See https://github.com/WebAssembly/simd/issues/33 .

### DIFF
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -48,7 +48,7 @@ For example, the instructions `f32x4.extract_lane` and `i64x2.extract_lane`
 perform the same semantic operation: extracting the scalar value of a vector
 lane. However, the `f32x4.extract_lane` instruction returns a 32-bit wide
 floating point value, while the `i64x2.extract_lane` instruction returns a
-64-bit wide floating point value.
+64-bit wide integer value.
 
 The interpretations form a hierarchy with successive refinements, where `v128`
 is at the top of the hierarchy and interprets the `v128` vector type as a bag of
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -50,12 +50,10 @@ lane. However, the `f32x4.extract_lane` instruction returns a 32-bit wide
 floating point value, while the `i64x2.extract_lane` instruction returns a
 64-bit wide integer value.
 
-The interpretations form a hierarchy with successive refinements, where `v128`
-is at the top of the hierarchy and interprets the `v128` vector type as a bag of
-bits. The `v32x4` refines `v128` (`v32x4 : v128`) and interprets the vector as 4
-32-bit wide lane, while the `i32x4` and `f32x4` both refine `v32x4` and
-interpret the content of the vector lanes as being of integer or floating-point
-types.
+The `v128` vector type interpretation interprets the vector as a bag of bits. 
+The `v{lane_width}x{n}` interpretations (e.g. `v32x4`) interpret the vector as
+`n` lanes of `lane_width` bits. The `{t}{lane_width}x{n}` interpretations (e.g.
+`i32x4` or `f32x4`) interpret the vector as `n` lanes of type `{t}{lane_width}`.
 
 ### Lane division interpretation
 
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -13,12 +13,13 @@ of immediate operands used by the SIMD instructions.
 
 ## SIMD value type
 
-The `v128` value type has a concrete mapping to a 128-bit representation with bits
-numbered 0–127. The `v128` type corresponds to a vector register in a typical
-SIMD ISA. The interpretation of the 128 bits in the vector register is provided
-by the individual instructions. When a `v128` value is represented as 16 bytes,
-bits 0-7 go in the first byte with bit 0 as the LSB, bits 8-15 go in the second
-byte, etc.
+The `v128` value type is the _only_ type introduced in this extension. It has a
+concrete mapping to a 128-bit representation with bits numbered 0–127. The
+`v128` type corresponds to a vector register in a typical SIMD ISA. The
+interpretation of the 128 bits in the vector register is provided by the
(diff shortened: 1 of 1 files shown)


gnzlbg:
So now I think this PR is incorrect, since some instructions like `extract_lanes` do not appear to require immediate mode operands (or at least the spec doesn't say so). I am re-opening the issue #33 .

gnzlbg:
So looking at the binary encoding it appears that the extract_lanes / replace_lanes / ... all take immediate mode operands... so I think this PR is correct, but incomplete, in that the nomenclature should be made consistent for those other operations as well.

gnzlbg:
So I've pushed another commit with new changes that:

* consistently prefix immediate mode arguments with `imm`
* consistently describes how these immediate mode arguments are encoded in the binary
* contains a text-form explanation of the semantics of the `shuffle` instruction indices
* contains further explanations and examples of what the interpretation and the hierarchy are since they are only a naming scheme for instructions.

gnzlbg:
@penzn 

> I don't know why the wording in the spec is the way it is, but you can definitely try submitting a PR to correct it :)

I've got your back :P

penzn:
> @penzn
> 
> > I don't know why the wording in the spec is the way it is, but you can definitely try submitting a PR to correct it :)
> 
> I've got your back :P

I see, I should've noticed that you already opened it :)

proposals/simd/SIMD.md
```diff
@@ -27,17 +28,34 @@ encoded as individual bytes in the binary encoding. Many have a limited valid
...
+
+The instructions introduced in this specification are named according to the
+following schema: `{interpretation}.{operation}`. Where the `{interpretation}`
+prefix denotes how the bytes of the `v128` type are interpreted by the `{operation}`. 
+
+For example, the instructions `f32x4.extract_lane` and `i64x2.extract_lane`
+perform the same semantic operation: extracting the scalar value of a vector
+lane. However, the `f32x4.extract_lane` instruction returns a 32-bit wide
+floating point value, while the `i64x2.extract_lane` instruction returns a
+64-bit wide floating point value.
```
dtig: This should be 64-bit wide integer value.

proposals/simd/SIMD.md
```diff
@@ -27,17 +28,34 @@ encoded as individual bytes in the binary encoding. Many have a limited valid
...
+
+The interpretations form a hierarchy with successive refinements, where `v128`
+is at the top of the hierarchy and interprets the `v128` vector type as a bag of
+bits. The `v32x4` refines `v128` (`v32x4 : v128`) and interprets the vector as 4
+32-bit wide lane, while the `i32x4` and `f32x4` both refine `v32x4` and
+interpret the content of the vector lanes as being of integer or floating-point
+types.
```
dtig: I'm wondering if hierarchy is the right word here? To me, these are just different interpretations of v128 values.

proposals/simd/SIMD.md
```diff
@@ -188,15 +208,17 @@ The `_s` and `_u` variants will sign-extend or zero-extend the lane value to
...
+* `i16x8.replace_lane(a: v128, imm: ImmLaneIdx8, x: i32) -> v128`
+* `i32x4.replace_lane(a: v128, imm: ImmLaneIdx4, x: i32) -> v128`
+* `i64x2.replace_lane(a: v128, imm: ImmLaneIdx2, x: i64) -> v128`
+* `f32x4.replace_lane(a: v128, imm: ImmLaneIdx4, x: f32) -> v128`
+* `f64x2.replace_lane(a: v128, imm: ImmLaneIdx2, x: f64) -> v128`
+
+Return a new vector with lanes identical to `a`, except for the lane specified
+in the immediate mode operand `imm` which has the value `x`. The
+`{interpretation}.replace_lane` instructions are encoded with one immediate byte
+providing the index of the lane to whose value is to be replaced.
```
dtig: Language nit: Would it be okay to change the last bit of the sentence to  "instructions are encoded with an immediate byte providing the index of the lane the value of which is to be replaced."?

proposals/simd/SIMD.md
```diff
@@ -557,18 +583,20 @@ def S.ne(a, b):
...
-memory access size is 16 bytes which is also the natural alignment.
+scalar WebAssembly load and store instructions (see
+[memarg](https://webassembly.github.io/spec/core/bikeshed/index.html#syntax-memarg).
+The difference is that the memory access size is 16 bytes which is also the
+natural alignment.
 
 ### Load
 
-* `v128.load(memarg) -> v128`
+* `v128.load(imm: memarg) -> v128`
```
dtig: The memarg is specified in the syntax so the "imm:" here is inconsistent with how instructions in the other proposals are specified. Same comment for the store operation below.

proposals/simd/SIMD.md
gnzlbg: It did feel off to me as well, but that's the word the text was using :/ 

I would just write something like:

> The `v128` vector type interpretation interprets the vector as a bag of bits. The `v{lane_width}x{n}` interpretation (e.g. `v32x4`) interpret the vector as `n` lanes of `lane_width` bits. The `{t}{lane_width}x{n}` interpretations (e.g. `i32x4` or `f32x4`) interpret the vector as `n` lanes of type `{t}{lane_width}`.

proposals/simd/SIMD.md
gnzlbg: Sure, thank you!

proposals/simd/SIMD.md
dtig: Thanks for clarifying, using the text in your comment sgtm.

gnzlbg:
Thank you for your feedback @dtig , I hope I incorporated all of it correctly into the PR, let me know what you think!

PR #39, opened 2018-09-05
Add v2i64 comparison operations

tlively:
There did not seem to be a reason for i64x2 comparison operations to be omitted, so the PR adds them and adjusts instruction opcodes accordingly.


AndrewScheidecker:
They were explicitly removed by https://github.com/WebAssembly/simd/pull/22, citing consensus at the [May 2017 CG meeting](https://github.com/WebAssembly/meetings/blob/master/2017/CG-05.md#simd).

(I don't have an opinion here, just stating the history)

dtig:
It looks like the poll on the compare operations was ambivalent, and not contentious ("No consensus, but not objection either." ). I think it's reasonable to assume that these can be added back if there was performance, or usability evidence that these belong in the set of operations.  Is there a precedence for reserved opcode numbers? If so I think the opcodes can be retained without adding the comparison operations.

tlively:
Ok, I've changed this PR to add holes in the opcode space where operations we explicitly don't support would logically go to improve our forward compatibility story. These holes are represented by crossed out ops in the binary encoding table.

dtig:
I think it's best to leave reserved opcodes as opcode holes instead of representing them as strike-through ops because there's no consensus on these operations being included. Some of these could be added back, or none at all.

PR #40, opened 2018-09-21
Rename truncations to avoid colon in name

tlively:
Having colons in instruction names would require special handling in
existing assemblers, which seems like unnecessary extra work.


tlively:
Thanks, @sunfishcode. I wasn't aware of the names in the nontrapping conversions proposal, so I will go chime in over there.

tlively:
Closing this for now, since it is outdated and the discussion is happening elsewhere.

PR #42, opened 2018-09-28
Update BinarySIMD.md

dtig:
Update opcodes to use hex values, and include reserved opcodes for integer min/max operations.

### DIFF
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -101,7 +101,7 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
 | `i64x2.all_true`          |    `0x48`| -                  |
 | `i8x16.eq`                |    `0x49`| -                  |
 | `i16x8.eq`                |    `0x4a`| -                  |
-| `i32x4.eq`                |    `0x4c`| -                  |
+| `i32x4.eq`                |    `0x4b`| -                  |
 | `f32x4.eq`                |    `0x4d`| -                  |
 | `f64x2.eq`                |    `0x4e`| -                  |
 | `i8x16.ne`                |    `0x4f`| -                  |
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -27,143 +27,143 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
 
 | Instruction               | `simdop` | Immediate operands |
 | --------------------------|---------:|--------------------|
-| `v128.const`              |        0 | i:ImmByte[16]      |
-| `v128.load`               |        1 | m:memarg           |
-| `v128.store`              |        2 | m:memarg           |
-| `i8x16.splat`             |        3 | -                  |
-| `i16x8.splat`             |        4 | -                  |
-| `i32x4.splat`             |        5 | -                  |
-| `i64x2.splat`             |        6 | -                  |
-| `f32x4.splat`             |        7 | -                  |
-| `f64x2.splat`             |        8 | -                  |
-| `i8x16.extract_lane_s`    |        9 | i:LaneIdx16        |
-| `i8x16.extract_lane_u`    |       10 | i:LaneIdx16        |
-| `i16x8.extract_lane_s`    |       11 | i:LaneIdx8         |
-| `i16x8.extract_lane_u`    |       12 | i:LaneIdx8         |
-| `i32x4.extract_lane`      |       13 | i:LaneIdx4         |
-| `i64x2.extract_lane`      |       14 | i:LaneIdx2         |
-| `f32x4.extract_lane`      |       15 | i:LaneIdx4         |
-| `f64x2.extract_lane`      |       16 | i:LaneIdx2         |
-| `i8x16.replace_lane`      |       17 | i:LaneIdx16        |
-| `i16x8.replace_lane`      |       18 | i:LaneIdx8         |
-| `i32x4.replace_lane`      |       19 | i:LaneIdx4         |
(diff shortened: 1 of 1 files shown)


proposals/simd/BinarySIMD.md
```diff
@@ -27,143 +27,143 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
...
+| `i16x8.mul`               |    `0x21`| -                  |
+| `i32x4.mul`               |    `0x22`| -                  |
+| `i8x16.neg`               |    `0x24`| -                  |
+| `i16x8.neg`               |    `0x25`| -                  |
+| `i32x4.neg`               |    `0x26`| -                  |
+| `i64x2.neg`               |    `0x27`| -                  |
+| `i8x16.add_saturate_s`    |    `0x28`| -                  |
+| `i8x16.add_saturate_u`    |    `0x29`| -                  |
+| `i16x8.add_saturate_s`    |    `0x2a`| -                  |
+| `i16x8.add_saturate_u`    |    `0x2b`| -                  |
```
tlively: Maybe reserve opcodes here as well?

proposals/simd/BinarySIMD.md
```diff
@@ -27,143 +27,143 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
...
+| `i8x16.sub_saturate_s`    |    `0x2c`| -                  |
+| `i8x16.sub_saturate_u`    |    `0x2d`| -                  |
+| `i16x8.sub_saturate_s`    |    `0x2e`| -                  |
+| `i16x8.sub_saturate_u`    |    `0x2f`| -                  |
```
tlively: And here

proposals/simd/BinarySIMD.md
```diff
@@ -27,143 +27,143 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
...
+| `i16x8.any_true`          |    `0x42`| -                  |
+| `i32x4.any_true`          |    `0x43`| -                  |
+| `i64x2.any_true`          |    `0x44`| -                  |
+| `i8x16.all_true`          |    `0x45`| -                  |
+| `i16x8.all_true`          |    `0x46`| -                  |
+| `i32x4.all_true`          |    `0x47`| -                  |
+| `i64x2.all_true`          |    `0x48`| -                  |
+| `i8x16.eq`                |    `0x49`| -                  |
+| `i16x8.eq`                |    `0x4a`| -                  |
+| `i32x4.eq`                |    `0x4c`| -                  |
```
tlively: Looks like the hole is in the wrong place here.

proposals/simd/BinarySIMD.md
```diff
@@ -27,143 +27,143 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
...
+| `i32x4.ge_s`              |    `0x77`| -                  |
+| `i32x4.ge_u`              |    `0x78`| -                  |
+| `f32x4.ge`                |    `0x7b`| -                  |
+| `f64x2.ge`                |    `0x7c`| -                  |
+| `f32x4.neg`               |    `0x7d`| -                  |
+| `f64x2.neg`               |    `0x7e`| -                  |
+| `f32x4.abs`               |    `0x7f`| -                  |
+| `f64x2.abs`               |    `0x80`| -                  |
+| `f32x4.min`               |    `0x87`| -                  |
+| `f64x2.min`               |    `0x88`| -                  |
```
tlively: This part of the opcode space is mostly float-only operations. Would it make sense to reserve opcodes for integer min/max elsewhere? It could be similar to add, sub, and mul. Alternatively, we could reorganize to put floating point arithmetic ops next to their integer counterparts.

dtig: That makes sense, I mostly just ported an old pull request to the new space. I could go ahead and re-organize so the floating point arithmetic ops are next to the integer ones, or just move the sprinkled in Floating point ops to the end (which is somewhat consistent with other Wasm opcodes - the separation between integer/floating point opcodes). Alternatively I would prefer to order these according to type - mostly because some types have operations that are not supported on other types, and it makes the overall organization cleaner IMO. WDYT?

proposals/simd/BinarySIMD.md
dtig: Can reserve, did you have something specific in mind to fill the space?

proposals/simd/BinarySIMD.md
dtig: Same question as above.

proposals/simd/BinarySIMD.md
tlively: I was thinking `{i32x4,i64x2}.add_sat_{s,u}` could logically go here, even if it could not be implemented reasonably today. My thinking is that we should reserve opcodes to be maximally future proof and ensure we have a nice layout no matter how hardware or our priorities change. But if there is no way we would ever want to add these ops, then reserving opcodes for them probably isn't necessary.

proposals/simd/BinarySIMD.md
tlively: `{i32x4,i64x2}.sub_saturate_{s,u}`

proposals/simd/BinarySIMD.md
tlively: Moving all fp ops to the end sounds good. I generally prefer ordering by operation over ordering by type, though.

binji: The MVP instructions are grouped first by what kind of instruction (unary, binary, conversion, etc.) then by type. It would be nice to do the same here, if we could.

dtig:
Closing in favor of #48

PR #51, opened 2018-11-07
Opcode adjustments

tlively:
Move unary instructions before binary instructions in each opcode block. Move reserved space for reciprocal sqrt and approx reciprocal to immediately after sqrt.

Unresolved questions:

 - Should conversion ops be left in their own section? They're mixed in with the other instructions in MVP.

 - Where should shifts go? They're not really unary or binary operations. They're currently at the end of the binary operation sections.


binji:
> Should conversion ops be left in their own section? They're mixed in with the other instructions in MVP.

They're all at the end, right?

> Where should shifts go? They're not really unary or binary operations. They're currently at the end of the binary operation sections.

Chatted with @dtig about this -- I didn't realize that we were planning to make the SIMD shifts take an immediate shift amount. Seems like it may be worth discussing in a separate issue, since it is a pretty subtle difference from the scalar shifts.

proposals/simd/BinarySIMD.md
```diff
@@ -98,67 +98,67 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
...
+| `i64x2.any_true`          |    `0x85`| -                  |
+| `i64x2.all_true`          |    `0x86`| -                  |
+| `i64x2.add`               |    `0x87`| -                  |
+| `i64x2.sub`               |    `0x8a`| -                  |
+| `i64x2.shl`               |    `0x92`| -                  |
+| `i64x2.shr_s`             |    `0x93`| -                  |
+| `i64x2.shr_u`             |    `0x94`| -                  |
+| `f32x4.abs`               |    `0x95`| -                  |
+| `f32x4.neg`               |    `0x96`| -                  |
+| `f32x4.sqrt`              |    `0x97`| -                  |
```
binji: what's the gap here for?

tlively: Experimentation with reciprocal sqrt and approximate reciprocal, as recommended by @dtig.

binji: Ah, cool. Sounds good!

tlively:
> > Should conversion ops be left in their own section? They're mixed in with the other instructions in MVP.
> 
> They're all at the end, right?

Yes, my mistake. I saw `f32.trunc` and thought it was a conversion op.

> 
> > Where should shifts go? They're not really unary or binary operations. They're currently at the end of the binary operation sections.
> 
> Chatted with @dtig about this -- I didn't realize that we were planning to make the SIMD shifts take an immediate shift amount. Seems like it may be worth discussing in a separate issue, since it is a pretty subtle difference from the scalar shifts.

Either way, they are not quite the same as other unary operations (V128 -> V128) or binary operations (V128, V128 -> V128).

dtig:
> Move unary instructions before binary instructions in each opcode block. Move reserved space for reciprocal sqrt and approx reciprocal to immediately after sqrt.
> 
> Unresolved questions:
> 
> * Should conversion ops be left in their own section? They're mixed in with the other instructions in MVP.
> * Where should shifts go? They're not really unary or binary operations. They're currently at the end of the binary operation sections.

I've opened #52 to figure out shift semantics, but I suggest moving them to between the unary and binary operations that way they won't need to be moved if immediates are used, but still make sense if they're not as they're different from the other binary operations.

dtig:
> > Should conversion ops be left in their own section? They're mixed in with the other instructions in MVP.
> 
> They're all at the end, right? (edit: in the MVP, I mean)
> 
> > Where should shifts go? They're not really unary or binary operations. They're currently at the end of the binary operation sections.
> 
> Chatted with @dtig about this -- I didn't realize that we were planning to make the SIMD shifts take an immediate shift amount. Seems like it may be worth discussing in a separate issue, since it is a pretty subtle difference from the scalar shifts.

There really wasn't a plan, I just incorrectly assumed that's what was specified. I've opened #52 to see if there should be any changes to the way shifts are currently specified.

PR #59, opened 2018-12-14
Remove text about flushing of subnormals

dtig:
Default to full IEEE 754 support for now. The issue of handling subnormals needs to be addressed (possibly by a global FTZ flag or mode switching), but allowing platforms to behave differently on FTZ semantics should not be baked into the specification.


aretmr:
LGTM

PR #75, opened 2019-04-30
Organizing sync notes into a 'meeting' folder for future reference. 

arunetm:
Organizing sync notes into a 'meeting' folder for future reference.


arunetm:
@dtig please take a look.

penzn:
@arunetm I think that is a good idea, thank you!

Issue #28, opened 2018-03-30
Adding new SIMD instructions to load sign and zero extend 8, 16 and 32 byte integers


rrwinterton:
[proposal-webAssembly-SIMD-modification.pdf](https://github.com/WebAssembly/simd/files/1864317/proposal-webAssembly-SIMD-modification.pdf)

arunetm:
@PeterJensen @kenchris

jfbastien:
Could you put the content of the PDF into this bug? PDFs aren't searchable in general.

kenchris:
Proposal WebAssembly SIMD Modification
===

Currently as proposed there is an instruction defined in the WASM SIMD ISA as follows.
Instruction – i8x16.mul which is a register to register operation that takes 16 8 bit integers and
multiplies them together resulting in an 8 bit value. If the distribution of the integers it flat this
would result in approximately 50 percent of the instructions with overflow. This instruction is
not too useful in real life applications that is known. This instruction could be left in the ISA mix
but for IA would require a significant amount of instructions and would have to require
overflow handling and specifications to achieve cross platform compatibility. This really is
messy for no known applications. This overflow handling is currently not defined in the WASM
SIMD specification.

Proposal is to remove this instruction as known applications don’t use i8x16.mul functionality
and add the widening load integer instructions. This was previously proposed as a candidate in
bullet two back in November last year.

![image](https://user-images.githubusercontent.com/1101339/38148184-ae56ccce-3455-11e8-9a01-47dd62ac545c.png)

Proposed new instructions:
===
In conjunction with the potential change of the i8x16.mul removal it is proposed to create an
instruction that would have six new load instructions to make integer multiplies easier. The
first is i8x16. This would make i8, i16, i32 multiplies useful and more practical for applications
such as machine learning, image compression and video and rendering data processing.
The new instructions would take consecutive integers of the corresponding size and zero sign
extend and sign extend the consecutive bytes, words or dword to

![image](https://user-images.githubusercontent.com/1101339/38148202-bfb1334c-3455-11e8-9cb2-3fb7b2d8e11c.png)

Intel and ARM both have this capability by doing the following:

movzxbw,
movzxwd
movzxdq
movsxbw
movsxwd
movsxdq

ARM:

![image](https://user-images.githubusercontent.com/1101339/38148228-d1d0c772-3455-11e8-97d5-0edd957b0604.png)

So the new instructions for WASM would be defined as follows:

[...]

Issue #21, opened 2017-05-25
Add opcodes for converting between integer vectors of different lane size.


billbudge:
Integer conversions are useful when data is stored in a small integer format, but calculations are performed at a higher precision. They allow data to be unpacked and packed. Synthesizing these with opcodes in the spec is awkward. They are well supported on all architectures.

Widening conversions, with signed or unsigned saturation. These are unary operations, converting low or high half of the smaller lane vector in order to avoid multiple return values.

i32x4.SConvertI16x8Low(x: v128) -> v128
i32x4.SConvertI16x8High(x: v128) -> v128
i32x4.UConvertI16x8Low(x: v128) -> v128
i32x4.UConvertI16x8High(x: v128) -> v128

i16x8.SConvert8x16Low(x: v128) -> v128
i16x8.SConvert8x16High(x: v128) -> v128
i16x8.UConvert8x16Low(x: v128) -> v128
i16x8.UConvert8x16High(x: v128) -> v128

Narrowing conversions, signed and unsigned. These are binary operations, converting two input vectors into a single smaller lane vector.

i16x8.SConvertI32x4(x: v128, y:v128) -> v128
i16x8.UConvertI32x4(x: v128, y:v128) -> v128

i8x16.SConvertI16x8(x: v128, y:v128) -> v128
i8x16.UConvertI16x8(x: v128, y:v128) -> v128

stoklund:
I think you swapped "narrowing" and "widening".

The widening conversions could be implemented as a swizzle and a right shift. I don't think we want to depend on implementations recognizing that pattern, though.

billbudge:
Fixed that. Thanks.

PR #77, opened 2019-05-03
simdsignunsignedextendedload.md

rrwinterton:
proposes 8, 16, 32 signed and unsigned extended load instructions.


proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
+### **Proposal WebAssembly SIMD Modification**
+
+Currently as proposed there is an instructions defined in the WASM SIMD ISA as follows.
+
+**i8x16.mul** which is a register to register operation that takes 16 8 bit integers and
+
+multiplies them together resulting in an 8 bit value. If the distribution of the integers it flat this
```
gnzlbg: typo: it -> is

---

Also what do you mean with the distribution of the integers being flat? And even if that is the case, does it always result in overflow? e.g., if all integers are `0`, then the distribution would be as flat as it gets but no overflow would happen.

proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
...
+
+would result in a large percent of the instructions with overflow. This is a problem for many applications.
```
gnzlbg: i8x16.mul implements wrapping integer multiplication, therefore, I don't see why overflow there is a problem (it's not an error, nor UB, but perfectly defined)

proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
...
+
+
+### Proposed new instructions
+
+Six new load instructions are being proposed to make integer multiplies easier. i8x16zxload, i8x16sxload, i16x8zxload, i16x8sxload, i32x4zxload, i32x4sxload. This would make i8, i16, i32 multiplies useful and more practical for applications such as machine learning, image compression and video and rendering data processing.The new instructions would take consecutive integers of the corresponding size and zero sign extend and sign extend the consecutive bytes, words or dword to the promoted size of signed or unsigned data. An example of zero sign extend is shown below:  Intel and ARM both have this capability by doing the following:
```
gnzlbg: Nowhere does the text say what the proposed instructions actually do, what are their argument types, results, operational semantics, etc.

proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
...
+*   i32x2.zxload
+*   i8x8.sxload
+*   i16x4.sxload
+*   i32x2.sxload
+
+As a result of these new instructions a multiply can now be done without worrying about signed
+
+and unsigned overflow on the data it operates on.
+
+The following is a partial sample example of how sign extended loads are be used in a matrix multiply of 8 bit integers:
```
gnzlbg: Please show WASM instead.

proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
...
+So the new instructions for WASM would be defined as follows:
+
+
+
+*   i8x8.zxload
+*   i16x4.zxload
+*   i32x2.zxload
+*   i8x8.sxload
+*   i16x4.sxload
+*   i32x2.sxload
```
gnzlbg: All these types are 64-bit wide vector types, but the spec does not define `v64` anywhere.

proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
...
+
+```
+       "pmovzxbw 0x00(%[mem]), %%xmm0\n\t"
+       "pshufd $0x00,%%xmm1,%%xmm2     \n\t"
+       "pshufd $0x55,%%xmm1,%%xmm3     \n\t"
+       "pmaddwd %%xmm0, %%xmm2         \n\t"
+       "pmaddwd %%xmm0, %%xmm3         \n\t"
+       "paddd %%xmm2, %%xmm4           \n\t"
+       "paddd %%xmm3, %%xmm5           \n\t"
+
```
gnzlbg: typo: backticks missing

proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
...
+*   i8x8.zxload
+*   i16x4.zxload
+*   i32x2.zxload
+*   i8x8.sxload
+*   i16x4.sxload
+*   i32x2.sxload
+
+As a result of these new instructions a multiply can now be done without worrying about signed
+
```
gnzlbg: typo: line break

proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
+### **Proposal WebAssembly SIMD Modification**
+
+Currently as proposed there is an instructions defined in the WASM SIMD ISA as follows.
+
+**i8x16.mul** which is a register to register operation that takes 16 8 bit integers and
+
+multiplies them together resulting in an 8 bit value. If the distribution of the integers it flat this
+
```
gnzlbg: typo: line break

proposals/simd/simdsignandunsignedextendedloads.md
```diff
@@ -0,0 +1,64 @@
+### **Proposal WebAssembly SIMD Modification**
+
+Currently as proposed there is an instructions defined in the WASM SIMD ISA as follows.
+
+**i8x16.mul** which is a register to register operation that takes 16 8 bit integers and
+
```
gnzlbg: typo: line break

gnzlbg:
Please link the issue (https://github.com/WebAssembly/simd/issues/28) and the [rendered document](https://github.com/rrwinterton/simd/blob/patch-1/proposals/simd/simdsignandunsignedextendedloads.md) in the top-level comment.

proposals/simd/simdsignandunsignedextendedloads.md
gnzlbg: I find the document and the issue quite unclear, so I might be completely misunderstanding what problem this is trying to address, but if what you want is to be able load, e.g., 4 x 16-bit integers into a `v128` containing 4 x 32-bit integers by sign or zero extending them, then I would prefer something that's more similar to the scalar intrinsics, e.g., like `i32x4.load_i16x4_s(memarg) -> v128`.

penzn:
I believe we have populated the opcodes and instruction descriptions in this PR. One thing to note is that it implements #28 and not #21 (latter proposes vector to vector operations).

proposals/simd/simdsignandunsignedextendedloads.md
dtig: I think the intent here is to introduce something equivalent to the intel pmovzx(b/d/q) instruction, so the operation would be as follows - 

Packed_Zero_Extend_BYTE_to_WORD(DEST, SRC)
DEST[15:0] <- ZeroExtend(SRC[7:0]);
DEST[31:16] <- ZeroExtend(SRC[15:8]);
DEST[47:32] <- ZeroExtend(SRC[23:16]);
DEST[63:48] <- ZeroExtend(SRC[31:24]);
DEST[79:64] <- ZeroExtend(SRC[39:32]);
DEST[95:80] <- ZeroExtend(SRC[47:40]);
DEST[111:96] <- ZeroExtend(SRC[55:48]);
DEST[127:112] <- ZeroExtend(SRC[63:56]);

So it's not strictly a V128->V128 operation. I do agree that this would benefit from clearer names instead of zx/sx convention that's not very readable.

proposals/simd/docs/SIMD-sign-and-zero-extended-loads.md
```diff
@@ -0,0 +1,64 @@
...
+
+### Proposed new instructions
+
+Six new load instructions are being proposed to make integer multiplies easier. i8x16zxload, i8x16sxload, i16x8zxload, i16x8sxload, i32x4zxload, i32x4sxload. This would make i8, i16, i32 multiplies useful and more practical for applications such as machine learning, image compression and video and rendering data processing.The new instructions would take consecutive integers of the corresponding size and zero sign extend and sign extend the consecutive bytes, words or dword to the promoted size of signed or unsigned data. An example of zero sign extend is shown below:  Intel and ARM both have this capability by doing the following:
+
+Intel Instructions:
+
+
+
+*   movzxbw
```
dtig: Did you mean to include the packed versions of these instructions (ex: pmovzxbw/pmovsxbw)? Please link the exact instruction you mean if that's not the one.

proposals/simd/docs/SIMD-sign-and-zero-extended-loads.md
```diff
@@ -0,0 +1,64 @@
+### **Proposal WebAssembly SIMD Modification**
```
dtig: Can this document be removed? While this is useful information, I don't see a need for this to be included here.

proposals/simd/SIMD.md
```diff
@@ -666,6 +666,15 @@ natural alignment.
 
 Load a `v128` vector from the given heap address.
 
+Extended loads:
```
dtig: Can this be made clear here that this is not operating on 64bit widths widening to a 128-bit vector? Having a concise description here will be useful.

proposals/simd/docs/SIMD-sign-and-zero-extended-loads.md
penzn: Do you want it to be included in the body of the PR or just simply remove it? @rrwinterton, what do you think?

proposals/simd/SIMD.md
penzn: Do you mean to clarify that we are extending every lane, rather than widening a 64-bit value?

penzn:
The description of the change (so that we can delete the file):

### Proposed new instructions

Six new load instructions are being proposed to make integer multiplies easier. i8x16.zxload, i8x16.sxload, i16x8.zxload, i16x8.sxload, i32x4.zxload, i32x4.sxload. This would make i8, i16, i32 multiplies useful and more practical for applications such as machine learning, image compression and video and rendering data processing. The new instructions would take consecutive integers of the corresponding size and zero sign extend and sign extend the consecutive bytes, words or dword to the promoted size of signed or unsigned data. An example of zero sign extend is shown below:  Intel and ARM both have this capability by doing the following:

Intel Instructions:

*   `movzxbw`
*   `movzxwd`
*   `movzxdq`
*   `movsxbw`
*   `movsxwd`
*   `movsxdq`

ARM Instructions:

*   `LDR X0, [X1]` Load from the address in X1
*   `LDR X0, [X1, #8]` Load from address X1 + 8
*   `LDR X0, [X1, X2]` Load from address X1 + X2
*   `LDR X0, [X1, X2, LSL, #3]` Load from address X1 + (X2 << 3)
*   `LDR X0, [X1, W2, SXTW]` Load from address X1 + sign extend(W2)
*   `LDR X0, [X1, W2, SXTW, #3]` Load from address X1 + (sign extend(W2) << 3)

So the new instructions for WASM would be defined as follows:

*   `i8x8.zxload`
*   `i16x4.zxload`
*   `i32x2.zxload`
*   `i8x8.sxload`
*   `i16x4.sxload`
*   `i32x2.sxload`

As a result of these new instructions a multiply can now be done without worrying about signed

and unsigned overflow on the data it operates on.

The following is a partial sample example of how sign extended loads are be used in a matrix multiply of 8 bit integers:


```
       "pmovzxbw 0x00(%[mem]), %%xmm0\n\t"
       "pshufd $0x00,%%xmm1,%%xmm2     \n\t"
       "pshufd $0x55,%%xmm1,%%xmm3     \n\t"
       "pmaddwd %%xmm0, %%xmm2         \n\t"
       "pmaddwd %%xmm0, %%xmm3         \n\t"
       "paddd %%xmm2, %%xmm4           \n\t"
       "paddd %%xmm3, %%xmm5           \n\t"
```

AndrewScheidecker:
These names would be more consistent with the scalar extending load instructions (i.e. `i32.load8_s`):

| original | alternative 1 | alternative 2 |
| -------- | ------------ | --------- |
| `i8x8.zxload` | `i16x8.load8x8_u` | `i16x8.load8_u` |
| `i16x4.zxload` | `i32x4.load16x4_u` | `i32x4.load16_u` |
| `i32x2.zxload` | `i64x2.load32x2_u` | `i64x2.load32_u` |
| `i8x8.sxload` | `i16x8.load8x8_s` | `i16x8.load8_s` |
| `i16x4.sxload` | `i32x4.load16x4_s` | `i32x4.load16_s` |
| `i32x2.sxload` | `i64x2.load32x2_s` | `i64x2.load32_s` |

tlively:
Thanks for those suggestions, @AndrewScheidecker! I personally prefer alternative 1 because they are more explicit about how much memory is actually being loaded.

penzn:
I like that scheme as well, incorporated in rrwinterton/simd#4

Issue #78, opened 2019-05-15
Naming shuffle instructions


tlively:
#71 added a new shuffle instruction but was explicitly merged without consensus on the name of the new instruction or the instruction that PR renamed to `v8x16.shuffle2_imm` (previously `v8x16.shuffle`). Now that the instruction is merged into the proposal, we should return to the bikeshed of the names.

In the previous PR I held an informal vote for different parts of the naming schemes: https://github.com/WebAssembly/simd/pull/71#issuecomment-472691503

At time of writing, these are the results:
 - 3 votes for using `permute*` and `shuffle*` as the base names
 - 1 vote for using `shuffle1*` and `shuffle2*` as the base names
 - 0 votes for using `_dyn` as a suffix marking dynamic indices
 - 1 vote for using `_vec` as a suffix marking dynamic indices
 - 3 votes for not having a suffix to mark dynamic indices

If we went with the most popular results, our two instructions would be named `v8x16.permute` and `v8x16.shuffle_imm`. What does everyone think of that?

tlively:
Bump! With our new stage 2 status and @Honry about to start writing spec tests, it would be good to get the naming squared away. Please give me at least a quick :+1: if you would be fine with `v8x16.permute` and `v8x16.shuffle_imm` and a :confused: if you prefer something else.

gnzlbg:
The whole point of the names, as opposed to the binary encoding, is clarity. I still find it misleading that `v8x16.permute` does not perform a permutation. Something like `v8x16.shuffle1` and `v8x16.shuffle2_imm` feel more accurate to me.

tlively:
We talked about this at the last meeting, and reached a general consensus in the room that `swizzle` would be a good name for the single-argument version and `shuffle` for the two-argument version. @gnzlbg does that sound acceptable to you?

gnzlbg:
Sounds good, `swizzle` operations take one argument, allow repetition, and is a technical term in computer graphics.

PR #84, opened 2019-07-18
Rename shuffles

tlively:


proposals/simd/BinarySIMD.md
```diff
@@ -166,5 +166,5 @@ The `v8x16.shuffle2_imm` instruction has 16 bytes after `simdop`.
 | `f32x4.convert_u/i32x4`   |    `0xb0`| -                  |
 | `f64x2.convert_s/i64x2`   |    `0xb1`| -                  |
 | `f64x2.convert_u/i64x2`   |    `0xb2`| -                  |
-| `v8x16.shuffle1`          |    `0xc0`| -                  |
-| `v8x16.shuffle2_imm`      |    `0xc1`| s:LaneIdx32[16]    |
\ No newline at end of file
+| `v8x16.swizzle`           |    `0xc0`| -                  |
+| `v8x16.shuffle_imm`       |    `0xc1`| s:LaneIdx32[16]    |
```
dtig: IIRC, we agreed on swizzle, and shuffle and not shuffle_imm. Not opposed to having an explicit name, but just pointing out in case anyone has strong opinions about this.

tlively: You're right, the [comment](https://github.com/WebAssembly/simd/issues/78#issuecomment-508282544)  I wrote at the time just uses `shuffle`. I'm not sure why I still had the `_imm` suffix in my head.

PR #86, opened 2019-07-23
This is to assign the opcodes for the load extend instructions

rrwinterton:
This proposed change is to add the 6 load extend instructions for signed and unsigned integers using the naming convention suggested by others in the pull request https://github.com/WebAssembly/simd/issues/28


tlively:
Should this be part of #77?

penzn:
#77 has four of those (missing 32->64 operations)

tlively:
Can we close this and add the extra instructions in #77, then?

penzn:
@rrwinterton, @tlively, would something linke rrwinterton/simd#3 work?

tlively:
Yes, that would be ideal. The discussion around these instructions is very spread out at the moment, and I think it would be good to concentrate all development and discussion on #77 so it is all in one place.

penzn:
rrwinterton/simd#3 is merged, which populated #77, this can probably be closed now

PR #89, opened 2019-07-31
Add integer widen/narrow conversions

### LINKED ISSUE #21: Add opcodes for converting between integer vectors of different lane size.

ngzhian:
Fixes #21 
@billbudge
@dtig


ngzhian:
Lots of changes due to formatting of the markdown table, but if you choose to "view in rich diff" you should see the changes more clearly.

proposals/simd/ImplementationStatus.md
```diff
@@ -1,150 +1,162 @@
...
+| `i16x8.convert_u/i8x16High` |                           |                    |                    |                    |
+| `i32x4.convert_s/i16x8Low`  |                           |                    |                    |                    |
+| `i32x4.convert_s/i16x8High` |                           |                    |                    |                    |
+| `i32x4.convert_u/i16x8Low`  |                           |                    |                    |                    |
+| `i32x4.convert_u/i16x8High` |                           |                    |                    |                    |
+| `i16x8.convert_s/i32x4`     |                           |                    |                    |                    |
+| `i16x8.convert_u/i32x4`     |                           |                    |                    |                    |
+| `i8x16.convert_s/i8x16`     |                           |                    |                    |                    |
+| `i8x16.convert_u/i8x16`     |                           |                    |                    |                    |
```
dtig: V8 implements some of these, so can the table reflect that? Also cc @tlively does the toolchain support this already?

ngzhian: Done, i also found some typos and ordering issues, fixed in next commit

proposals/simd/SIMD.md
```diff
@@ -784,3 +784,25 @@ Lane-wise saturating conversion from floating point to integer using the IEEE
...
+
+### Integer to integer widening
+* `i16x8.convert_s/i8x16Low(a: v128) -> v128`
+* `i16x8.convert_s/i8x16High(a: v128) -> v128`
+* `i16x8.convert_u/i8x16Low(a: v128) -> v128`
+* `i16x8.convert_u/i8x16High(a: v128) -> v128`
+* `i32x4.convert_s/i16x8Low(a: v128) -> v128`
+* `i32x4.convert_s/i16x8High(a: v128) -> v128`
+* `i32x4.convert_u/i16x8Low(a: v128) -> v128`
+* `i32x4.convert_u/i16x8High(a: v128) -> v128`
```
tlively: These names should probably be updated since capital letters and slashes are now inconsistent with the rest of the instructions. How about `i16x8.{widen,narrow}_i8x16_s` and similar.

proposals/simd/SIMD.md
```diff
@@ -784,3 +784,25 @@ Lane-wise saturating conversion from floating point to integer using the IEEE
...
+
+Converts low or high half of the smaller lane vector to a larger lane vector,
+with signed or unsigned saturation.
+
+### Integer to integer narrowing
+* `i8x16.convert_s/i16x8(a: v128, b: v128) -> v128`
+* `i8x16.convert_u/i16x8(a: v128, b: v128) -> v128`
+* `i16x8.convert_s/i32x4(a: v128, b: v128) -> v128`
+* `i16x8.convert_u/i32x4(a: v128, b: v128) -> v128`
```
tlively: I'm not sure what the difference between signed and unsigned narrowing is. Can that be elaborated?

proposals/simd/SIMD.md
```diff
@@ -784,3 +784,25 @@ Lane-wise saturating conversion from floating point to integer using the IEEE
...
+
+### Integer to integer widening
+* `i16x8.convert_s/i8x16Low(a: v128) -> v128`
+* `i16x8.convert_s/i8x16High(a: v128) -> v128`
+* `i16x8.convert_u/i8x16Low(a: v128) -> v128`
+* `i16x8.convert_u/i8x16High(a: v128) -> v128`
+* `i32x4.convert_s/i16x8Low(a: v128) -> v128`
+* `i32x4.convert_s/i16x8High(a: v128) -> v128`
+* `i32x4.convert_u/i16x8Low(a: v128) -> v128`
+* `i32x4.convert_u/i16x8High(a: v128) -> v128`
```
tlively: These names should be updated to be consistent with the naming scheme of the rest of the instructions. How about `i16x8.widen_{low,high}_{s,u}_i8x16` and similar? Same for the narrowing.

proposals/simd/SIMD.md
dtig: The proposal still uses the old scheme, I'm okay with the proposed {widen, narrow} convention - will update the rest of the proposal to match too.

ngzhian: what would i32x4.convert_s/i16x8Low look like?
- i32x4.widen_i16x8_s_low
- i32x4.widen_low_i16x8_s

proposals/simd/SIMD.md
ngzhian: Good question @billbudge can you help here?
I think its whether the truncation care about the sign.
E.g. given 428, 0000000110101100
a truncation unsigned will take the LSB, 10101100, and return that modulo 128, so 44 (0101100).
a truncation signed will return 10101100, which is -84.
I referred to the answer here https://stackoverflow.com/questions/19273658/integer-conversionsnarrowing-widening-undefined-behaviour

proposals/simd/SIMD.md
tlively: I think we agreed on a convention that the `_s` or `_u` signifying "signed" or "unsigned" is always at the very end of the instruction, so it would be your second example.

proposals/simd/SIMD.md
tlively: I would have interpreted the "convert" in that SO answer as being a reinterpret that does not change any bytes. In particular LLVM does not have signed/unsigned truncation, although it does allow you to *interpret* the result as signed or unsigned after you have thrown out the most significant bits.

dtig: The signed/unsigned in the narrowing operations refers to how these operations will handle overflow conditions. For example, in the case of i16x8.convert_s/i32x4, this will convert the 4 i32 values to 8 i16 values using signed saturation on overflow, i.e. it will map diirectly to [packssdw](https://www.felixcloutier.com/x86/packsswb:packssdw).

ngzhian: Thanks for your explanation! I added a short note in 39a7408, please see if that's okay.

proposals/simd/SIMD.md
ngzhian: Done, changed all the widen/narrow operation names to follow this convention.

proposals/simd/SIMD.md
tlively: Ahh, I understand now. I hadn't realized there was saturation involved :+1:

proposals/simd/SIMD.md
```diff
@@ -784,3 +784,27 @@ Lane-wise saturating conversion from floating point to integer using the IEEE
...
+* `i16x8.widen_high_i8x16_s(a: v128) -> v128`
+* `i16x8.widen_low_i8x16_u(a: v128) -> v128`
+* `i16x8.widen_high_i8x16_u(a: v128) -> v128`
+* `i32x4.widen_low_i16x8_s(a: v128) -> v128`
+* `i32x4.widen_high_i16x8_s(a: v128) -> v128`
+* `i32x4.widen_low_i16x8_u(a: v128) -> v128`
+* `i32x4.widen_high_i16x8_u(a: v128) -> v128`
+
+Converts low or high half of the smaller lane vector to a larger lane vector,
+with signed or unsigned saturation.
```
tlively: This should probably be "sign extended" or "zero extended" rather than saturation.

ngzhian: Done

dtig:
These had previous consensus, where we agreed they are useful to include, merging.

AndrewScheidecker:
Can these opcodes be renumbered to leave space for `i64` variants of these instructions?

ngzhian:
I was advised to put these opcodes at the end, I don't think these are finalized or anything yet, so it should be okay to renumber them.

Issue #94, opened 2019-08-14
Unsigned narrowing


penzn:
Widening and narrowing operations were proposed in #21 and added in #89. There is a discussion about what should be the input for unsigned narrowing instructions in #91. The issue is that x86(64) SIMD narrowing instructions treat input as _signed_, even when output is not. ARM [supports][a] both signed-unsigned and unsigned-unsigned narrowing.

I can see value in "signed to unsigned" narrowing for things like RGBA graphics -- results of signed integer arithmetic that would be packed into unsigned RGBA output. For one example, see [Sobel operator][s], other image filters would use it as well. Would "unsigned to unsigned" narrowing be equally useful? Where would it be used?

Another problem is that emulating "unsigned to unsigned" narrowing on x86(64) requires about 4 instructions, which is not a good value proposition for operation that would work on 2*, 4, or 8 lanes, as only the 8-lane version would be faster than scalar.

*if 64 bit lanes are supported

[a]: https://github.com/WebAssembly/simd/pull/91#issuecomment-520385576
[s]: https://en.wikipedia.org/wiki/Sobel_operator

penzn:
My back of the envelope instruction sequence to emulate unsigned-unsigned narrowing in SSE would look something like the following, there are other ways as well:

```
pcmpeq(w/d/q) ;; on the same register, get lanes filled with 0xFF..F
psrl(w/d/q)   ;; shift right by one to get 0x7FF..F
pminu(w/d/q)  ;; replace negative lanes with 0x7FF..F (as they are bigger when cast to unsigned)
packus(dw/wb) ;; get the final result
```

arunetm:
Can we have a spc PR with these recommendations and seek feedback there? Unless there are use cases benefiting from unsigned to unsigned narrowing, it is justifiable to drop them from the initial proposal.

penzn:
#91 does the needful by stating that the inputs are signed.

penzn:
Sorry, a longer response now. Another way to look at it is that there are two questions: performance of unsigned narrowing with unsigned inputs and the benefits of unsigned narrowing with signed input.

#91 has the spec change to substitute the former with the latter.

Alternatively, we can add the other variant in alongside the ones that have been added. The advantages of having unsigned narrowing with signed inputs are described above, and it can be done in one instruction on both x86 and ARM.

penzn:
Closing as we merged the alternative in PR #91

PR #95, opened 2019-08-15
Introduce unsigned narrowing with signed inputs

penzn:
Narrowing signed inputs to unsigned outputs is widely used in image filtering, when output of a transformation with signed integer coefficients is packed into an array of unsigned integers. The operation is supported in hardware on both x86(64) and ARM.

This PR renames `i8x16.narrow_i16x8_u` to `i8x16.narrow_u16x8_u` and `i16x8.narrow_i32x4_u` to `i16x8.narrow_u32x4_u`, highlighting the unsigned-to-unsigned semantics; it adds two new opcodes for `i8x16.narrow_i16x8_u` and `i16x8.narrow_i32x4_u` for unsigned narrowing with signed inputs.

Instructions corresponding to narrowing of signed values to unsigned results are `VQMOVUN` on ARM and `PACKUSDW` on x86; signed to signed narrowing is `VQMOVN` on ARM and `PACKSSDW` on x86; unsigned to unsigned is also `VQMOVN` on ARM (using a different data type) without exact match in x86 SIMD.

For issue #94


dtig:
Thanks for the PR, could you add a comment here with mappings to the exact instructions on different architectures? This information is split right now across multiple threads, so having this here for folks to review may be useful.

penzn:
Instructions corresponding to narrowing of signed values to unsigned results are `VQMOVUN` on ARM and `PACKUSDW` on x86.

For more context on the instructions already added, not in scope of this change, signed to signed narrowing is `VQMOVN` on ARM and `PACKSSDW` on x86; unsigned to unsigned is also `VQMOVN` on ARM (using a different data type) without exact match in x86 SIMD.

penzn:
Ping @dtig @arunetm

dtig:
Given that this has been contentious and the discussion in PR #91, I suggest we put this to vote and resolve this after a reasonable time to vote, possibly in next week's sync meeting. I think the options here are -

 - Include both narrowing operations here, and the ones already merged in the Spec ( 👌)
 - Merge #91, to clarify that the inputs are signed ( 🚀)
 - Exclude all unsigned narrowing operations  ( 👍)

If you have opinions about the semantics of narrowing operations, please vote!

AndrewScheidecker:
FYI I don't see a way to react to your post with 👌. I propose using 👀 instead.

I think there's a clear case for the signed->unsigned narrowing instructions, since there's lots of code out there that uses the corresponding SSE instruction, and NEON has a direct equivalent.

The case for unsigned->unsigned narrowing is less clear. I can imagine it being useful, and hard to reproduce the optimal SSE code-gen for it without a dedicated instruction. It would be nice to have a kernel that benefits from the unsigned->unsigned narrowing instructions to prove that.

I'm leaning toward 👌 (👀) but I'd also be fine with 🚀.

dtig:
Oops, thanks for pointing that out. Updated to use  👀 instead of 👌.

penzn:
I prefer the instruction names from this PR, as they make inputs and outputs more explicit, aside of that I am fine with #91 as well.

Also leaning towards this one 👀, but would not oppose removing purely unsigned instructions (with renaming) 🚀

penzn:
@dtig looks like I had some issues with opcodes, pushed an extra commit and updated the top comment. Yes, if we are merging #91, renaming should not be necessary (as there would be no unsigned inputs), just the semantics would change.

penzn:
Closing as we merged the alternative in #91

Issue #23, opened 2017-11-05
Load with extension operation


Maratyszcza:
There is no efficient way to represent loading of narrow-type vector with extension to wide-type vector, e.g. **Load 4 uint16_t values and extend to 4 x uint32_t vector**. To simulate such operation with the current API, we'd need to load values as a 64-bit scalar (potentially spilling to two registers on 32-bit architectures), transfer to SIMD register (expensive!), and then use shuffles to get it into proper places. With the native SIMD ISA, it can be implemented more efficiently:

- `PMOVZXWD xmm, [mem]` on x86 with SSE4.1
- `MOVQ xmm, [mem] + PXOR xmm0, xmm0 + PUNPCKLWD xmm, xmm0` on SSE2
- `VLD1.16 {dX}, [rAddr] + VMOVL.U16 qX, dX` on ARMv7+NEON
- `LD1 {Vx.4H}, xAddr + UXTL Vx.4S, Vx.4H` on ARM64

brion:
I've run into this doing vectorizing on the dav1d video codec (expanding i8 to i16 lanes); shuffles work but seem awkward.

In theory the runtime could detect the shuffle pattern as an interleave/de-interleave and optimize it, I guess, but I'm not sure I want to rely on that.

dtig:
Closing as #98 is merged.

PR #98, opened 2019-08-28
Introduce Load and Extend

penzn:
Rebasing #77 on current master and incorporating the latest review feedback.

This change proposes six new load instructions, that would combine memory read of "half-size" vector with extending each lane to the next standard lane size. Motivating workloads are machine learning, image compression, video rendering, and data processing. There is widespread hardware support. Also see #23, #28, #77

Hardware support from #23:

* `PMOVZXWD xmm, [mem]` on x86 with SSE4.1
* `MOVQ xmm, [mem] + PXOR xmm0, xmm0 + PUNPCKLWD xmm, xmm0` on SSE2
* `VLD1.16 {dX}, [rAddr] + VMOVL.U16 qX, dX` on ARMv7+NEON
* `LD1 {Vx.4H}, xAddr + UXTL Vx.4S, Vx.4H` on ARM64

### DIFF
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -109,7 +109,6 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
 | `i8x16.sub`                |    `0x5a`| -                  |
 | `i8x16.sub_saturate_s`     |    `0x5b`| -                  |
 | `i8x16.sub_saturate_u`     |    `0x5c`| -                  |
-| `i8x16.mul`                |    `0x5d`| -                  |
 | `i16x8.neg`                |    `0x62`| -                  |
 | `i16x8.any_true`           |    `0x63`| -                  |
 | `i16x8.all_true`           |    `0x64`| -                  |
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -86,7 +86,6 @@
 | `i8x16.sub`                |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `i8x16.sub_saturate_s`     |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `i8x16.sub_saturate_u`     |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
-| `i8x16.mul`                |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `i16x8.neg`                |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `i16x8.any_true`           |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `i16x8.all_true`           |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -367,7 +367,6 @@ def S.sub(a, b):
 ```
 
 ### Integer multiplication
-* `i8x16.mul(a: v128, b: v128) -> v128`
 * `i16x8.mul(a: v128, b: v128) -> v128`
 * `i32x4.mul(a: v128, b: v128) -> v128`
 


penzn:
pinging @dtig, @tlively, @Maratyszcza, and @arunetm

tlively:
Looks fine to me from a tooling perspective. @ngzhian would the renumbering of the narrowing and widening operations here be problematic?

proposals/simd/ImplementationStatus.md
```diff
@@ -145,6 +145,12 @@
...
 | `v8x16.shuffle`            |                           |                    | :heavy_check_mark: | :heavy_check_mark: |
+| `i16x8.load8x8_u`          |                           |                    |                    |                    |
+| `i16x8.load8x8_s`          |                           |                    |                    |                    |
+| `i32x4.load16x4_u`         |                           |                    |                    |                    |
+| `i32x4.load16x4_s`         |                           |                    |                    |                    |
+| `i64x2.load32x2_u`         |                           |                    |                    |                    |
+| `i64x2.load32x2_s`         |                           |                    |                    |                    |
 | `i8x16.narrow_i16x8_s`     |                           | :heavy_check_mark: | :heavy_check_mark: |                    |
 | `i8x16.narrow_i16x8_u`     |                           | :heavy_check_mark: | :heavy_check_mark: |                    |
```
tlively: I wonder if we should clear the checkmarks for all tools for renumbered instructions. Thoughts?

penzn: Good question. Those used to work, maybe change them to question marks if GH markdown has those?

We should do that outside of this PR though, IMO :)  I'd be happy to help.

tlively: I think it would make sense to do this change in the same PR since the old information becomes outdated the moment this gets merged, but I don't feel strongly about it. `:question:` (:question:) or `:grey_question:` (:grey_question:) would work well.

proposals/simd/BinarySIMD.md
```diff
@@ -172,15 +172,21 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
...
-| `i32x4.widen_low_i16x8_s`  |    `0xce`| -                  |
-| `i32x4.widen_high_i16x8_s` |    `0xcf`| -                  |
-| `i32x4.widen_low_i16x8_u`  |    `0xd0`| -                  |
-| `i32x4.widen_high_i16x8_u` |    `0xd1`| -                  |
+| `i16x8.load8x8_u`          |    `0xc6`| m:memarg           |
+| `i16x8.load8x8_s`          |    `0xc7`| m:memarg           |
+| `i32x4.load16x4_u`         |    `0xc8`| m:memarg           |
+| `i32x4.load16x4_s`         |    `0xc9`| m:memarg           |
+| `i64x2.load32x2_u`         |    `0xca`| m:memarg           |
+| `i64x2.load32x2_s`         |    `0xcb`| m:memarg           |
```
ngzhian: V8 has implementated these using the older opcodes, see https://cs.chromium.org/chromium/src/v8/src/wasm/wasm-opcodes.h?l=418-429&rcl=e576e7591361f0f32a9b02c8f67a63ef1963a33d
Any reason why you put the loads before narrow/widen? If they go after the narrow/widen then v8 won't need any change. Otherwise it's not hard to change v8's implementation too.

penzn: The idea is that they would be next to load and splat instructions, but I don't feel very strongly about that (load and plats are not next to regular loads, for example). @tlively, @dtig, what do you think?

proposals/simd/ImplementationStatus.md
penzn: Now I understand what you meant. That would work, or I can also move new instruction to the bottom of the list, that way we would not need to invalidate anything. I put extended loads next to load+splats, for consistency sake, but that can be debated.

proposals/simd/BinarySIMD.md
tlively: I think we should throw them in wherever for now without thinking about aesthetics too much. I expect we will do a final renumbering later in the standardization process (maybe just before someone decides it's time to ship) once we have a better idea of what will make it into MVP SIMD and possibly what will be in the next iteration of SIMD.

proposals/simd/ImplementationStatus.md
penzn: Moved new opcodes to the bottom of the list, now we would not need to invalidate renumbered instructions.

Maratyszcza:
LGTM

dtig:
Thanks for distilling multiple PRs/Issues into this one, I'm in favor of merging this because collaborators from different application domains have indicated that this would be a useful operation to have, this has been a requested addition since 2017 and seems to be reasonably well supported on all architectures. A couple of outstanding things to make sure we handle previous concerns. 
 - There were some concerns about this set of operations when presented previously, but IIRC load+extend operations as enumerated here would be okay to include. cc @sunfishcode 
 - This was originally proposed in conjunction with removal of i8x16.mul because of the large number of overflow cases, but this PR doesn't seem to address that, is the motivation still to remove the i8x16.mul operation?

arunetm:
There was general consensus on removing i8x16.mul. @penzn can we update this PR to address it as well.

penzn:
Do you mean the [discussion][d] in#28? Will add a commit to remove those if there is no objections.

[d]: https://github.com/WebAssembly/simd/issues/28#issuecomment-377585364

ngzhian:
Sorry I'm late to point this out:

`VLD1.16 {dX}, [rAddr]` doesn't allow for load via base + index (both in registers).
According to the manual, ARM DDI 0487D.b C7-1629, it only allows for a post index addressing mode, or an immediate offset.
So it will need a temporary add of base + offset, before passing the final memory address to VLD1.
Am I right in this interpretation of VLD1 instruction?

Maratyszcza:
You are right that `VLD1` instructions don't support addressing with offset. There is also an option of using `VLDR` instruction, which support immediate (but **not** register) offset from a register base.

ngzhian:
Thanks, I think I'll implement it using add to temporary first, and then add the optimization to use VLDR if the offset can be an immediate.

penzn:
@ngzhian are you working on implementing that in V8? We wanted to get some timings for this, can lend a hand with implementation.

ngzhian:
Yup I am, I have done it all for x64, so if you are interested only for x64 you can build locally and run. I have not started on arm/arm64/ia32 yet, so if you want to pick those up, lmk!

penzn:
I believe @rrwinterton was interested in testing this. Arm - probably not, let us think about ia32, could be somebody else would be willing to take it as well.

ngzhian:
Sounds good, I'll be working on arm/arm64 soon, and will leave ia32 to yall for now, please keep me updated (here or via email) so we don't overlap :) Thanks!

Issue #60, opened 2018-12-18
Spec tests


tlively:
When it comes time to write up SIMD spec tests, it will be useful to be aware of the tests that exist in other places.

Binaryen: https://github.com/WebAssembly/binaryen/blob/master/test/spec/simd.wast
Wabt: https://github.com/WebAssembly/wabt/tree/master/test/interp simd-*.txt

AndrewScheidecker:
I also wrote some pretty limited tests for WAVM [here](https://github.com/WAVM/WAVM/blob/master/Test/simd.wast).

Trying the binaryen and wabt tests in WAVM I noticed a few incompatibilities:

* wabt implements SIMD compares to return 1 for true comparisons, but the proposed spec is to return -1 for true comparisons. Binaryen (and now WAVM) return -1 for true comparisons.
* Neither the Binaryen or wabt tests use the [proposed](https://github.com/WebAssembly/simd/blob/master/proposals/simd/TextSIMD.md) `v128.const` syntax.
* wabt has a `v8x16.shuffle` [test](https://github.com/WebAssembly/wabt/blob/master/test/interp/simd-lane.txt#L88) that uses a 4xi32 immediate. WAVM and Binaryen use a 16xi5 syntax.

gnzlbg:
> wabt implements SIMD compares to return 1 for true comparisons, but the proposed spec is to return -1 for true comparisons. 

That's a bug, I've sent a PR that fixes that: https://github.com/WebAssembly/wabt/pull/1031

tlively:
> * wabt has a `v8x16.shuffle` [test](https://github.com/WebAssembly/wabt/blob/master/test/interp/simd-lane.txt#L88) that uses a 4xi32 immediate. WAVM and Binaryen use a 16xi5 syntax.

I submitted #60 to standardize this.

gnzlbg:
> wabt has a v8x16.shuffle test that uses a 4xi32 immediate. WAVM and Binaryen use a 16xi5 syntax.

I've implemented that (hopefully correctly) in WABT https://github.com/WebAssembly/wabt/pull/1032

tlively:
@gnzlbg Thanks for your work on this! Are you planning to update WABT's v128.const syntax as well? No worries if not.

gnzlbg:
> Are you planning to update WABT's v128.const syntax as well? 

[...]

PR #99, opened 2019-09-04
Adding new spec tests

arunetm:
Moving the initial set of SIMD spec tests verified through WAVM CI.
This helps implementations to begin using these tests for validation.


arunetm:
https://github.com/WebAssembly/simd/issues/60 
@tlively @Honry @abrown

AndrewScheidecker:
Can you include the Python scripts used to generate some of these files? https://github.com/WAVM/WAVM/tree/master/Test/spec/simd/meta

For reference, the bulk-memory-operations proposal also uses Python to generate some test files, and includes those scripts in [test/meta](https://github.com/WebAssembly/bulk-memory-operations/tree/master/test/meta).

arunetm:
Yes, I think its fine to include the scripts. Will add.

PR #102, opened 2019-09-10
ANDNOT operation

Maratyszcza:
Introduction
=========

ANDNOT is a widely supported SIMD operation which computes `a & ~b`. ANDNOT is involved in a common idiom of zeroing out elements which don't satisfy a condition, i.e.

```c
float x = ...;
const bool cond = ...;
if (!cond) x = 0;
```

In the present SIMD instruction set, a vectorized version of this snippet would require two WAsm SIMD instructions, `v128.and` and `v128.not`:

```c
v128 x = ...;
const v128 cond = ...;
x = v128.and(x, v128.not(cond));
```

Representing ANDNOT as two instructions is inefficient on all architectures:

- On ARM, ARM64, and PowerPC it requires two machine instructions even though ANDNOT has an exact equivalent in their SIMD instruction sets.
- On x86 and x86-64 it requires three machine instructions because x86 SIMD extensions (until AVX512) do not include SIMD `NOT` operation. Thus, WAsm engine would typically generate three instructions: two (`PXOR tmp_zero, tmp_zero` to zero a temporary register and `PANDNOT b, tmp_zero` **(!)** to emulate `v128.not`), and an extra `PAND`.

This PR introduce combined ANDNOT instruction to enable WebAssembly engines to directly leverage architecture-specific ANDNOT instructions without doing complicated and expensive analysis of the instruction stream

===========================

This section illustrates how the new WebAssembly instruction can be lowered on common instruction sets. However, these patterns are provided only for convenience, compliant WebAssembly implementations do not have to follow the same code generation patterns.

---------------------------------------------

- `c = v128.andnot(a, b)` maps to `VANDNPS xmm_c, xmm_b, xmm_a` (note the **inverted order of operands**)

x86/x86-64 processors with SSE instruction set
---------------------------------------------

- `c = v128.andnot(a, b)` maps to `MOVAPS xmm_c, xmm_b` + `ANDNPS xmm_c, xmm_a` (note the **inverted order of operands**)
- `b = v128.andnot(a, b)` maps to `ANDNPS xmm_b, xmm_a` (note the **inverted order of operands**)

ARMv7+ processors with NEON instruction set
---------------------------------------------

- `c = v128.andnot(a, b)` maps to `VBIC Qc, Qa, Qb`

ARM64 processors
--------------------------

- `c = v128.andnot(a, b)` maps to `BIC Vc.16B, Va.16B, Vb.16B`

POWER processors with Vector facility (VMX)
--------------------------------------

- `c = v128.andnot(a, b)` maps to `VANDC Vc, Va, Vb`

### DIFF
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -189,3 +189,4 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
 | `i32x4.load16x4_u`         |    `0xd5`| m:memarg           |
 | `i64x2.load32x2_s`         |    `0xd6`| m:memarg           |
 | `i64x2.load32x2_u`         |    `0xd7`| m:memarg           |
+| `v128.andnot`              |    `0xd8`| -                  |
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -71,6 +71,7 @@
 | `f64x2.ge`                 | `-munimplemented-simd128` |                    | :heavy_check_mark: | :heavy_check_mark: |
 | `v128.not`                 |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `v128.and`                 |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
+| `v128.andnot`              |                           |                    |                    |                    |
 | `v128.or`                  |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `v128.xor`                 |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `v128.bitselect`           |               `-msimd128` | :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -521,6 +521,12 @@ The logical operations defined on the scalar integer types are also available
 on the `v128` type where they operate bitwise the same way C's `&`, `|`, `^`,
 and `~` operators work on an `unsigned` type.
 
+### Bitwise AND-NOT
+
+* `v128.andnot(a: v128, b: v128) -> v128`
+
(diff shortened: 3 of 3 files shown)


dtig:
The NOT operation being sub optimal on x86 and x86-64 stood out during implementation, and having a more optimal ANDNOT operation when supported across architectures would make sense to me. After preliminary research, it doesn't look like any of the hardware instructions these operations map to have performance issues. That said, I'd like to make sure that the operations this is widely used. Could you add examples where this is already in use? Anything that makes use of the equivalent XMM/Neon intrinsics would be helpful.

Maratyszcza:
GitHub search shows [127K+ files using `_mm_andnot_ps`](https://github.com/search?l=C%2B%2B&q=_mm_andnot_ps&type=Code) (x86 instrinsic for ANDNOT) and [40K+ C++ files using `vbicq_u32`](https://github.com/search?l=C%2B%2B&q=vbicq_u32&type=Code) (ARM NEON intrinsic for ANDNOT). From my experience, ANDNOT is useful when implementing vector mathematical functions, e.g. `expf` must return +0.0f when input is below `-0x1.9FE368p+6f`:
```c
y = _mm_andnot_ps(_mm_cmpgt_ps(x, _mm_set1_ps(-0x1.9FE368p+6f)), x);
```
Note: inverting comparison and using `_mm_and_ps` wouldn't work, because it change result for NaN values.

dtig:
Thanks @Maratyszcza! I'm in favor of merging this because this is not introducing a class of new operations, but a more efficient combination of operations that are already supported by this proposal, and have an exact match on relevant architectures, and is widely used as pointed out above. Are there any objections to including this in the proposal?

proposals/simd/BinarySIMD.md
```diff
@@ -30,6 +30,7 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
 | `v128.load`                |    `0x00`| m:memarg           |
 | `v128.store`               |    `0x01`| m:memarg           |
 | `v128.const`               |    `0x02`| i:ImmByte[16]      |
+| `v128.andnot`              |    `0x03`| -                  |
```
dtig: Could you move this down to the end of the opcode space instead of here? This belongs with the other logical operations, but there isn't a good spot for it right now. It's what we've tried to do with other new opcodes when it doesn't fit into the current opcode space.

Maratyszcza:
Done: added an entry in `ImplementationStatus.md` and moved to the end of the opcode list.

PR #106, opened 2019-09-13
Fix pseudocode, was causing syntax highlight issues

ngzhian:


proposals/simd/SIMD.md
```diff
@@ -300,7 +300,7 @@ def S.shuffle(a, b, s):
         if s[i] < S.lanes:
             result[i] = a[s[i]]
         else:
-            result[i] = b[s[i] - S.lanes
+            result[i] = b[s[i]] - S.lanes
```
penzn: I think that closing brace should be at the end of the line, it is trying to normalize the index

ngzhian: Thanks for catching this!

PR #107, opened 2019-09-13
Add Python pseudocode to memory operations

penzn:
Introducing pseudocode for memory operations.

### DIFF
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -667,7 +667,7 @@ Load a `v128` vector from the given heap address.
 
 ```python
 def S.load(memarg):
-    return S.New(memory.bytes[memarg.start:memarg.start + 16])
+    return S.from_bytes(memory[memarg.start:memarg.start + 16])
 ```
 
 ### Load and Splat
@@ -682,9 +682,9 @@ Load a single element and splat to all lanes of a `v128` vector.
 ```python
 def S.load_splat(memarg):
     result = S.New()
-    val = memory.bytes[memarg.start:memarg.start + S.LaneBytes])
+    val = memory[memarg.start:memarg.start + S.LaneBytes])
     for i in range(S.Lanes):
-        S.bytes[i:i + S.LaneBytes] = val
+        result[i] = S.LaneType.from_bytes(val)
     return result
 ```
 ### Load and Extend
@@ -701,9 +701,9 @@ Fetch consequtive integers up to 32-bit wide and produce a vector with lanes up 
 ```python
 def S.load_extend(ext, memarg):
     result = S.New()
-    bytes = memory.bytes[memarg.start:memarg.start + 8])
+    bytes = memory[memarg.start:memarg.start + 8])
     for i in range(S.Lanes):
-        result[i] = ext(S.LaneType.from_byte(bytes[(i * S.LaneBytes/2):((i+1) * S.LaneBytes/2)]))
+        result[i] = ext(S.LaneType.from_bytes(bytes[(i * S.LaneBytes/2):((i+1) * S.LaneBytes/2)]))
     return result
 
 def S.load_extend_s(memarg):
@@ -721,7 +721,7 @@ Store a `v128` vector to the given heap address.
 
 ```python
 def S.store(memarg, a):
-    memory.bytes[memarg.start:memarg.start + 16] = S.bytes
+    memory[memarg.start:memarg.start + 16] = bytes(a)
 ```
 
 ## Floating-point sign bit operations
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -685,6 +685,7 @@ def S.load_splat(memarg):
     val = memory.bytes[memarg.start:memarg.start + S.LaneBytes])
     for i in range(S.Lanes):
         S.bytes[i:i + S.LaneBytes] = val
+    return result
 ```
 ### Load and Extend
 
@@ -696,6 +697,21 @@ def S.load_splat(memarg):
 * `i64x2.load32x2_s(memarg) -> v128`: load two 32-bit integers and sign extend each one to a 64-bit lane
 
 Fetch consequtive integers up to 32-bit wide and produce a vector with lanes up to 64 bits.
+
+```python
+def S.load_extend(ext, memarg):
+    result = S.New()
+    bytes = memory.bytes[memarg.start:memarg.start + 8])
+    for i in range(S.Lanes):
+        result[i] = ext(S.LaneType.from_byte(bytes[(i * S.LaneBytes/2):((i+1) * S.LaneBytes/2)]))
+    return result
+
+def S.load_extend_s(memarg):
+    return S.load_extend(Sext, memarg)
+
+def S.load_extend_u(memarg):
+    return S.load_extend(Zext, memarg)
+```
 
 ### Store
 
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -667,7 +667,7 @@ Load a `v128` vector from the given heap address.
 
 ```python
 def S.load(memarg):
-    return S.from_bytes(memory[memarg.start:memarg.start + 16])
+    return S.from_bytes(memory[memarg.offset:memarg.offset + 16])
 ```
 
 ### Load and Splat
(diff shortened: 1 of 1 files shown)


penzn:
@tlively @ngzhian what do you think about this style?

proposals/simd/SIMD.md
```diff
@@ -665,6 +665,11 @@ natural alignment.
 
 Load a `v128` vector from the given heap address.
 
+```python
+def S.load(memarg):
+    return S.New(memory.bytes[memarg.start:memarg.start + 16])
```
tlively: `S.New` is still ambiguous, so how about using a for loop over the lanes using `S.LaneType.Load`?

proposals/simd/SIMD.md
```diff
@@ -674,6 +679,14 @@ Load a `v128` vector from the given heap address.
 
 Load a single element and splat to all lanes of a `v128` vector.
 
+```python
+def S.load_splat(memarg):
+    result = S.New()
+    val = memory.bytes[memarg.start:memarg.start + S.LaneBytes])
+    for i in range(S.Lanes):
+        S.bytes[i:i + S.LaneBytes] = val
```
tlively: Missing a multiply by lane size here. Could also just use `result[i] = val`.

proposals/simd/SIMD.md
ngzhian: Dumb suggestion, let's split `S.new()` and the actual loading of memory into S. We seem to have a style going on here where we do:
```
result = S.New()
for i in ???:
  do operation
```

So maybe this can look like
```
def S.load(memarg):
  result = S.new()
  for i in range(16):
    result[i] = memory.bytes[memarg.start + i]
```

Otherwise we might want to define what `S.New(*args)` means?

proposals/simd/SIMD.md
```diff
@@ -674,6 +679,14 @@ Load a `v128` vector from the given heap address.
...
+    return result
```
ngzhian: Why are we using `S.bytes[i:i + S.LaneBytes] = val`, and not `result[i]`?

Also, do we want to result S.splat?
```
def S.load_splat(memarg):
    val = memory.bytes[memarg.start:memarg.start + S.LaneBytes])
   return S.splat(val)
```

proposals/simd/SIMD.md
```diff
@@ -685,12 +698,32 @@ Load a single element and splat to all lanes of a `v128` vector.
...
+
 ### Store
 
 * `v128.store(memarg, data: v128)`
 
 Store a `v128` vector to the given heap address.
 
+```python
+def S.store(memarg, a):
+    memory.bytes[memarg.start:memarg.start + 16] = S.bytes
```
ngzhian: We might want to define what `S.bytes` mean?

proposals/simd/SIMD.md
```diff
@@ -685,12 +698,32 @@ Load a single element and splat to all lanes of a `v128` vector.
 
 Fetch consequtive integers up to 32-bit wide and produce a vector with lanes up to 64 bits.
 
+```python
+def S.load_extend(ext, memarg):
+    result = S.New()
+    bytes = memory.bytes[memarg.start:memarg.start + 8])
+    for i in range(S.Lanes):
+        result[i] = ext(S.LaneType.from_byte(bytes[(i * S.LaneBytes/2):((i+1) * S.LaneBytes/2)]))
```
tlively: Maybe something like `T.Load(memory, ...)` to make it clear that the semantics are the same as normal loads for value type `T`? I suppose `T` would have to be passed as an argument along with `ext`.

proposals/simd/SIMD.md
penzn: I agree, maybe not the best idea to pass memory to `New`, will change that. It looks like other operations use `result[i]` for individual lanes, I am thinking of using something like `result.bytes[]` for binary representation of the vector value.

proposals/simd/SIMD.md
penzn: I guess we might need a better way to describe that, since `S` seems to be an array of lane values in other cases.

proposals/simd/SIMD.md
penzn: Decided to use `from_bytes` for instantiating `S`, this works with some of Python types.

proposals/simd/SIMD.md
penzn: `bytes()` seems to be a the Python function to convert something to a byte array, that is probably better than `S.bytes`

proposals/simd/SIMD.md
penzn: Good idea, fixed

proposals/simd/SIMD.md
penzn: The idea to have a single load, which gets broken up into lane values (though multiple loads should achieve the same binary result), that seems to be true-er to what actually would happen and also makes this snippet similar to snippets on other memory ops, which I feel is more instructive.

proposals/simd/SIMD.md
penzn: Yeah, that is good point, changing to `from_bytes`, which seems to be a standard-ish way to do this in Python

proposals/simd/SIMD.md
```diff
@@ -674,6 +679,12 @@ Load a `v128` vector from the given heap address.
 
 Load a single element and splat to all lanes of a `v128` vector.
 
+```python
+def S.load_splat(memarg):
+    val = memory[memarg.start:memarg.start + S.LaneBytes])
+    return S.splat(S.LaneType.from_bytes(val))
```
ngzhian: I don't think we need `S.LaneType.from_bytes(val)`, val is already a value of the right type (since we load S.LaneBytes of it). And the `S` and `S.splat` has the right lane type associated.

penzn: I assume that `memory` is a byte array (maybe it is better to call it `linearMemory` or something like that), that's why `val` is a slice of a byte array.  `splat` operation defined in the document would not work on it. I guess it would not hurt to make this a bit more transparent.

penzn:
Sorry, I lost track of this. I was under impression that alignment hint should not affect the address, that's why only offset is used in this PR. Alignment has seemed a bit of a murky area to me, any clarity would be greatly appreciated :)

From docs on [webassembly.org][d]:

> If the effective address of a memory access is a multiple of the alignment attribute value of the memory access, the memory access is considered aligned, otherwise it is considered misaligned. Aligned and misaligned accesses have the same behavior.

Also, similar language in current [spec][s] on github:

> The alignment \memarg.\ALIGN in load and store instructions does not affect the semantics. It is an indication that the offset \X{ea} at which the memory is accessed is intended to satisfy the property \X{ea} \mod 2^{\memarg.\ALIGN} = 0. A WebAssembly implementation can use this hint to optimize for the intended use.

[d]: https://webassembly.org/docs/semantics/#alignment
[s]: https://github.com/WebAssembly/spec/blob/9fd0547e536b569448bbc116e8ebee05a704240e/document/core/exec/instructions.rst#memory-instructions

tlively:
Right, alignment hints are part of the `memarg` but do not change semantics in any way. I'm not sure it's necessary to explicitly mention them since memargs are already specified, but perhaps linking to the relevant parts of the spec would be helpful (https://webassembly.github.io/spec/core/syntax/instructions.html#syntax-memarg, https://webassembly.github.io/spec/core/binary/instructions.html#binary-memarg). It would also be good to mention somewhere that the natural alignment for v128 is 16 bytes.

To keep the naming consistent, the pseudocode should use `memarg.offset` rather than `memarg.start`. I believe that's what @dtig was getting at as well.

penzn:
Good point on offset, changed that.

dtig:
Sorry I wasn't clear in my earlier comment, I did mean that memarg.offset should be used instead of memory start as it's confusing to use terminology not a part of memarg, looks like that's fixed now so that part of the PR lgtm. By accounting for alignment as well, I meant that if the specified alignment is larger than the natural alignment of s128 type, that will be a [validation failure](https://webassembly.github.io/spec/core/valid/instructions.html#memory-instructions), and pseudo code should probably account for that? I can see how that's probably out of scope for this PR as the rest of the text in SIMD.md doesn't talk about validation failures right now so approving. But this is more noticeable in pseudocode as that is more explicit about the operation.

binji:
The semantics here are not quite right, `memarg` is an immediate for a load/store operation, and has an immediate offset. The instruction itself also has an `i32` operand which must be included in the address calculation.

PR #115, opened 2019-10-01
Update 64x2 instructions

### LINKED ISSUE #101: Re-introduce some 64x2 instructions
In #81, we wanted some performance data to support the inclusion of 64x2 instructions. In https://github.com/ngzhian/simd-benchmarks we documented steps to run benchmarks, and the results of running those benchmarks.

I would like to suggest the inclusion of 64x2 instructions back into the spec because we see real world workloads that benefit from these 64x2 instructions.

However, we don't see all 64x2 instructions getting the same usage and exposure, as such, we might want to consider including only a subset of these instructions.

Option 1, all 64x2 instructions :eyes: 
- i64x2.{splat,extract_lane,replace_lane}
- i64x2.{add,sub,mul,neg}
- i64x2.{shl,shr_s,shr_u}
- i64x2.{any_true,all_true}
- i64x2.{eq,le,lts,ltu,gts,gtu,les,leu,ges,geu}
- i64x2.{abs,min,max}
- i64x2.trunc_{s,u}/f64x2
- f64x2.{splat,extract_lane,replace_lane}
- f64x2.{add,sub,mul,neg}
- f64x2.{div,sqrt}
- f64x2.{shl,shr_s,shr_u}
- f64x2.{any_true,all_true}
- f64x2.{eq,le,lt,ge,gt}
- f64x2.{abs,min,max}
- f64x2.convert_{s,u}/i64x2

Option 2, only f64x2 instructions :rocket: 
- all the f64x2 instructions (including conversions) from Option 1

Option 3, f64x2 instructions and common i64x2 instructions :tada: 
- all the f64x2 instructions (including conversions) from Option 1
- i64x2.{splat,extract_lane,replace_lane}
- i64x2.{add,sub,mul,neg}
- i64x2.{shl,shr_s,shr_u}

I can put up a PR for changes to the docs when one of these options is decided.

Feel free to vote on this issue, eyes :eyes:  for 1, rocket :rocket:  for 2, tada :tada:  for 3.

ngzhian:
A poll was held at CG meeting today, the consensus was for Option 3 of
https://github.com/WebAssembly/simd/issues/101.

This means we have all the f64x2 instructions, and some of the more
common i64x2 that have been benchmarked. This change reflects that
option in the proposal text, binary text, and implementation status.

Fixes #101.

### DIFF
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -132,12 +132,11 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
 | `i32x4.sub`                |    `0x7c`| -                  |
 | `i32x4.mul`                |    `0x7f`| -                  |
 | `i64x2.neg`                |    `0x84`| -                  |
-| `i64x2.any_true`           |    `0x85`| -                  |
-| `i64x2.all_true`           |    `0x86`| -                  |
 | `i64x2.shl`                |    `0x87`| -                  |
 | `i64x2.shr_s`              |    `0x88`| -                  |
 | `i64x2.shr_u`              |    `0x89`| -                  |
 | `i64x2.add`                |    `0x8a`| -                  |
+| `i64x2.mul`                |    `0x8c`| -                  |
 | `i64x2.sub`                |    `0x8d`| -                  |
 | `f32x4.abs`                |    `0x95`| -                  |
 | `f32x4.neg`                |    `0x96`| -                  |
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -110,13 +110,12 @@
 | `i32x4.sub`                |               `-msimd128` |    :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `i32x4.mul`                |               `-msimd128` |    :heavy_check_mark: | :heavy_check_mark: | :heavy_check_mark: |
 | `i64x2.neg`                | `-munimplemented-simd128` |                       | :heavy_check_mark: | :heavy_check_mark: |
-| `i64x2.any_true`           | `-munimplemented-simd128` |                       | :heavy_check_mark: | :heavy_check_mark: |
-| `i64x2.all_true`           | `-munimplemented-simd128` |                       | :heavy_check_mark: | :heavy_check_mark: |
 | `i64x2.shl`                | `-munimplemented-simd128` |                       | :heavy_check_mark: | :heavy_check_mark: |
 | `i64x2.shr_s`              | `-munimplemented-simd128` |                       | :heavy_check_mark: | :heavy_check_mark: |
 | `i64x2.shr_u`              | `-munimplemented-simd128` |                       | :heavy_check_mark: | :heavy_check_mark: |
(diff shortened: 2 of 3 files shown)


tlively:
Does the V8 implementation status need to be updated as well, or do we still have unimplemented 64x2 instructions?

ngzhian:
Good point, they are only implemented for x64 and arm64 (and some for ia32), so I'm not sure if it counts as implemented or not.

dtig:
For V8 implementation status, it's fully supported when the operations are implemented on all (officially) supported architectures to make sure that a binary that works on one platform does not fail decode with an unimplemented opcode on a different platform.

ngzhian:
> Shouldn't the i64x2.trunc_sat_* operations be removed too? #101 lists them under Option 1.

I was thinking of these as "conversion" operators, since these are i64x2 -> f64x2.


> For V8 implementation status, 

Thanks for clarifying, in that case the implementation status should be left empty.

proposals/simd/BinarySIMD.md
```diff
@@ -132,12 +132,11 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
 | `i32x4.sub`                |    `0x7c`| -                  |
 | `i32x4.mul`                |    `0x7f`| -                  |
 | `i64x2.neg`                |    `0x84`| -                  |
-| `i64x2.any_true`           |    `0x85`| -                  |
-| `i64x2.all_true`           |    `0x86`| -                  |
 | `i64x2.shl`                |    `0x87`| -                  |
 | `i64x2.shr_s`              |    `0x88`| -                  |
 | `i64x2.shr_u`              |    `0x89`| -                  |
 | `i64x2.add`                |    `0x8a`| -                  |
+| `i64x2.mul`                |    `0x8c`| -                  |
```
AndrewScheidecker: I thought we were reserving `0x90` for `i64x2.mul`.
`i8x16.mul`(`0x5d`) = `i8x16.neg`(`0x51`) + 12
`i16x8.mul`(`0x6e`) = `i16x8.neg`(`0x62`) + 12
`i32x4.mul`(`0x7f`) = `i32x4.neg`(`0x73`) + 12
`i64x2.mul`(`0x8c`) = `i64x2.neg`(`0x84`) + 8?

ngzhian: Good point, I don't think these are final yet, we will look at the actual opcodes again once we have a complete set. There will probably be reordering required.

tlively: We were pretty intentional in laying out the opcode space about a year ago before we started adding new instructions, so I think we should respect those decisions where possible.

ngzhian: I'm happy to change the opcode of i64x2.ul, but with new instructions like the load_splat and load_extend, the space of i8x16 and the space of i64x2 will not be the same size anymore, and this nice equation mapping *.neg to *.mul will conflict with the goal of keeping all the i8x16 instructions together (supposing there is such a goal).

ngzhian: https://github.com/WebAssembly/simd/pull/116

tlively: Right, I agree that for new instructions like load_splat, load_extend, narrowing, and widening just putting them at the end of the opcode space for now is the best move.

PR #116, opened 2019-10-04
Update i64x2.mul opcode

ngzhian:


ngzhian:
Yea good suggestion, fixed!

PR #120, opened 2019-10-19
Integrate latest spec tests from WAVM

Honry:
Tests have been verified through WAVM CI.


Honry:

abrown:
Thanks for upstreaming these tests! I can use these over in wasmtime once merged (cc @sunfishcode)

PR #127, opened 2019-10-28
i32x4.dot_i16x8_s instruction

Maratyszcza:
Introduction
=========

Integer arithmetic instructions in WebAssembly SIMD produce results of the same element type as their inputs. To avoid overflow, developers have to pre-extend inputs to wider element types and do (twice more) arithmetic operations on twice wider elements. The need to use this work-around is particularly concerning for multiplications:

- Unlike additions and subtractions, where overflow is an exceptional situation, multiplications normally produce twice wider result than their inputs, and overflowing input type is a common, rather than exceptional, case. Thus nearly every multiplication must be accompanied by pre-extending inputs to wider element types.
- Both on x86 and on ARM, multiplications which produce twice wider results are cheaper (typically by 2x) than multiplications on the wider type itself. The table below quantifies the throughput cost (from [1]) of a combination of `PMULLW` + `PMULHW` instructions (which together compute 32-bit products of 8 16-bit inputs) and a combination of two `PMULLD` instructions (which together compute 32-bit products of 8 32-bit inputs) on various x86 microarchitectures:

Microarchitecture | 2x `PMULLD xmm, xmm` | `PMULLW xmm, xmm` + `PMULHW xmm, xmm` |
-- | :--: | :--: |
AMD Piledriver | 4 | 2 |
AMD Zen | 4 | 2 |
Intel Nehalem | 4 | 2 |
Intel Sandy Bridge | 2 | 2 |
Intel Haswell | 4 | 2 |
Intel Skylake | 2 | 1 |
Intel Goldmont | 4 | 2 |

Operations that produce twice wider results would need to return two SIMD vectors, and thus depend on the future multi-value proposal. To stay within baseline WebAssembly features, we have to aggregate the two wide results into one, so the instruction can produce a single output vector. Luckily, there is an aggregating operation directly supported in x86, MIPS, and POWER instruction sets, that can also be efficiently lowered on ARM and ARM64: addition of adjacent multiplication results. The resulting combination of a full multiplication and addition of adjacent products can be interpreted as a dot product of 2-wide subvectors within a SIMD vector, producing twice wider results than input vectors (albeit twice fewer result elements than input elements). 

This PR introduce the 2-element dot product instructions with signed 16-bit integer input elements and signed 32-bit integer output elements. We don't consider other data types, because they can't be efficiently expressed on x86 (e.g. the only multiplication on byte inputs on x86 multiplies signed bytes by unsigned bytes with signed saturation - too exotic to build a portable instruction on top of it). The new `i32x4.dot_i16x8_s` instruction returns the dot product right away, and `i32x4.dot_i16x8_add_s` additionally accumulates it with a third input vector of 32-bit elements. This second instruction was added because accumulation of dot product results is common, and many instruction sets provide a specialized instruction for this case.

[October 31 update] Applications
===========================

Below are examples of optimized libraries using close equivalents of the proposed `i32x4.dot_i16x8_s` and `i32x4.dot_i16x8_add_s` instructions:
- [Discrete Cosine Transform in libjpeg-turbo](https://github.com/libjpeg-turbo/libjpeg-turbo/blob/9a51a87af33026a2befb5abd1fdcd3362286b3dc/simd/x86_64/jfdctint-sse2.asm#L236-L239)
- [Free Lossless Audio Codec](https://github.com/xiph/flac/blob/1b5c09e4c692e243239945be3cba3ec72ea1699f/src/libFLAC/lpc_intrin_sse2.c#L83-L94)
- [Discrete Cosine Transform in VP9 video codec](https://github.com/InteractiviteVideoEtSystemes/libvpx/blob/873e38b72d4910cf59fb6930668930cca0a7dde0/vp9/encoder/x86/vp9_dct_intrin_sse2.c#L89-L97)
- [Sum of Squares in VP9 video codec](https://github.com/InteractiviteVideoEtSystemes/libvpx/blob/5183107cd4762e4834cbee1d3d77e80a0f86ae22/vpx_dsp/x86/sum_squares_sse2.c#L64-L79)
- [Subpixel filtering in VP9 video codec](https://github.com/InteractiviteVideoEtSystemes/libvpx/blob/5183107cd4762e4834cbee1d3d77e80a0f86ae22/vpx_dsp/x86/vpx_subpixel_4t_intrin_sse2.c#L727-L735)
- [(DCT?) Tranformation in SVT-HEVC video codec](https://github.com/OpenVisualCloud/SVT-HEVC/blob/master/Source/Lib/ASM_SSSE3/EbTransforms_Intrinsic_SSSE3.c#L268-L286)
- [Discrete Cosine Transform in de265 video codec](https://github.com/strukturag/libde265/blob/b500546cc40649b9786f4f258ffe3829f1bd59cd/libde265/x86/sse-dct.cc#L364-L369)
- [Sum of Squares in MJPEGTools library](https://github.com/sergiomb2/mjpegtools/blob/01a8142be246e4f4625922845d994462a732f282/utils/altivec/sumsq.c#L244)
- [Discrete Cosine Transform in MJPEGTools library](https://github.com/sergiomb2/mjpegtools/blob/01a8142be246e4f4625922845d994462a732f282/utils/altivec/field_dct_best.c#L132-L135)
- [Discrete Cosine Transform in VPX library](https://github.com/extware/libvpx/blob/802569640731033bd649c2d3b0f065a0ff674010/vpx_dsp/ppc/inv_txfm_vsx.c#L1244-L1248)
- [Packing Matrices for Matrix-Matrix Multiplication in gemmlowp library](https://github.com/google/gemmlowp/blob/6c8f5d41aa27aec992bcb75e1279a4bcea5ba6ea/internal/pack_sse.h#L97-L113)
- [Integer Matrix-Matrix Multiplication in QNNPACK library](https://github.com/pytorch/QNNPACK/blob/901e9d40aeedb2c1b212b28f022fb099ffe617c2/src/q8gemm/4x4c2-sse2.c#L64-L108)

===========================

x86/x86-64 processors with AVX512VNNI and AVX512VL instruction sets

- **i32x4.dot_i16x8_add_s**
  - `c = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `VPDPWSSD xmm_c, xmm_a, xmm_b`
  - `y = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `VMOVDQA xmm_y, xmm_c + VPDPWSSD xmm_c, xmm_a, xmm_b`

x86/x86-64 processors with XOP instruction set

- **i32x4.dot_i16x8_add_s**
  - `y = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `VPMADCSWD xmm_y, xmm_a, xmm_b, xmm_c`

- **i32x4.dot_i16x8_s**
  - `y = i32x4.dot_i16x8_s(a, b)` is lowered to `VPMADDWD xmm_y, xmm_a, xmm_b`
- **i32x4.dot_i16x8_add_s**
  - `y = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `VPMADDWD xmm_tmp, xmm_a, xmm_b + VPADDD xmm_y, xmm_tmp, xmm_c`

- **i32x4.dot_i16x8_s**
  - `a = i32x4.dot_i16x8_s(a, b)` is lowered to `PMADDWD xmm_a, xmm_b`
  - `y = i32x4.dot_i16x8_s(a, b)` is lowered to `MOVDQA xmm_y, xmm_a + PMADDWD xmm_y, xmm_b`
- **i32x4.dot_i16x8_add_s**
  - `c = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `MOVDQA xmm_tmp, xmm_a + PMADDWD xmm_tmp, xmm_b + PADDD xmm_c, xmm_tmp`
  - `y = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `MOVDQA xmm_y, xmm_a + PMADDWD xmm_y, xmm_b + PADDD xmm_y, xmm_c`

ARM64 processors

- **i32x4.dot_i16x8_s**
  - `y = i32x4.dot_i16x8_s(a, b)` is lowered to:
    - `SMULL Vtmp.4S, Va.4H, Vb.4H`
    - `SMULL2 Vtmp2.4S, Va.8H, Vb.8H`
    - `ADDP Vy.4S, Vtmp.4S, Vtmp2.4S `
- **i32x4.dot_i16x8_add_s**
  - `y = i32x4.dot_i16x8_add_s(a, b)` is lowered to:
    - `SMULL Vtmp.4S, Va.4H, Vb.4H`
    - `SMULL2 Vtmp2.4S, Va.8H, Vb.8H`
    - `ADDP Vtmp.4S, Vtmp.4S, Vtmp2.4S `
    - `ADD Vy.4S, Vy.4S, Vtmp.4S`

- **i32x4.dot_i16x8_s**
  - `y = i32x4.dot_i16x8_s(a, b)` is lowered to:
    - `VMULL.S16 Qtmp, Da_lo, Vb_lo`
    - `VMULL.S16 Qtmp2, Da_hi, Db_hi`
    - `VPADD.I32 Dy_lo, Dtmp_lo, Dtmp_hi`
    - `VPADD.I32 Dy_hi, Dtmp2_lo, Dtmp2_hi`
- **i32x4.dot_i16x8_add_s**
  - `y = i32x4.dot_i16x8_add_s(a, b)` is lowered to:
    - `VMULL.S16 Qtmp, Da_lo, Vb_lo`
    - `VMULL.S16 Qtmp2, Da_hi, Db_hi`
    - `VPADD.I32 Dtmp_lo, Dtmp_lo, Dtmp_hi`
    - `VPADD.I32 Dtmp_hi, Dtmp2_lo, Dtmp2_hi`
    - `VADD.I32 Qy, Qy, Qtmp`

POWER processors with VMX (Altivec) instruction set

- **i32x4.dot_i16x8_s**
  - `y = i32x4.dot_i16x8_s(a, b)` is lowered to `VXOR VRy, VRy, VRy + VMSUMSHM VRy, VRa, VRb, VRy`
- **i32x4.dot_i16x8_add_s**
  - `y = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `VMSUMSHM VRy, VRa, VRb, VRc`

MIPS processors with MSA instruction set

- **i32x4.dot_i16x8_s**
  - `y = i32x4.dot_i16x8_s(a, b)` is lowered to `DOTP_S.W Wy, Wa, Wb`
- **i32x4.dot_i16x8_add_s**
  - `c = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `DPADD_S.W Wc, Wa, Wb`
  - `y = i32x4.dot_i16x8_add_s(a, b, c)` is lowered to `MOVE.V Wy, Wc + DPADD_S.W Wy, Wa, Wb`

References
=========

[1] Fog, A. "Instruction tables (2019)." URL: [www.agner.org/optimize/instruction_tables.pdf](http://www.agner.org/optimize/instruction_tables.pdf).

### DIFF
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -149,6 +149,8 @@
 | `i32x4.min_u`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.max_s`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.max_u`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
+| `i32x4.dot_i16x8_s`        |                           |                    |                    |                    |                    |
+| `i32x4.dot_i16x8_add_s`    |                           |                    |                    |                    |                    |
 | `i64x2.neg`                |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i64x2.shl`                |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i64x2.shr_s`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -150,7 +150,6 @@
 | `i32x4.max_s`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.max_u`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.dot_i16x8_s`        |                           |                    |                    |                    |                    |
-| `i32x4.dot_i16x8_add_s`    |                           |                    |                    |                    |                    |
 | `i64x2.neg`                |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i64x2.shl`                |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i64x2.shr_s`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -181,6 +181,7 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
 | `i32x4.min_u`              |    `0xb7`| -                  |
 | `i32x4.max_s`              |    `0xb8`| -                  |
 | `i32x4.max_u`              |    `0xb9`| -                  |
+| `i32x4.dot_i16x8_s`        |    `0xba`| -                  |
 | `i64x2.neg`                |    `0xc1`| -                  |
 | `i64x2.shl`                |    `0xcb`| -                  |
 | `i64x2.shr_s`              |    `0xcc`| -                  |
--- a/proposals/simd/NewOpcodes.md
+++ b/proposals/simd/NewOpcodes.md
@@ -98,13 +98,13 @@
 | i8x16.sub            | 0x71   | i16x8.sub                | 0x91   | i32x4.sub                | 0xb1   | i64x2.sub   | 0xd1   |
 | i8x16.sub_sat_s      | 0x72   | i16x8.sub_sat_s          | 0x92   | ---- sub_sat ----        | 0xb2   | ----        | 0xd2   |
 | i8x16.sub_sat_u      | 0x73   | i16x8.sub_sat_u          | 0x93   | ---- sub_sat ----        | 0xb3   | ----        | 0xd3   |
-| ---- dot ----        | 0x74   | ---- dot ----            | 0x94   | ---- dot ----            | 0xb4   | ----        | 0xd4   |
+| -------------        | 0x74   | -------------            | 0x94   | -------------            | 0xb4   | ----        | 0xd4   |
 | ---- mul ----        | 0x75   | i16x8.mul                | 0x95   | i32x4.mul                | 0xb5   | i64x2.mul   | 0xd5   |
 | i8x16.min_s          | 0x76   | i16x8.min_s              | 0x96   | i32x4.min_s              | 0xb6   | ----        | 0xd6   |
 | i8x16.min_u          | 0x77   | i16x8.min_u              | 0x97   | i32x4.min_u              | 0xb7   | ----        | 0xd7   |
 | i8x16.max_s          | 0x78   | i16x8.max_s              | 0x98   | i32x4.max_s              | 0xb8   | ----        | 0xd8   |
 | i8x16.max_u          | 0x79   | i16x8.max_u              | 0x99   | i32x4.max_u              | 0xb9   | ----        | 0xd9   |
-| ---- avgr_s ----     | 0x7a   | ---- avgr_s ----         | 0x9a   | ---- avgr_s ----         | 0xba   | ----        | 0xda   |
+| ----------------     | 0x7a   | ----------------         | 0x9a   | i32x4.dot_i16x8_s        | 0xba   | ----        | 0xda   |
 | i8x16.avgr_u         | 0x7b   | i16x8.avgr_u             | 0x9b   | ---- avgr_u ----         | 0xbb   | ----        | 0xdb   |
 
 | f32x4 Op        | opcode | f64x2 Op        | opcode |
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -395,6 +395,11 @@ def S.mul(a, b):
     return S.lanewise_binary(mul, a, b)
 ```
 
+### Integer dot product
+* `i32x4.dot_i16x8_s(a: v128, b: v128) -> v128`
+
+Lane-wise multiply signed 16-bit integers in the two input vectors and add adjacent pairs of the full 32-bit results.
+
 ### Integer negation
 * `i8x16.neg(a: v128) -> v128`
 * `i16x8.neg(a: v128) -> v128`


dtig:
Thanks for the detailed write-up, what do you think about paring this down to just including the dot2_s operations and not the dot2add_s operations. The dot2_s operations as detailed here are useful to have, but the dot2add_s operations to me seem out of scope for MVP of the SIMD proposal. 

The i16x8.dot2add_s operation maps directly to an instruction only on x86-64 with XOP enabled, on most other architectures, the codegen would be equivalent to generating an add operation after the dot2_s operation.

tlively:
For the instruction naming, the type prefix should be `i32x4` because those prefixes always specify the output type. Also, there is no need for the `_s` suffix because there are no accompanying unsigned variant. Finally, we have avoided using numerals in instruction names this far and I would like to maintain that. How about using `i32x4.dot` or `i32x4.dot_prod` instead?

Maratyszcza:
@dtig I updated the PR description with the list of codes using `_mm_madd_epi16` (SSE2 intrinsic for `PMADDWD`) or `vec_msum` (VMX/Altivec intrinsic for `VMSUMSHM`). Nearly all use cases for `_mm_madd_epi16` involve `_mm_add_epi32` (SIMD addition) on its result (on POWER all cases involve addition because POWER doesn't provide an instruction *without* accumulation of dot product result). Thus, having "2-wide dot product with addition" instruction would be useful in practice, and deliver speedups at least on some processors:
- The most recent Intel processors with AVX512-VNNI and AVX512-VL, e.g. Cascade Lake and Ice Lake.
- AMD processors with XOP (Bulldozer/Piledriver/Steamroller/Excavator cores)
- All POWER processors with SIMD capabilities
- All MIPS processors with MSA instruction set

Maratyszcza:
@penzn I added a list of applications to the PR description. Compared to using load-with-extend and 32x32->32 multiplications, these "dot product" operations have several performance advantages:
- "Load with extend" doesn't exit on ARM and will be lowered into two instructions
- On most x86 microarchitectures 32x32->32 multiplication is twice more expensive (in throughput) than `PMADDWD`. E.g. Intel Skylake can issue two `PMADDWD` instructions per cycle, but only one `PMULLD` instruction per cycle.
- On ARM Cortex-A72/A73/75/etc, 32x32->32 multiplication is twice more expensive (in throughput) than 16x16->32 multiplication used to simulate `i16x8.dot2`.
- Not only is the 32x32->32 multiplication more expensive, but we'd need twice more 32x32->32 multiplication instructions to do the same number of multiplications as `i16x8.dot2`.

Maratyszcza:
@tlively Good point about `i32x4` output type, I renamed the instructions to `i32x4.dot2_s` and `i32x4.dot2_add_s` accordingly. As for `_s` suffix and `2` in the name, I think it is best to leave them to allow for future extensions:
- AVX512 VNNI and NEON DOT extensions provide instructions for a dot product of 4 8-bit integers with accumulation to 32 bits. Thus, we might need to distinguish between `dot2` and `dot4` in the future.
- It is also possible that we'd want to have *unsigned* variants in the future, and need to distinguish between `dot2_s` and `dot2_u` variants. Unsigned variants wouldn't lower as nicely across all architectures, so I left them out of this proposal.

tlively:
If we may need to differentiate between different input types, how about `i32x4.dot_i16x8_s`?

Maratyszcza:
@tlively Sounds reasonable, there are already instructions with similar names. Updated the commit & PR description.

tlively:
@Maratyszcza Could you add pseudocode for the semantics of these operations as well? I want to make sure I implement the interpreter correctly in Binaryen.

Maratyszcza:
@tlively I don't quite understand the pseudo-code specification in WAsm SIMD, especially given that these "dot product" instructions are the few to have some "horizontal" component in them. You may refer to PMADDWD instruction in Intel architecture manual which is the analog of `i32x4.dot_i16x8_s`.

tlively:
The opcodes in this PR collide with the opcodes used for the {i8x16,i16x8}.avgr_u instructions. Since the averaging instructions have been merged, I will reassign the opcodes for the dot product instructions to 0xdb and 0xdc in the LLVM and Binaryen implementations.

Maratyszcza:
@tlively Renumbered opcodes in the PR similarly to LLVM

bjacob:
I would like to +1 this request, especially the 8bit*8bit accumulating into 32bit flavor as in VNNI. It would be a necessary prerequisite to even consider targeting WebAsm for integer-quantized neural network inference code. Without such an instruction, 8bit quantization of neural networks simply won't provide a meaningful computational advantage over float; people would merely 8bit-quantize to shrink the download size but then dequantize to float for the client side computation.

On the ARM side, the VNNI-equivalent instruction is SDOT/UDOT, available in current commercially available ARM CPUs / Android devices such as the Pixel4 and available in lower-end CPUs as well now (Cortex-A55) so it's a present issue, not future. It is a 4x speed difference today.

(For both points made above, see [this data](this data)).

Example production code (used by TensorFlow Lite using these instructions)  (the machine-encodings are here to support older assemblers).
https://github.com/google/ruy/blob/57e64b4c8f32e813ce46eb495d13ee301826e498/ruy/kernel_arm64.cc#L3066

Maratyszcza:
Updated opcodes to 0xba-0xbb in the new schema. Can we have it implemented in Emscripten & V8?

tlively:
I believe this is already usable in Emscripten + V8 via `__builtin_wasm_dot_s_i32x4_i16x8`. I haven't been adding intrinsics for functions that haven't been merged to the proposal yet to avoid misleading users about what instructions are expected to be available in the final spec.

ngzhian:
Sorry I don't think this is in V8. Someone we dropped the ball here. I'll file a bug and get this going. @dtig Maybe I'm not looking at the right opcodes, can you confirm?

dtig:
I think my original concern here still stands - i.e. for the base MVP proposal we've established the baseline to be SSE4.1/Neon, and in the absence of XOP support on engines will only generate SSE2 mapping. Not opposed to prototyping etc. for the i32x4.dot_i16x8_s operation, but as the i32x4.dot_i16x8_add_s doesn't directly map to underlying hardware without XOP enabled, I'm not convinced that this needs an operation as the code generated would be the same with separate dot + add intrinsics. 

@ngzhian Didn't think we had consensus here, so I didn't follow up more. 

As there are conflicting opinions, should we poll to see how we should move forward?

Maratyszcza:
@dtig All examples I could find do accumulation on the dot product result, so if we can afford only one instruction, I suggest to leave `i32x4.dot_i16x8_add_s`. It would result in the same code-generation as `i32x4.dot_i16x8_s` + `i32x4.add` on SSE2, but with small WAsm binary size and better performance on recent processors. This is also in-line with other modern instructions sets, which often provide fixed-point dot product instruction only in combination with accumulation (e.g. `DP4A` in NVIDIA PTX or `UDOT`/`SDOT` in ARM64).

ngzhian:
With https://crrev.com/c/2230787 merged we are in sync with toolchain, i32x4.dot_i16x8_s is prototyped in x86, x64, arm, and arm64.

kpu:
@bjacob I want to see 8-bit * 8-bit accumulating into 32-bit as well.  What about sign though since Intel bizarrely does unsigned 8 * signed 8 whereas others do unsigned * unsigned or signed * signed?

bjacob:
@kpu, great questions, sorry I had created confusion earlier by discussing this topic on the present issue which is about something different. I have filed Issue #328 now, let's take discussion here.

tlively:
@Maratyszcza, we discussed this in the sync meeting this morning and the conclusion we came to was that we would like to figure out how important these instructions are to the codec use case before including them in the MVP proposal. Would you be able to do a codec-oriented benchmark of these instructions?

Maratyszcza:
@tlively I have a patch to evaluate the performance effect of these instructions on quantized neural network inference in google/XNNPACK#1007 (currently fails in runtime due to opcode mismatch between LLVM and V8). If you want evaluation specifically on codecs, I'm not the right PoC for that. I remember @dtig showed data from a port of WebP codec in one of the CG meetings. I don't know who did the port, but they would be the right PoC about this instruction too.

tlively:
Thanks @Maratyszcza, I'll try to follow up with the WebP folks to see if they would be interested in prototyping this.

Maratyszcza:
I evaluated the performance effect of these instructions by modifying WebAssembly SIMD microkernels in [XNNPACK library](https://github.com/google/XNNPACK) of neural network operators to use `i32x4.dot_i16x8_s` instruction (`i32x4.dot_i16x8_add_s` is not implemented in LLVM) in matrix-matrix multiplication (GEMM and Indirect GEMM) microkernels. The experiment demonstrated healthy end-to-end speedup on MobileNet family of models:

System | Model | Latency with WAsm SIMD | Latency with WAsm SIMD + `i32x4.dot_i16x8_add_s` | Speedup |
-- | -- | -- | -- | --
ThinkStation P520 (x86-64) | MobileNet v1 | 72 ms | 41 ms | 76% |
ThinkStation P520 (x86-64) | MobileNet v2 | 46 ms | 29 ms | 59% |
Google Pixel 2 (ARM64) | MobileNet v1 | 247 ms | 214 ms | 15% |
Google Pixel 2 (ARM64) | MobileNet v2 | 160 ms | 143 ms | 12% |
Google Pixel 3a (ARM64) | MobileNet v1 | 225 ms | 199 ms | 13% |
Google Pixel 3a (ARM64) | MobileNet v2 | 147 ms | 134 ms | 10% |

The code modifications can be seen in google/XNNPACK#1007

tlively:
We decided at the last SIMD sync to merge in just the non-adding version of this instruction for now. @Maratyszcza could you prepare a version of this PR with just the `i32x4.dot_i16x8_s` instruction that we can merge in?

Maratyszcza:
@tlively Done. Probably we should assign a different opcode as well.

lars-t-hansen:
> Probably we should assign a different opcode as well.

As a general remark, maybe we can do another grand renumbering around the time when we know we're about finished instead of changing opcodes now, seeing as how many of the instructions we're standardizing are implemented in a handful of places that all must be updated (tools, engines).

proposals/simd/ImplementationStatus.md
```diff
@@ -149,6 +149,8 @@
 | `i32x4.min_u`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.max_s`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.max_u`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
+| `i32x4.dot_i16x8_s`        |                           |                    |                    |                    |                    |
+| `i32x4.dot_i16x8_add_s`    |                           |                    |                    |                    |                    |
```
tlively, line 153:
```suggestion
```

ngzhian:
FYI V8 is using 0xfd 0xba currently, so consistent with this PR right now. LGTM

lars-t-hansen:
> FYI V8 is using 0xfd 0xba currently, so consistent with this PR right now.

SpiderMonkey ditto.

bytecodealliance/wasm-tools looks out of sync (0xfd 0xb4).  @alexcrichton: https://github.com/bytecodealliance/wasm-tools/blob/main/crates/wast/src/ast/expr.rs#L943

ngzhian:
@tlively any other concerns with merging this? This LGTM. Maybe the title of PR needs to change, but when merging we can fix it to reflect that only the dot instruction is added.

Maratyszcza:
@ngzhian @tlively Updated the title

tlively (approved):
LGTM!

ngzhian:
There is no pseducode for this op, but I think the text description is straightforward enough. Merging.

julian-seward1:
@Maratyszcza are there any `.wast` testcases available for `i32x4.dot_i16x8_s`?  I looked around the repo but didn't find any -- I may have looked in the wrong place though.

Maratyszcza:
Not that I'm aware of. @ngzhian and @tlively probably know better.

ngzhian:
Not yet, this is a pretty new instruction, so it's not implemented in the interpreter yet. Ofc, contributions welcome, let me know if you're interested (either to contribute implementation or tests, or both).

PR #132, opened 2019-11-04
[test] Integrate latest v128 constant tests

Honry:
New tests for 64x2 constant and more literal
 testing per https://github.com/WAVM/WAVM/issues/195


test/core/simd/simd_const.wast
```diff
@@ -116,21 +216,271 @@
...
+  "unexpected token"
+)
+
+(assert_malformed
+  (module quote "(func (v128.const i8x16) drop)")
+  "unexpected token"
+)
+(assert_malformed
+  (module quote "(func (v128.const i8x16 0x 0x 0x 0x 0x 0x 0x 0x 0x 0x 0x 0x 0x 0x 0x 0x) drop)")
+  "unknown operator"
```
tlively: Why is this message "unknown operator" instead of something more specifically about malformed constants? Changing it would be out of scope here, obviously, but I am curious.

Honry: Wasm[ interpreter](https://github.com/WebAssembly/spec/tree/master/interpreter) verifies all error messages but WAVM doesn't. 

In current interpreter, "unknown operator" contains malformed constants. That's why it's `unknown operator` here. But there should be a specific message for malformed constants, it's a new feature request for interpreter. :)

PR #135, opened 2019-11-04
[test] Integrate tests for 64x2 arithmetic ops from WAVM

Honry:
This PR also centralizes the processing of float-point ops in one file named simd_float_op.py,
see WAVM PR: https://github.com/WAVM/WAVM/pull/209


test/core/simd/meta/simd_i64x2_arith.py
```diff
@@ -0,0 +1,171 @@
...
+        return ['0x01', '-0x01', '-0x8000000000000000', '-0x7fffffffffffffff',
+                '0x7fffffffffffffff', '0x8000000000000000', '0xffffffffffffffff']
+
+    @property
+    def i64x2_i8x16_test_data(self):
+        return {
+            'i64x2.add': [
+                [['0x7fffffffffffffff', ['0', '0', '0', '0', '0', '0', '0', '0x80'] * 4], '-1',
+                 ['i64x2', 'i8x16', 'i64x2']],
+                [['1', '255'], '0', ['i64x2', 'i8x16', 'i64x2']]
```
tlively: Is there documentation anywhere on how these structures are interpreted? As it is they seem very magical.

Honry: It's interpreted in https://github.com/WebAssembly/simd/blob/master/test/core/simd/meta/simd.py#L26.

The first element `['1', '255']` presents the arguments of the assertion, the second element `'0'` presents the test expected result. the last element `['i64x2', 'i8x16', 'i64x2']` presents v128 interpretations of previous elements with one-one correspondence. If the value of each lane are the same, the script will splat the argument to a v128 with corresponding interpretations. 
i.e. this code will be interpreted as `[(v128.const i64x2 1,1), (v128 i8x16 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255 255), (v128.const i64x2 0 0)]`

This looks magical, but simplified the script a lot.

tlively: I see, thanks. It would be good to add a comment in the code pointing to where they are interpreted to make it easier for readers to find out where they need to look.

Honry: Agree! Done.

PR #138, opened 2019-11-12
[Interpreter] infrastructure

penzn:
Provide a buildable skeleton for the interpereter, with end-to end
types, but with lots of gaps in functionality. Unimplemented functions
throw "TODO v128" error; can parse `v128` local type and nothing more.

### DIFF
--- a/interpreter/exec/v128.ml
+++ b/interpreter/exec/v128.ml
@@ -1,0 +1,5 @@
+include Vector.Make
+  (struct
+    include Bytes
+    let bytewidth = 16
+  end)
--- a/interpreter/runtime/memory.ml
+++ b/interpreter/runtime/memory.ml
@@ -111,6 +111,7 @@ let load_value mem a o t =
   | I64Type -> I64 n
   | F32Type -> F32 (F32.of_bits (Int64.to_int32 n))
   | F64Type -> F64 (F64.of_bits n)
+  | V128Type -> failwith "TODO v128"
 
 let store_value mem a o v =
   let x =
@@ -119,6 +120,7 @@ let store_value mem a o v =
     | I64 x -> x
     | F32 x -> Int64.of_int32 (F32.to_bits x)
     | F64 x -> F64.to_bits x
+    | V128 x -> failwith "TODO v128" (* FIXME V128.to_bits x requires store to accept something other than int64 *)
   in storen mem a o (Types.size (Values.type_of v)) x
 
 let extend x n = function
--- a/interpreter/text/lexer.mll
+++ b/interpreter/text/lexer.mll
@@ -50,6 +50,7 @@ let value_type = function
   | "i64" -> Types.I64Type
   | "f32" -> Types.F32Type
   | "f64" -> Types.F64Type
+  | "v128" -> Types.V128Type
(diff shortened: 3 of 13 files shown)


penzn:
@ngzhian let's join forces if you are still interested in the interpreter.

ngzhian:
Cool! I have been hacking a bit at the interpreter too, but haven't cleaned it up enough to put up a PR. We went for similar implementation to use Bytes. I've implemented a bit of the parsing for v128.const so it at least chugs through simd_const.wast test case.

https://github.com/ngzhian/simd/commit/d86bf001cf47a5881269b2aaebc0a17083126b96

interpreter/exec/v128.ml
```diff
@@ -0,0 +1,5 @@
+include Vector.Make
```
ngzhian: Maybe we can have v128 without using Vector? Having a Vector doesn't seem necessary at this point since we only have v128, unlike floats or ints.

penzn: Yep, maybe. I thought about this too, but decided to still do it because this structure is assumed in some other places and operations on vectors probably would be relatively universal (as they would probably work on an array of bytes). Also, this might help with vectors 2.0 (whichever way they would end up).

arunetm: from a design standpoint, we should consider easy extension to >128bit types as well.

penzn:
@ngzhian, I just noticed the link to your change, the frontend part looks good, may be we can merge it together with this one somehow.

CC/FYI @rossberg, as we are hacking at the interpreter.

ngzhian:
I think after this is merged I can try to rebase the frontend changes from my patch. That way there is less code to review in each change.

interpreter/runtime/memory.ml
```diff
@@ -119,6 +120,7 @@ let store_value mem a o v =
     | I64 x -> x
     | F32 x -> Int64.of_int32 (F32.to_bits x)
     | F64 x -> F64.to_bits x
+    | V128 x -> failwith "TODO v128" (* FIXME V128.to_bits x requires store to accept something other than int64 *)
```
rossberg: Not sure I understand this comment.

interpreter/text/lexer.mll
```diff
@@ -139,7 +140,8 @@ let reserved = ([^'\"''('')'';'] # space)+  (* hack for table size *)
 
 let ixx = "i" ("32" | "64")
 let fxx = "f" ("32" | "64")
-let nxx = ixx | fxx
+let v = "v128"
```
rossberg: `vxxx` seems more adequate.

interpreter/runtime/memory.ml
penzn: I concluded that the section (inside of `match`) operates on `I64`, though I might have been wrong.

penzn:
> I think after this is merged I can try to rebase the frontend changes from my patch. That way there is less code to review in each change.

`const` is needed in order to verify other functionality. Let me know if you need help with the change.

I opened #141 to track progress of adding the interpreter.

@arunetm @dtig should we merge this?

PR #140, opened 2019-11-15
Update some of the test assertions in `simd_lane.wast`

alexcrichton:
This updates a few of the assertions in `simd_lane.wast` found after writing a parser which consumes `*.wast` files and choking on a few of the test cases here. I'm not 100% sure if all of these changes are correct, so feedback is appreciated!

Additionally, are these tests currently run through a reference implementation somewhere? I'm not sure if the spec interpreter in this repo has been updated for the tests yet.

### DIFF
--- a/test/core/simd/simd_lane.wast
+++ b/test/core/simd/simd_lane.wast
@@ -1,7 +1,7 @@
 ;; Tests for the extract_lane, replace_lane, swizzle and shuffle group instructions
 
 
-(module 
+(module
   (func (export "i8x16_extract_lane_s-first") (param v128) (result i32)
     (i8x16.extract_lane_s 0 (local.get 0)))
   (func (export "i8x16_extract_lane_s-last") (param v128) (result i32)
@@ -738,9 +738,9 @@
 (assert_return (invoke "as-i64x2_splat-operand" (v128.const i64x2 -1 0)) (v128.const i64x2 -1 -1))
 (assert_return (invoke "as-f64x2_splat-operand" (v128.const f64x2 inf nan)) (v128.const f64x2 inf inf))
 (assert_return (invoke "as-i8x16_add-operands"
-  (v128.const i8x16 0xff 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16) (i32.const 1) 
+  (v128.const i8x16 0xff 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16) (i32.const 1)
   (v128.const i8x16 16 15 14 13 12 11 10 9 8 7 6 5 4 3 2 0xff) (i32.const 1))
-  (v128.const i8x16 17 17 17 17 17 17 17 17 17 17 17 17 17 17 17 17)) 
+  (v128.const i8x16 17 17 17 17 17 17 17 17 17 17 17 17 17 17 17 17))
 (assert_return (invoke "as-i16x8_add-operands"
   (v128.const i16x8 -1 4 9 16 25 36 49 64) (i32.const 1)
   (v128.const i16x8 64 49 36 25 16 9 4 -1) (i32.const 1))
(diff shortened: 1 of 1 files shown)


test/core/simd/simd_lane.wast
```diff
@@ -489,14 +489,8 @@
...
-(assert_invalid (module (func (result v128)
-  (v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14
-  (v128.const i8x16 15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0)
-  (v128.const i8x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15)))) "invalid lane length")
-(assert_invalid (module (func (result v128)
-  (v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16
-  (v128.const i8x16 15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0)
-  (v128.const i8x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15)))) "invalid lane length")
+(assert_malformed (module quote "(func v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14)") "invalid lane length")
+(assert_malformed (module quote "(func v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16)") "invalid lane length")
```
penzn: Would this work like this without arguments?

penzn:
We are working on interpreter, skeleton code is in #138, @ngzhian has some code he might be able to add soon after. I can try to prioritize support for shuffles, but there might be other things on the way to it.

Also adding @Honry

test/core/simd/simd_lane.wast
alexcrichton: Hm sorry I'm not sure I quite understand, could you clarify what you're thinking should work without arguments?

penzn: Does the shuffle instruction need all arguments for this test to work correctly? You are not providing the two v128 arguments to it.

alexcrichton: Ah ok, I see what you mean, and sure! I've updated the test to typecheck but still have this expected assertion failure.

Sorry I also don't know how to have multiline strings in the wat format so it's a bit messy...

test/core/simd/simd_lane.wast
```diff
@@ -489,14 +489,8 @@
...
-(assert_invalid (module (func (result v128)
-  (v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14
-  (v128.const i8x16 15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0)
-  (v128.const i8x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15)))) "invalid lane length")
-(assert_invalid (module (func (result v128)
-  (v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16
-  (v128.const i8x16 15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0)
-  (v128.const i8x16 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15)))) "invalid lane length")
+(assert_malformed (module quote "(func (param v128) (result v128) local.get 0 local.get 0 v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14)") "invalid lane length")
+(assert_malformed (module quote "(func (param v128) (result v128) local.get 0 local.get 0 v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16)") "invalid lane length")
```
Honry:
```suggestion
(assert_malformed (module quote "(func (param v128) (result v128) "
  "(local.get 0) (local.get 0)"
  "v8x16.shuffle 0 1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16)")
  "invalid lane length")
```

Honry: It's more readable with above suggestion. :)

alexcrichton: Oh right!

PR #144, opened 2019-11-19
Add support for v128.const

ngzhian:
This adds partial support for v128.const, I test it using the simd_const.wast file, and manages to get most of it working. I'm still working out some issues (parse error messages), but wanted to put this up first to get comments on how this looks. In particular, I'm not confident that the logic in lexer/parser is in the right place.

E.g. we end up hitting the same `CONST` case in the lexer, but we don't really do anything there in the `v128.const` case, and instead rely on the parser to match on `CONST SIMD_SHAPE literal_list`, because the signature of `numop` won't match in the case of `v128.const`, which takes a list of strings, rather than a single string.

Note that this requires OCaml 4.08, and might require you installing it yourself (distributions might not have this version yet).

Also I commented out 2 tests in simd_const.wast because they rely on implementing i32x4 operations, which will be done in the future. to_string needs to be fixed as well.

Feedback very welcome, ptal.


ngzhian:
@penzn Have to + you into this thread this way, not sure why I cannot added you via "Reviewers" in the sidebar. Ptal at this, thank you!

interpreter/text/lexer.mll
```diff
@@ -65,12 +65,15 @@ let floatop t f32 f64 =
   | "f64" -> f64
   | _ -> assert false
 
-let numop t i32 i64 f32 f64 =
+let fail_v128 = (fun s -> assert false)
```
penzn: Does it show function name in the error message? If not you can try something like `failwith "<some message>"`, would be easier to see if you are hitting one of those cases or something else.

interpreter/text/parser.mly
```diff
@@ -39,6 +39,45 @@ let ati i =
...
+let f32_of_string s = F32.to_bits (literal (fun s -> F32.of_string s.it) s)
+let f64_of_string s = F64.to_bits (literal (fun s -> F64.of_string s.it) s)
+
+let simd_literal f shape ss =
+    let open Bytes in
+    let b = create 16 in
+    (match shape with
+    | "i8x16" ->
+            List.iteri (fun i s -> set_uint8 b i (int8_of_string s)) ss;
+            if List.length ss != 16 then parse_error "unexpected token";
```
rossberg: Instead all the separate conditionals for length tests, you can put guards like "when List.length ss = 2" on the patterns. Then you can unify and move the parse error to the default case. Or more visually, you could even `match shape, ss with` and then have patterns like `| "i64x2", [_; _] ->

interpreter/text/parser.mly
```diff
@@ -39,6 +39,45 @@ let ati i =
...
+    if i > max_32 || i < min_32 then error at "constant out of range" else i_32
+
+let int8_of_string s = range_check (I32.of_string s.it) (-128) 255 s.at
+let int16_of_string s = range_check (I32.of_string s.it) (-32768) 65535 s.at
+let int32_of_string s = I32.to_bits (literal (fun s -> I32.of_string s.it) s)
+let int64_of_string s = I64.to_bits (literal (fun s -> I64.of_string s.it) s)
+let f32_of_string s = F32.to_bits (literal (fun s -> F32.of_string s.it) s)
+let f64_of_string s = F64.to_bits (literal (fun s -> F64.of_string s.it) s)
+
+let simd_literal f shape ss =
```
rossberg: The `f` parameter seems to be unused? How do you detect, e.g., `(i32.const f64x2 0 0)`?

To avoid that, it may be easiest to turn `v128.const` into a separate keyword.

interpreter/text/parser.mly
```diff
@@ -39,6 +39,45 @@ let ati i =
...
+            if List.length ss != 2 then parse_error "unexpected token";
+    | "f32x4" ->
+            List.iteri (fun i s -> set_int32_le b i (f32_of_string s)) ss;
+            if List.length ss != 4 then parse_error "unexpected token";
+    | "f64x2" ->
+            List.iteri (fun i s -> set_int64_le b i (f64_of_string s)) ss;
+            if List.length ss != 2 then parse_error "unexpected token";
+    | _ -> assert false);
+    let v = V128.of_bits b in
+    (v128_const (v @@ (ati 1)), Values.V128 v)
```
rossberg:
```suggestion
    (v128_const (v @@ ati 1), Values.V128 v)
```

interpreter/text/parser.mly
```diff
@@ -39,6 +39,45 @@ let ati i =
...
+let int64_of_string s = I64.to_bits (literal (fun s -> I64.of_string s.it) s)
+let f32_of_string s = F32.to_bits (literal (fun s -> F32.of_string s.it) s)
+let f64_of_string s = F64.to_bits (literal (fun s -> F64.of_string s.it) s)
+
+let simd_literal f shape ss =
+    let open Bytes in
+    let b = create 16 in
+    (match shape with
+    | "i8x16" ->
+            List.iteri (fun i s -> set_uint8 b i (int8_of_string s)) ss;
```
rossberg: I can't figure out where these `set_` functions are coming from. But it would be nicer to abstract away the construction of a SIMD value from its lanes into some suitable constructor functions in the Values module.

interpreter/text/parser.mly
```diff
@@ -319,6 +363,7 @@ plain_instr :
   | STORE offset_opt align_opt { fun c -> $1 $3 $2 }
   | MEMORY_SIZE { fun c -> memory_size }
   | MEMORY_GROW { fun c -> memory_grow }
+  | CONST SIMD_SHAPE literal_list { fun c -> fst (simd_literal $1 $2 $3) }
```
rossberg: Nit: move this after the regular const case.

interpreter/text/parser.mly
ngzhian: These come from Bytes (see `let open Bytes` right above).
"abstract away the construction of a SIMD value from its lanes"
I don't understand this, can you illustrate with an example? Since the "lanes" of a SIMD value (v128 value in the interpreter) are all the same (Bytes type), I'm not sure how having different constructors will help.
Do you mean having constructors like:
build_i8x16 : string -> V128.t
build_i16x8 : string -> V128.t
etc?

interpreter/text/parser.mly
ngzhian: Separate v128.const into a separate keyword, thanks for the suggestion!

interpreter/text/parser.mly
ngzhian: Good tip, made that change to use guards.

interpreter/text/lexer.mll
ngzhian: Removed this in the latest commit after splitting v128.const into its own keyword, thanks!

ngzhian:
I still have to work through the test failure messages, I think some of them might need to change, e.g. (v128.const i64x2 0) is missing a constant, what should the error message say?

interpreter/text/parser.mly
penzn: > These come from Bytes (see `let open Bytes` right above).

This requires OCaml 4.08 (Ubuntu has 4.05 currently).

From source point of view lanes are not always bytes - `i16x8` has two byte lanes, `i32x4` -  four-bye and so on. I think passing array or list of strings, containing source text for corresponding lanes, would work for constructors.

interpreter/text/parser.mly
```diff
@@ -161,6 +192,7 @@ let inline_type_explicit (c : context) x ft at =
 %token INPUT OUTPUT
 %token EOF
 
+%token<string> SIMD_SHAPE
```
rossberg: Can you introduce a proper Vector.shape enum to represent this?

interpreter/text/parser.mly
```diff
@@ -39,6 +39,37 @@ let ati i =
 let literal f s =
   try f s with Failure _ -> error s.at "constant out of range"
 
+let range_check i32 min max at =
+  let i = Int32.to_int i32 in
+  if i > max || i < min then error at "constant out of range" else i
+
+let int8_of_string s = range_check (I32.of_string s.it) (-128) 255 s.at
```
rossberg: Nit: i8 (same below)
Down the line you'll probably need to introduce proper I8 and I16 modules to implement various lane-wise ops, which would eliminate the range_check hack.

interpreter/text/parser.mly
```diff
@@ -39,6 +39,37 @@ let ati i =
...
+let int16_of_string s = range_check (I32.of_string s.it) (-32768) 65535 s.at
+let int32_of_string s = I32.to_bits (literal (fun s -> I32.of_string s.it) s)
+let int64_of_string s = I64.to_bits (literal (fun s -> I64.of_string s.it) s)
+let f32_of_string s = F32.to_bits (literal (fun s -> F32.of_string s.it) s)
+let f64_of_string s = F64.to_bits (literal (fun s -> F64.of_string s.it) s)
+
+let simd_literal shape ss =
+  let open Bytes in
+  let b = create 16 in
+  (match shape with
```
rossberg: Likewise, the logic for constructing simd values from lanes will probably have to move to vector.ml, since you'll need it elsewhere.

interpreter/text/parser.mly
ngzhian: Yes you're right, I will add a note on OCaml 4.08 to the PR.

interpreter/text/parser.mly
ngzhian: Done in newest commits.

interpreter/text/parser.mly
ngzhian: I moved the construction of Simd128 values into s128, the parser deals with validation, then calls the of_strings constructor to build the value.

interpreter/exec/v128.ml
```diff
@@ -2,4 +2,24 @@ include Simd.Make
   (struct
     include Bytes
     let bytewidth = 16
+
+    (* Constructs v128 of a certain shape from a list of strings. Does not do any validation,
```
rossberg: The other of_string functions do validation, so it would be more consistent to do that here as well (other callers might screw it up as well). The parser can simply catch the exception and generate an error.

ngzhian: Makes sense, thanks for pointing this out. Moved the validation into v128.to_strings.
The error reporting for this is bad, defaults to the whole line, where it should pinpoint the constant that is out of bounds, and I can fix it in a subsequent patch.

interpreter/text/parser.mly
ngzhian: Added a TODO for this, can add a I8 and I16 module in a future patch.

interpreter/exec/simd.ml
```diff
@@ -16,7 +28,9 @@ sig
   type bits
   val default : t (* FIXME good name for default value? *)
   val to_string : t -> string
+  val of_bits : bits -> t
   val to_bits : t -> bits
+  val of_strings : shape -> string list -> t
 end
 
 module Make (Rep : RepType) : S with type bits = Rep.t =
```
rossberg: Aside: Do we actually need this functor? It doesn't seem to do anything interesting.

interpreter/exec/v128.ml
```diff
@@ -1,5 +1,30 @@
...
+        List.iteri (fun i s -> set_uint16_le b i (i16_of_string s)) ss
+      | Simd.I32x4 ->
+        List.iteri (fun i s -> set_int32_le b i (I32.of_string s)) ss
+      | Simd.I64x2 ->
+        List.iteri (fun i s -> set_int64_le b i (I64.of_string s)) ss
+      | Simd.F32x4 ->
+        List.iteri (fun i s -> set_int32_le b i (F32.to_bits (F32.of_string s))) ss
+      | Simd.F64x2 ->
+        List.iteri (fun i s -> set_int64_le b i (F64.to_bits (F64.of_string s))) ss);
+      if List.length ss != (Simd.lanes shape) then raise (Invalid_argument "wrong length")
```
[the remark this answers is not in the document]
ngzhian: Good point, I put the length check after the constants parsing because the error messages in the test indicates that the ordering should be like so.
I.e. (v128.const f64x2 nan:0x0) will throw a constant out of range, rather than a unexpected token.

I wonder if we want to change the test assertions, or I can create a bytes of the right size, then truncate it at the end.

interpreter/exec/simd.ml
ngzhian: From https://github.com/WebAssembly/simd/pull/138#discussion_r345508572 it seemed like we wanted this for extensibility.

interpreter/exec/v128.ml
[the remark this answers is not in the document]
rossberg: The existing test suite tries to avoid prescribing any specific choice of diagnostic in the case where a program has multiple errors, simply by making sure that every negative test contains only one error. So I would propose changing the tests that way, such that the order of the checks does no longer matter.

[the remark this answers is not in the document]
rossberg: And you probably want to add a test that has too many literals.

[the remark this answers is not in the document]
ngzhian: In the latest commit I removed these 3 of these checks, they looked redundant to the existing i32x4 checks. I moved 1 down and checked it to assert "unexpected token". Also added a test with too many literals.

interpreter/text/parser.mly
```diff
@@ -39,48 +39,15 @@ let ati i =
...
-  | Simd.F64x2 -> List.iter check_simd_f64 ss
-
 let simd_literal shape ss =
-  validate_simd_literal shape ss;
-  validate_simd_literal_length shape ss;
-  let v = V128.of_strings shape (List.map (fun s -> s.it) ss) in
-  (* This List.hd call is okay since we validated the length. *)
-  (v128_const (v @@ (List.hd ss).at), Values.V128 v)
+  try
+      let v = V128.of_strings shape (List.map (fun s -> s.it) ss) in
```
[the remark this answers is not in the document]
ngzhian: Done.

interpreter/text/parser.mly
```diff
@@ -39,48 +39,15 @@ let ati i =
...
+      (* This List.hd call is okay an exception would have been raise if length is 0. *)
+      (v128_const (v @@ (List.hd ss).at), Values.V128 v)
```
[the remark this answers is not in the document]
ngzhian: Done in latest commit. The script part is a bit weird, i pass `at ()` to simd_literal, and pass `ati 3` again. (As I write this I realize passing ati 3 is wrong now, will fix it in another commit).

interpreter/text/parser.mly
```diff
@@ -335,7 +334,7 @@ plain_instr :
   | MEMORY_SIZE { fun c -> memory_size }
   | MEMORY_GROW { fun c -> memory_grow }
   | CONST literal { fun c -> fst (literal $1 $2) }
-  | V128_CONST SIMD_SHAPE literal_list { fun c -> fst (simd_literal $2 $3) }
+  | V128_CONST SIMD_SHAPE literal_list { fun c -> fst (simd_literal $2 $3 (at ())) }
```
rossberg: Since this is deferred by a fun, you need to let-bind the at value outside the fun, otherwise the parser state will have change already when the fun gets called.

interpreter/text/parser.mly
```diff
@@ -39,15 +39,14 @@ let ati i =
...
-      (* This List.hd call is okay an exception would have been raise if length is 0. *)
-      (v128_const (v @@ (List.hd ss).at), Values.V128 v)
+    let v = V128.of_strings shape (List.map (fun s -> s.it) ss) in
+    (v128_const (v @@ at), Values.V128 v)
   with
     (* TODO better location for error messages. *)
-    | Failure _ -> error (ati 1) "constant out of range"
-    | Invalid_argument _ -> error (ati 1) "unexpected token"
+    | Failure _ -> error at "constant out of range"
+    | Invalid_argument _ -> error at "unexpected token"
```
rossberg: Nit: "wrong number of lane literals"

penzn:
> Sorry for taking this long. I think you are on the right track.

I am sorry @ngzhian the comment originally said "not on the right track", I just noticed. That was _not_ intentional, my apologies. I think I just typed something else and fixed the comment before clicking send without really re-reading it.

interpreter/exec/simd.ml
penzn: Compatibility with some other code as well, there is some code that expects `RetType` and `S` distinction.

penzn:
ping @arunetm 😄

test/core/simd/simd_const.wast
```diff
@@ -394,7 +394,7 @@
 )
 (assert_malformed
   (module quote "(func (v128.const f32x4 nan:0x0 nan:0x0 nan:0x0 nan:0x0) drop)")
-  "unknown operator"
+  "constant out of range"
```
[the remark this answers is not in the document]
Honry: `nan:0x0` is not a valid NaN literal, so looks like `constant out of range` does not make sense.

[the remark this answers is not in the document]
ngzhian: Right now `(module (func (f32.const nan:0x0)))` gives the same `constant out of range` error message. Maybe we can have a new error message for `invalid NaN literal`?

interpreter/text/parser.mly
ngzhian: Fixed, and changed test assertion messages.

interpreter/text/parser.mly
ngzhian: Good catch, thanks!

test/core/simd/simd_const.wast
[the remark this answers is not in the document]
Honry: Sounds good to me.

test/core/simd/simd_const.wast
```diff
@@ -1076,7 +1073,7 @@
   (func (export "i32x4.test") (result v128) (return (v128.const i32x4 0x0bAdD00D 0x0bAdD00D 0x0bAdD00D 0x0bAdD00D)))
   (func (export "i32x4.smax") (result v128) (return (v128.const i32x4 0x7fffffff 0x7fffffff 0x7fffffff 0x7fffffff)))
   (func (export "i32x4.neg_smax") (result v128) (return (v128.const i32x4 -0x7fffffff -0x7fffffff -0x7fffffff -0x7fffffff)))
-  (func (export "i32x4.inc_smin") (result v128) (return (i32x4.add (v128.const i32x4 -0x80000000 -0x80000000 -0x80000000 -0x80000000) (v128.const i32x4 1 1 1 1))))
+  (; (func (export "i32x4.inc_smin") (result v128) (return (i32x4.add (v128.const i32x4 -0x80000000 -0x80000000 -0x80000000 -0x80000000) (v128.const i32x4 1 1 1 1)))) ;)
```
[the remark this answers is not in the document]
ngzhian: I commented this out because it uses i32x4.add and I did not want to implement i32x4.add inside this patch. In future patches when we have i32x4.add implemented, we can uncomment this.
Actually, now that I think of it, it doesn't really need to be commented out, since we are not running these tests by default yet. I will upload a patch to uncomment this.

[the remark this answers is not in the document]
ngzhian: Fixed in #159 159

PR #145, opened 2019-11-20
[test] Sync tests from WAVM#223

Honry:
https://github.com/WAVM/WAVM/pull/223

    [simd] Improve simd test generation tool

    - Use names in the string interpolation so its clear
      what each part is supposed to be.
    - Use unified naming converson for params.

### DIFF
--- a/test/core/simd/meta/simd.py
+++ b/test/core/simd/meta/simd.py
@@ -9,66 +9,66 @@ currently only supports generating v128 const constant data.
 class SIMD(object):
 
     # Constant template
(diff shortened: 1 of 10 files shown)


Honry:

abrown:
Is this only renaming of variables or are there functionality changes? Could you point to the functionality changes, if any? I can't see them 😄.

Honry:
@abrown, no functionality changes here, just try to make the naming convention consistent.

Honry:
Friendly ping. :)

PR #147, opened 2019-12-04
[test] Sync tests from WAVM#220

Honry:
https://github.com/WAVM/WAVM/pull/220

    [simd] Add more literal tests for lane ops

    Remove some redundant tests BTW


Honry:
@arunetm, can you land this?

PR #154, opened 2019-12-11
[test] Add more literal tests for min/max/abs ops

Honry:
This PR covers tests for some float literal introduced in https://github.com/WebAssembly/spec/pull/1069, specific for following ops:

f32x4.abs
f32x4.min
f32x4.max
f64x2.abs
f64x2.min
f64x2.max

These tests are pulled from https://github.com/WAVM/WAVM/pull/234/.


Honry:

test/core/simd/meta/simd_float_op.py
```diff
@@ -160,28 +167,40 @@ def binary_op(self, op: str, p1: str, p2: str) -> str:
         if op == 'min':
             if '-0x0p+0' in [p1, p2] and '0x0p+0' in [p1, p2]:
                 return '-0x0p+0'
-            result = min(f1, f2)
+            if hex_form:
+                result = min(f1, f2)
```
tlively: This logic would be easier to follow if this was simply `return min(f1, f2).hex()`. Same below.

Honry: Indeed, done.

PR #155, opened 2019-12-11
[test] Add more literal tests for rest ops

Honry:
This PR covers tests for some float literal introduced in
https://github.com/WebAssembly/spec/pull/1069, specific for following
ops:

Bit shifts
Bitwise operations
Boolean horizontal reductions
Conversions
Integer comparisons
Store

These tests are pulled from https://github.com/WAVM/WAVM/pull/236/


Honry:

PR #156, opened 2019-12-16
[test] Use unified assert_result* assertion

Honry:
to replace assert_return_arithmetic_nan and assert_return_canonical_nan

Tests are pulled from https://github.com/WAVM/WAVM/pull/240


Honry:
@arunetm, @tlively, @abrown, PTAL, thanks!

test/core/simd/meta/simd_f64x2_arith.py
```diff
@@ -101,110 +101,59 @@ def combine_binary_arith_test_data(self):
...
         }
 
-    def get_normal_case(self):
-        return super().get_normal_case().replace('nan_f32x4', 'nan_f64x2')
-
     def get_invalid_cases(self):
         return super().get_invalid_cases().replace('32', '64')
 
     @property
     def mixed_sqrt_nan_test_data(self):
```
tlively: why does this function have `sqrt` in the name but many different operations in its data?

test/core/simd/simd_f32x4.wast
```diff
@@ -1066,294 +937,246 @@
...
+                                   (v128.const f32x4 nan:arithmetic nan:arithmetic nan:arithmetic nan:arithmetic))
+(assert_return (invoke "f32x4.min" (v128.const f32x4 -nan:0x200000 -nan:0x200000 -nan:0x200000 -nan:0x200000)
+                                   (v128.const f32x4 -nan -nan -nan -nan))
+                                   (v128.const f32x4 nan:arithmetic nan:arithmetic nan:arithmetic nan:arithmetic))
+(assert_return (invoke "f32x4.min" (v128.const f32x4 -nan:0x200000 -nan:0x200000 -nan:0x200000 -nan:0x200000)
+                                   (v128.const f32x4 nan:0x200000 nan:0x200000 nan:0x200000 nan:0x200000))
+                                   (v128.const f32x4 nan:arithmetic nan:arithmetic nan:arithmetic nan:arithmetic))
+(assert_return (invoke "f32x4.min" (v128.const f32x4 -nan:0x200000 -nan:0x200000 -nan:0x200000 -nan:0x200000)
+                                   (v128.const f32x4 -nan:0x200000 -nan:0x200000 -nan:0x200000 -nan:0x200000))
+                                   (v128.const f32x4 nan:arithmetic nan:arithmetic nan:arithmetic nan:arithmetic))
```
tlively: It looks like the number of tests has changed. Is that expected?

test/core/simd/simd_f32x4.wast
```diff
@@ -98,129 +61,37 @@
...
-  (invoke "call_indirect_vv_v_f32x4_extract_lane_1"
-    (v128.const f32x4 nan 0 0 1)
-    (v128.const f32x4 0 -nan 1 0)
-    (i32.const 1)
   )
+  (v128.const f32x4 nan:canonical nan:canonical 0 0)
 )
+;; f32x4.max
 (assert_return
-  (invoke "call_indirect_vv_v_f32x4_extract_lane_2"
```
tlively: Why do we no longer have these extract_lane tests?

test/core/simd/meta/simd_f64x2_arith.py
Honry: Yeah, this is for combination testing, the name really does not make sense, I will update it. :)

test/core/simd/simd_f32x4.wast
Honry: Before we have unified assert_result* assertion, if we want to test `(invoke "f32x4.min" (v128.const f32x4 nan 0 0 1) (v128.const f32x4 0 -nan 1 0))`, we have to check each lane of the test result which contains partial canonical NAN, that’s why we use `indirect_call` + `extract_lane`.

test/core/simd/simd_f32x4.wast
Honry: Yes, it's expected. That's because we simplified those extract_lane tests.

Honry:
> Also, would it make more sense for min and max tests to be in the same files as the other arithmetic tests?

@tlively, the rule of test classification was according to the sections of [SIMD spec](https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md), one more consideration was that these two files are too large at present, if we merge them into one it's not convenience for reading.

test/core/simd/meta/simd_f64x2_arith.py
Honry: Done.

tlively:
Will wait for someone else to take a look as well, then happy to merge.

tlively:
cc @arunetm

arunetm:
LGTM.

ngzhian:
Was "nan:canonical" added to the parser? I don't seem to find the code doing so, and doing a test run on the file with this token gives me a "syntax error: unknown operator"

ngzhian:
Also as a follow up, I can't seem to have "nan:canonical" work well in cases where we have a binary operation involving -NaN.
```
(assert_return
  (invoke "f32x4.min"
    (v128.const f32x4 nan 0 0 1)
    (v128.const f32x4 0 -nan 1 0)
  )
  (v128.const f32x4 nan:canonical nan:canonical 0 0)
)
```

First we need to pick a value that `nan:canonical` parses into, is it positive or negative?
Let's arbitrarily pick positive, then this test case will fail in the second lane.
If we pick negative, then the test case will fail in the first lane.
(Canonicalization does nothing here since both pos_nan and neg_nan are already canonical).

It seems like we need to special case assertions for canonicals to ignore signs?
Or maybe we can always mask the sign since it is non-deterministic
"When the result of a floating-point operator other than fneg, fabs, or fcopysign is a NaN, then its sign is non-deterministic "?

AndrewScheidecker:
@ngzhian It was added to the base spec repo in https://github.com/WebAssembly/spec/pull/1104.

The idea is that `(f32.const nan:canonical)`, when used as the expected return value in `assert_return`, denotes a set of possible `f32` values that will be accepted. `nan:canonical` denotes the set containing only `+nan` and `-nan`, but `nan:arithmetic` denotes all the possible NaNs with the significand MSB set.

ngzhian:
Ah! Thanks, looks like this repo needs a sync then!

PR #158, opened 2019-12-17
Update README to specify requiring OCaml 4.08

ngzhian:
We now require OCaml 4.08, due to using set_* functions in Bytes module (https://caml.inria.fr/pub/docs/manual-ocaml/libref/Bytes.html) since https://github.com/WebAssembly/simd/pull/144

### DIFF
--- a/interpreter/README.md
+++ b/interpreter/README.md
@@ -15,7 +15,7 @@ The text format defines modules in S-expression syntax. Moreover, it is generali
 
 ## Building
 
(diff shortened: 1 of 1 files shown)


ngzhian:
Our mirror https://wasm.storage.googleapis.com/ doesn't have 4.08, should we point to OCaml's homepage?

penzn:
Looks like half of package managers which OCaml lists on its [install page](https://ocaml.org/docs/install.html) are behind 4.08. From experience it is a good idea to stick with provided packages, but I am not sure how bad of an issue that is for OCaml. @rossberg what do you think?

I took a quick look into what it would take to make interpreter compatible with OCaml 4.05, and the best I can come up with is using `Bytes.init` with a function that would extract correct byte from the value. If we decide to go the 4.05 route I think I can make it happen, otherwise readme note sounds good.

ngzhian:
Replaced by #378

PR #163, opened 2019-12-19
[test] Add spec tests for integer min/max ops

Honry:
Tests are pulled from https://github.com/WAVM/WAVM/pull/242


Honry:

arunetm:
Let's wait to merge until we get a second set of eyes on the python changes.

test/core/simd/meta/simd_integer_op.py
```diff
@@ -0,0 +1,65 @@
...
+        :param op: min_s, min_u, max_s, max_u
+        :param p1: integer value in hex
+        :param p2: integer value in hex
+        :lane_width: bit number of each lane in SIMD v128
+        :return:
+        """
+        if '0x' in p1:
+            base1 = 16
+        else:
+            base1 = 10
```
tlively: It looks like the parameters are allowed to be base-10, but the doc comment specifies that they must be hex. One of these should be updated.

test/core/simd/meta/simd_lane_value.py
```diff
@@ -0,0 +1,32 @@
...
+    def mask(self):
+        return pow(2, self.lane_width) - 1
+
+    @property
+    def mod(self):
+        return pow(2, self.lane_width)
+
+    @property
+    def quarter(self):
+        return pow(2, self.lane_width - 2)
```
tlively: What is `quarter` used for?

test/core/simd/simd_i16x8.wast
```diff
@@ -0,0 +1,387 @@
...
+
+;; Test operation with empty argument
+(assert_invalid
+  (module
+    (func $i16x8.min_s-1st-arg-empty (result v128)
+      (i16x8.min_s (v128.const i16x8 0 0 0 0 0 0 0 0))
+    )
+  )
+  "type mismatch"
+)(assert_invalid
```
tlively: This is a little funky without an extra newline in there.

test/core/simd/simd_i16x8.wast
```diff
@@ -0,0 +1,387 @@
+;; Tests for i16x8 [min_s, min_u, max_s, max_u] operations.
```
tlively: Should the file name include "min" and "max" or is it going to eventually include many more operations?

test/core/simd/simd_i8x16.wast
```diff
@@ -0,0 +1,387 @@
...
+)(assert_invalid
+  (module
+    (func $i8x16.max_u-all-args-empty (result v128)
+      (i8x16.max_u)
+    )
+  )
+  "type mismatch"
+)
+
+;; Combination
```
tlively: How are we deciding what combinations of instructions are useful to test?

Honry: Good question, actually I don't have a specific definition, I just choose some combinations randomly.

And we don't want to leverage full set of pairwise comparison, since it will cause exponential test cases with low efficiency to capture bugs.

test/core/simd/simd_i16x8.wast
tlively: Not any good ideas. The best I can think of is `simd_i16x8_val_arith.wast`. Adding "val" doesn't really convey much more information, but it does suggest that the contents are more specific than the name `simd_i16x8.wast` suggests. My other idea is `simd_i16x8_arith2.wast`, which has the same benefits and drawbacks.

test/core/simd/simd_i8x16.wast
tlively: Sounds good. I would err on the side of testing fewer combinations because I agree that we want to avoid having more tests than necessary and I don't think the combinations will find as many bugs.

tlively:
Thanks, @Honry! I'll merge this now, but feel free to change the filename in a followup PR if you want.

test/core/simd/simd_i16x8.wast
Honry: I'd prefer `simd_i16x8_arith2.wast`. :)

PR #165, opened 2019-12-20
Add some f32x4 instructions

ngzhian:
Add f32x4 min, max, abs extract lane, and i32x4 extract lane as examples
of how SIMD operations can be implemented.

extract_lane operations are missing validation, which can be added in
future patches.

With this patch, we don't fully support running the test `simd_f32x4.wast` yet, due to missing parsing of the `result` token in scripts. In particular, right now we only support const literal or a single canonical/arithmetic NaN. We will need to add support for parsing v128.const containing literals and/or canonical/arithmetic NaN, this can be done in a future patch.

### DIFF
--- a/interpreter/exec/v128.ml
+++ b/interpreter/exec/v128.ml
@@ -4,6 +4,17 @@ include Simd.Make
     let bytewidth = 16
     let to_string s = s
 
+    let to_i32x4 s =
+      List.map (fun i -> I32.of_bits @@ get_int32_le s i) Simd.f32x4_indices
+
+    let to_f32x4 s =
+      List.map (fun i -> F32.of_bits @@ get_int32_le s i) Simd.f32x4_indices
+
+    let of_f32x4 fs =
+      let b = create bytewidth in
+      List.iter2 (fun i f -> set_int32_le b i (F32.to_bits f)) Simd.f32x4_indices fs;
+      b
+
     let of_strings shape ss =
       if List.length ss <> Simd.lanes shape then raise (Invalid_argument "wrong length");
       let range_check i32 min max at =
--- a/interpreter/exec/v128.ml
+++ b/interpreter/exec/v128.ml
@@ -5,15 +5,15 @@ include Simd.Make
     let to_string s = s
 
     let to_i32x4 s =
-      List.map (fun i -> I32.of_bits @@ get_int32_le s i) Simd.f32x4_indices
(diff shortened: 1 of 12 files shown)


ngzhian:
Not proud of the many unreachable in the lexer, I couldn't figure out a way to say that an operation is not yet implemented (other that sprinkling lots of `error lexbuf "unimplemented" in line`. I think it's fine for now, as we implement more operations, those `unreachable` will go away.

binji:
@rossberg can you take a quick look? Thanks!

ngzhian:
@rossberg rebased this recently, ptal, thanks!

interpreter/exec/v128.ml
```diff
@@ -4,6 +4,17 @@ include Simd.Make
     let bytewidth = 16
     let to_string s = s
 
+    let to_i32x4 s =
+      List.map (fun i -> I32.of_bits @@ Bytes.get_int32_le (Bytes.of_string s) i) Simd.f32x4_indices
```
rossberg: Nit: Ocaml's `@@` is confusing with our own use of the operator. We should rename that at some point, but until then I would prefer to avoid using `@@`, esp in cases like this, where it actually makes the line longer.

interpreter/valid/valid.ml
```diff
@@ -95,7 +95,15 @@ let peek i (ell, ts) =
 (* Type Synthesis *)
 
 let type_value = Values.type_of
-let type_unop = Values.type_of
+let type_unop at = function
+    | Values.V128 unop ->
+      let open V128Op in
+      (match unop with
+      | I32x4ExtractLane _ -> V128Type, I32Type
```
rossberg: The idea of the unop/binop/relop/testop abstraction was that they are only used for homogeneously typed operators of the respective kind. That assumption shows up in a couple of places. Extract operations are different, so like for conversions, it's better to introduce a separate AST node for them.

interpreter/exec/v128.ml
ngzhian: Make sense, fixed!

interpreter/valid/valid.ml
ngzhian: In 905d249, introduced a new ast node called ExtractLane. I put this outside of the whole Values.op type, and directly deal with it in the eval.ml, which calls into the SimdOp module logic in eval_numeric, since it knows how to evaluate other simd stuff. LMK if this approach makes sense at all. Thanks for your review!

ngzhian:
I can definitely add `eval_simd` in a follow-up, thanks for the review!

PR #166, opened 2019-12-23
[test] Rename test files of integer min/max ops

Honry:
A followup for https://github.com/WebAssembly/simd/pull/163#discussion_r360507711


Honry:
@tlively, PTAL, thanks!

PR #170, opened 2020-01-08
Add encoding for v128 type and v128.const

ngzhian:
See https://github.com/WebAssembly/simd/blob/master/proposals/simd/BinarySIMD.md for binary encoding of v128 type and v128.const.

### DIFF
--- a/interpreter/binary/encode.ml
+++ b/interpreter/binary/encode.ml
@@ -95,7 +95,7 @@ let encode m =
       | I64Type -> vs7 (-0x02)
       | F32Type -> vs7 (-0x03)
       | F64Type -> vs7 (-0x04)
-      | V128Type -> failwith "TODO v128"
+      | V128Type -> vs7 (-0x05)
 
     let elem_type = function
       | FuncRefType -> vs7 (-0x10)
@@ -134,6 +134,7 @@ let encode m =
     open Memory
 
     let op n = u8 n
+    let simd_op n = op 0xfd; op n
     let end_ () = op 0x0b
 
     let memop {align; offset; _} = vu32 (Int32.of_int align); vu32 offset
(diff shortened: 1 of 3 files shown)


interpreter/binary/encode.ml
```diff
@@ -224,8 +225,7 @@ let encode m =
       | Const {it = I64 c; _} -> op 0x42; vs64 c
       | Const {it = F32 c; _} -> op 0x43; f32 c
       | Const {it = F64 c; _} -> op 0x44; f64 c
-      | Const {it = V128 c; _} ->
-        failwith "TODO v128"
+      | Const {it = V128 c; _} -> simd_op 0x02; Bytes.iter (put s) (V128.to_bits c)
```
rossberg: Nit: for symmetry with the other value types, factor out encoding of the value itself into a separate `v128` function.

Btw, V128.to_bits should return a `string`, not `bytes`. OCaml's choice of type names may be misleading, but strings are byte strings, while the latter type is simply a mutable version of string, which is not what you want for this purpose.

ngzhian: I see, thanks for clarifying. So the reason we want `to_bits` to return a `string` is so that it's immutable and callers cannot modify the underlying bytes?

ngzhian: It looks like we can change https://github.com/WebAssembly/simd/blob/master/interpreter/exec/simd.ml#L36 to `module Make (Rep : RepType) : S with type bits = String.t =`
and have `type bits = string` below. Does that sound reasonable?

rossberg: Yes and yes. You can also use `string` instead of `String.t`.

ngzhian: Done!

ngzhian:
@penzn ptal if you can too, thanks!

interpreter/exec/v128.ml
```diff
@@ -2,6 +2,8 @@ include Simd.Make
   (struct
     include Bytes
```
rossberg: Ideally, the rep type should be string, too (the representation of Wasm values should always be immutable). But we can fix that later.

penzn:
> @penzn ptal if you can too, thanks!

LGTM, I was away until yesterday.

PR #171, opened 2020-01-09
[test] Add spec tests for integer rounding average ops

Honry:
Optimize unknown operator test method to make it as a common method BTW

Sync tests from https://github.com/WAVM/WAVM/pull/250


Honry:
Tests have been reviewed in WAVM, @tlively, @arunetm, pls. check again. Thanks!

PR #172, opened 2020-01-10
Change Rep type of SIMD to be immutable string

ngzhian:
Following up from https://github.com/WebAssembly/simd/pull/170#discussion_r365121635

### DIFF
--- a/interpreter/exec/v128.ml
+++ b/interpreter/exec/v128.ml
@@ -1,9 +1,8 @@
 include Simd.Make
   (struct
-    include Bytes
+    include String
     let bytewidth = 16
-    let of_bits = Bytes.of_string
-    let to_bits = Bytes.to_string
+    let to_string s = s
 
     let of_strings shape ss =
       if List.length ss <> Simd.lanes shape then raise (Invalid_argument "wrong length");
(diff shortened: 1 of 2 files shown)


interpreter/exec/v128.ml
```diff
@@ -1,9 +1,8 @@
 include Simd.Make
   (struct
-    include Bytes
+    include String
     let bytewidth = 16
-    let of_bits = Bytes.of_string
-    let to_bits = Bytes.to_string
+    let to_string s = s
```
rossberg: This def seems unnecessary?

ngzhian: String has no `to_string` function, so this struct doesn't match RepType if we remove this.

rossberg: Oh, I see. But I would suggest to keep the name `to_bits` for this, to avoid confusion with pretty printing.

ngzhian: to_bits is still around, it is defined inside Simd.Make.
This to_string is for pretty print, and if you look at Simd.Make's `to_string`, you can see a FIXME comment, saying that this implementation is very wrong :)
Perhaps I should copy the FIXME here to make it clear that this should be fixed?

rossberg: Yes, I was wondering about the FIXME. I suppose I don't quite see how a fixed implementation would use Rep.to_string instead of doing its own thing on the "bits". So I would remove it until it's actually clear it's needed. But it's all a minor point, so feel free to leave it alone.

ngzhian: Sounds good, I'll leave it alone for now. We can probably change this to print hex values instead. Thanks!

PR #183, opened 2020-01-20
[test] Fix issue #181

### LINKED ISSUE #181: Is `(v128.const)` supposed to parse in the text format?
Some [recently added tests](https://github.com/WebAssembly/simd/blob/8986d5a90200cd8be7c66ff9ccda794f11b3fac6/test/core/simd/simd_load.wast#L187-L212) look like they're asserting that various text pieces such as:

* `(v128.const)`
* `(v128.const 0 0 0 0)`
* `(v128.const i32x4)`

I would have expected these to be `assert_malformed` since they're ideally invalid syntax, but wanted to confirm. @Honry was this perhaps a mistake in the test case generation?

Honry:
- Fix mistake tests in simd_load.wast
- Missing lane index should be malformed
- Test more values of lane index

Fix #181

### DIFF
--- a/test/core/simd/simd_lane.wast
+++ b/test/core/simd/simd_lane.wast
@@ -395,34 +395,34 @@
 
 ;; Malformed lane index value
 
-(assert_malformed (module (func (result i32) (i8x16.extract_lane_s  -1 (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)))) "malformed lane index")
-(assert_malformed (module (func (result i32) (i8x16.extract_lane_s 256 (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)))) "malformed lane index")
-(assert_malformed (module (func (result i32) (i8x16.extract_lane_u  -1 (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)))) "malformed lane index")
-(assert_malformed (module (func (result i32) (i8x16.extract_lane_u 256 (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)))) "malformed lane index")
-(assert_malformed (module (func (result i32) (i16x8.extract_lane_s  -1 (v128.const i16x8 0 0 0 0 0 0 0 0)))) "malformed lane index")
-(assert_malformed (module (func (result i32) (i16x8.extract_lane_s 256 (v128.const i16x8 0 0 0 0 0 0 0 0)))) "malformed lane index")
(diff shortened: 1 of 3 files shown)


Honry:
@alexcrichton, @abrown, @tlively, PTAL, thanks!

test/core/simd/simd_lane.wast
```diff
@@ -393,36 +393,67 @@
   (v128.const i64x2 01_234_567_890_123_456_789_0 0x0_1234_5678_90AB_cdef))
   (v128.const i32x4 0xeb1f_0ad2 0xab54_a98c 0x90ab_cdef 0x1234_5678))
 
+;; Malformed lane index value
+
+(assert_malformed (module (func (result i32) (i8x16.extract_lane_s  -1 (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)))) "malformed lane index")
```
alexcrichton: I may be misinterpreting the intention, but I was under the impression that `assert_malformed` should largely always be used with `(module quote "...")` because in theory this sort of parse error doesn't work in that when we try to parse the `assert_malformed` directive it will fail since the contents are designed to not parse.

abrown:
@Honry, looks good to me. @alexcrichton, should we revert https://github.com/bytecodealliance/wat/pull/45 then?

alexcrichton:
@abrown yeah I think it's safe to revert that, and I should also mention that apart from `assert_malformed` and `module quote ...`, also looks great to me. Thanks @Honry!

test/core/simd/simd_lane.wast
AndrewScheidecker: WAVM is more permissive here than it probably should be: an error in an unquoted module that the parser can recover from without skipping out of the `assert_malformed` will be treated the same as if the module was quoted.

alexcrichton: Searching around https://github.com/webassembly/testsuite it looks like all `assert_malformed` directives are either coupled with `(module quote` or `(module binary` currently.

AndrewScheidecker: @alexcrichton @Honry https://github.com/WAVM/WAVM/commit/165863c4542aaecd93c9de4d84d3fae523a9036e makes it an error to use anything other than a quoted or binary module in `assert_malformed`.

alexcrichton: Nice! That looks great to me, thanks! 👍

Honry: Thank you @AndrewScheidecker, @alexcrichton, I will make a copy of the fix from WAVM to this PR.

Honry: Done!

PR #184, opened 2020-01-22
Adjust two more tests from recent changes

alexcrichton:
The `assert_invalid` directive isn't intended to be used with `(module
quote ...)` and the `module quote` part can be removed here as well
since the module should successfully parse, but fail to validate.


Honry:
Thanks @alexcrichton! @tlively no necessary, as it is out of test generation code, pls. land it. (BTW, sorry for late response.)

PR #193, opened 2020-02-07
Add the notion of a basic SIMD type

dtig:
Add a S128 fixed-width SIMD type, some related macros, and surface this SIMD type in different sections.


document/core/syntax/values.rst
```diff
@@ -145,6 +145,19 @@ Conventions
...
+.. index:: ! simd, fixed-width, vector
+   pair: abstract syntax; packed simd number
+.. _syntax-simd:
+
+Fixed-Width SIMD
+~~~~~~~~~~~~~~~~
+
+Fixed-width SIMD data represents packed integers, or floating-floating point values. The packed values of a given SIMD value must all be of the same type, these packed values can be any of the signed, unsigned integer types, and floating-point types ocurring in this proposal.
+
+The number of packed values :math:`p` can be determined by the bit width :math:`N` of the packed value type.
```
binji: Define this as "lanes" here?

document/core/syntax/values.rst
```diff
@@ -145,6 +145,19 @@ Conventions
...
 * The meta variable :math:`z` ranges over floating-point values where clear from context.
 
+.. index:: ! simd, fixed-width, vector
+   pair: abstract syntax; packed simd number
+.. _syntax-simd:
+
+Fixed-Width SIMD
+~~~~~~~~~~~~~~~~
+
+Fixed-width SIMD data represents packed integers, or floating-floating point values. The packed values of a given SIMD value must all be of the same type, these packed values can be any of the signed, unsigned integer types, and floating-point types ocurring in this proposal.
```
binji: nit: maybe "... must be all of the same type, which can be any of the ... ". It is probably worth mentioning the types explicitly here too.

document/core/exec/runtime.rst
```diff
@@ -7,14 +7,14 @@ Runtime Structure
...
-.. index:: ! value, constant, value type, integer, floating-point
+.. index:: ! value, constant, value type, integer, floating-point, fixed-width simd
    pair: abstract syntax; value
 .. _syntax-val:
 
 Values
 ~~~~~~
 
-WebAssembly computations manipulate *values* of the four basic :ref:`value types <syntax-valtype>`: :ref:`integers <syntax-int>` and :ref:`floating-point data <syntax-float>` of 32 or 64 bit width each, respectively.
+WebAssembly computations manipulate *values* of the five basic :ref:`value types <syntax-valtype>`: :ref:`integers <syntax-int>` and :ref:`floating-point data <syntax-float>` of 32 or 64 bit width each  respectively, and :ref: `SIMD data <syntax-simd>` of 128 bit width representing packed integers of floating-point values.
```
binji: We should define "packed integers" explicitly. I thought it may have been used in the memory instruction section, but that just refers to "storage size" which is not quite what we want here.

dtig: I agree, just not sure where the right place for them would be. I'm wondering if just having them be in document/core/syntax/values.rst under a `Representations` section make sense? I initially thought not, because it's mostly addressing the structure of types - but it would make sense to have them there because <syntax-simd> would point to there.

document/core/syntax/values.rst
dtig: Fixed, and added a sentence about the explicit types.

document/core/syntax/values.rst
dtig: Done.

document/core/intro/overview.rst
```diff
@@ -23,6 +23,13 @@ This language is structured around the following concepts.
   Instead, integers are interpreted by respective operations
   as either unsigned or signed in two’s complement representation.
 
+  In addition to the basic value types above, a single 128 bit wide
+  value type is added to represent different types of packed data.
+  The supported representations are 4 32 bit, or 2 64 bit
+  |IEEE754|_ numbers, or different widths of packed integer values
+  specifically two 64 bit integers, four 32 bit integers, eight
```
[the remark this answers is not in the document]
dtig: I agree, left it out to be consistent with the rest of the text - using the hyphenated style for this section now.

document/core/exec/runtime.rst
binji: @rossberg thoughts?

dtig: @rossberg Explicitly linking from #195 to address syntax issue - looking for feedback on how to address the different representations better here so that can be incorporated into subsequent PRs. I'm looking for the right place to express different representations with the general syntax of `{t}{lane_width}x{lanes}` where `t` is the type, `lane_width` is the number of bits per packed value, and `lanes` is the number of packed values.

rossberg: The definition of iN in syntax/values is generic in N, so already defines i8 and i16. Likewise, all the ops in the numerics section are generic.

See comment below regarding abstract syntax of SIMD literals. The definition here has to match the abstract syntax of the const instruction.

document/core/intro/overview.rst
```diff
@@ -23,6 +23,13 @@ This language is structured around the following concepts.
   Instead, integers are interpreted by respective operations
   as either unsigned or signed in two’s complement representation.
 
+  In addition to the basic value types above, a single 128 bit wide
+  value type is added to represent different types of packed data.
```
rossberg: Avoid formulations like "is added", since once in the spec, a feature "just is".

document/core/syntax/values.rst
```diff
@@ -145,6 +145,19 @@ Conventions
...
 * The meta variable :math:`z` ranges over floating-point values where clear from context.
 
+.. index:: ! simd, fixed-width, vector
+   pair: abstract syntax; packed simd number
+.. _syntax-simd:
+
+Fixed-Width SIMD
+~~~~~~~~~~~~~~~~
+
+Fixed-width SIMD data represents packed integers, or floating-point values. The packed values of a given SIMD value must all be of the same type, which can be any of the signed, unsigned integer types, and floating-point types ocurring in this proposal. The supported representations of the packed values are 4 32-bit, or 2 64-bit |IEEE754|_ numbers, or different widths of packed integer values specifically 2 64-bit integers, 4 32-bit integers, 8 16-bit integers, or 16 8-bit integers.
```
rossberg: The abstract syntax needs to have one unique way of representing each SIMD value. Reinterpretation has to be expressed by explicit meta functions, similar to the "signed" function for integers.

The most canonical choice for a representation would simply be as a 128-bit integer, i.e. the spec just uses `i128` for SIMD values. Then this section doesn't need to define anything, except maybe include a note saying that i128 is also used to represent SIMD values.

At some place (probably along with the SIMD ops) the spec then has to define interpretation functions like `i16x8` that convert this to a sequence i16^8 and so on. For example,
```
i64x2(i) = i_1 i_2
where
  b* = bytes_i128(i)
  i_1 = bytes_i64^{-1}(b*[0 \slice 4])
  i_2 = bytes_i64^{-1}(b*[4 \slice 4])
```

The definition of iN above is generic, so already defines i8, i16, i128. Likewise, all the ops in the numerics section are generic.

document/core/exec/runtime.rst
```diff
@@ -26,7 +26,8 @@ It is convenient to reuse the same notation as for the |CONST| :ref:`instruction
      \I32.\CONST~\i32 \\&&|&
      \I64.\CONST~\i64 \\&&|&
      \F32.\CONST~\f32 \\&&|&
-     \F64.\CONST~\f64
+     \F64.\CONST~\f64 \\&&|&
+     \S128.\CONST~\s128
```
rossberg: Unfortunately, you cannot use `s128` as the meta variable for SIMD constants, since that already is defined to be a signed 128-bit integer. However, if you represent SIMD values as i128 (see comment below) you could use that here.

ngzhian:
Closed via #269

PR #197, opened 2020-02-20
[test] Add spec tests for integer abs ops

Honry:
Tests are pulled from https://github.com/WAVM/WAVM/pull/260


Honry:

test/core/simd/simd_i16x8_arith2.wast
```diff
@@ -16,16 +17,18 @@
...
-  (func (export "i16x8.max_s_with_const_15") (param v128) (result v128) (i16x8.max_s (local.get 0) (v128.const i16x8 0 0 1 1 2 2 3 3)))
-  (func (export "i16x8.max_u_with_const_16") (param v128) (result v128) (i16x8.max_u (local.get 0) (v128.const i16x8 -32768 -32768 32767 32767 16384 16384 65535 65535)))
-  (func (export "i16x8.max_u_with_const_17") (param v128) (result v128) (i16x8.max_u (local.get 0) (v128.const i16x8 0 0 1 1 2 2 3 3)))
-  (func (export "i16x8.avgr_u_with_const_18") (param v128) (result v128) (i16x8.avgr_u (local.get 0) (v128.const i16x8 -32768 -32768 32767 32767 16384 16384 65535 65535)))
-  (func (export "i16x8.avgr_u_with_const_19") (param v128) (result v128) (i16x8.avgr_u (local.get 0) (v128.const i16x8 0 0 1 1 2 2 3 3)))
+  (func (export "i16x8.abs_with_const_10") (result v128) (i16x8.abs (v128.const i16x8 -32768 -32768 32767 32767 16384 16384 65535 65535)))
+  (func (export "i16x8.abs_with_const_11") (result v128) (i16x8.abs (v128.const i16x8 65535 65535 16384 16384 32767 32767 -32768 -32768)))
```
tlively: I'm not sure it makes sense to export multiple "abs_with_const" functions, since abs only takes a single argument anyway. We could instead just export a single "abs" function then pass it different constants in the asserts.

Honry: Indeed, I will remove one of them.

Honry: Done.

tlively: I'm still not sure what the benefit of the remaining "abs_with_const" test is. Is there precedence in other spec tests for exporting a function applying a unary op to a const?

Honry: We have similar tests at https://github.com/WebAssembly/simd/blob/master/test/core/simd/simd_f32x4.wast#L29, which was suggested by @AndrewScheidecker 
 at https://github.com/WAVM/WAVM/pull/194#issuecomment-534553081, sorry I couldn't remember the benefit of this checkpoint, @AndrewScheidecker, could you help clarification this?

AndrewScheidecker: For WAVM, the function body is optimized, but the combination of the function body with the constant argument from the invoke is not. I assume that's true in general for other runtimes, though maybe those that use the test scripts by translating them to JavaScript have the potential to inline the call that the invoke is translated to.

Putting a constant in the function body is meant to test that there aren't any bugs in constant folding, or the optimizer in general.

Honry: @AndrewScheidecker, thank you!  @tlively, is that answer your concern?

PR #198, opened 2020-02-20
Update WAVM implementation status

Honry:


Honry:
@AndrewScheidecker, @tlively, PTAL, thanks!

PR #207, opened 2020-03-31
[test] Fixed mismatched op names in comments

Honry:
Pulled from https://github.com/WAVM/WAVM/pull/269


Honry:

Issue #129, opened 2019-10-30
Final opcode renumbering process


tlively:
Since we plan on softly freezing our instruction set when we reach stage 3, that will also be a good time to finalize the opcode names and numbers before many engines dive into their implementation work.

It won't be possible to synchronize a renumbering across the entire ecosystem all at once because different projects have different release cycles, but the closer we can come to approximating a coordinated release, the shorter the period of disruption will be for our bleeding-edge users. With that in mind, I would like to propose the following schedule of events.

1. We decide we are happy pushing to stage 3 and freezing our opcode set.
2. We decide on final names and opcodes.
3. We push for stage 3 in the CG.
4. We recommend that engines and tools update their naming and numbering no earlier than three weeks and no later than four weeks after the vote for stage 3.

Steps 2 and 3 can happen in parallel, but we will aim to have finished finalizing names and opcodes by the time of the vote for stage 3 or as soon as possible afterward. Our recommendation of renumbering window will be shared with the CG after the vote and also announced at that time in a dedicated issue in this repo.

The time span of three to four weeks after reaching stage 3 is meant to give implementors adequate time to implement any necessary changes.

zeux:
Just a quick note: I have come to regret changing v8x16.shuffle opcode. Defacto it’s still 0xfd 0x03 in LLVM/Chrome and the difference doesn’t help during development - e.g. I have a local WABT build with the old opcode...

If the renumbering will actually affect many opcodes then it doesn’t matter because we’re going to go through a major breakage period, but if only a few opcodes are going to get moved then we could consider moving it back to 0x03.

tlively:
Thanks, @zeux! I agree that in hindsight changing the shuffle opcode wasn't helpful, but I don't think it's critical to put it back right now. We have had a few bug reports of the mismatch between wabt and other projects, but it's probably not worth a patch since the opcode freeze and renumbering is hopefully not too far off anyway.

tlively:
We ended up doing a soft freeze of the opcodes after we went to phase 3, but the renumbering can be discussed concretely in https://github.com/WebAssembly/simd/blob/new-opcodes/proposals/simd/NewOpcodes.md. We should recommend that implementations aim to make the switch between three and four weeks after the renumbering is merged.

lars-t-hansen:
I commented on the PR too but I find the justifications for a renumbering at this stage to be fairly meagre.

PR #48, opened 2018-10-26
Reorganize opcodes by kind and type

tlively:
This is yet another opcode renumbering.

I tried to take into account all the comments on #42, particularly @binji's comment about trying to be consistent with the MVP opcode ordering. I also left explicit gaps in the chart for what I think should be reserved opcodes to make the reserved opcodes easier to see and talk about for the time being.

Feedback welcome!

cc @dtig

Issue #46, opened 2018-10-23
instruction encoding


tlively:
The proposal says, "All SIMD instructions are encoded as a 0xfd prefix byte followed by a SIMD-specific opcode in LEB128 format," but this is not consistent with the instruction encoding in other proposals. For example, the non-trapping float-to-int and threads proposals both encode their instructions as a single prefix byte followed by a single byte identifying the instruction.

The only reason I see to use LEB128 for SIMD instructions is to avoid running out of opcodes. But we don't seem to be in danger of running of opcodes for 128-bit SIMD, and if we were ever going to add many more SIMD instructions we could always use more prefixes as well.

Since this divergence from the norm seems unnecessary, I propose we change the encoding of SIMD instructions to be just two bytes, as in other proposals. Is there any context for this discussion that I am missing?

Adopting this change would obviate LLVM this LLVM bug https://bugs.llvm.org/show_bug.cgi?id=39272.

binji:
We discussed this in the [May 2017 CG meeting](https://github.com/WebAssembly/meetings/blob/master/2017/CG-05.md#webassembly-threads-proposal), and decided on a single prefix byte followed by LEB128 code. We can revisit this, but generally we only do so if there is new information.

If the other proposals are using a single byte followed by a single byte for opcode, then they should be updated to allow for a LEB128 encoding. We can generally encode this using a single byte, but we should allow for "long" encodings as well.

tlively:
It still seems weird to me, since all of the MVP opcodes are bytes rather than LEB128 values. SIMD is unique among proposed extensions in that it has enough opcodes that using LEB128 to encode them will require more space than just using two bytes.

kmiller68:
How many SIMD instructions do we actually expect to be in a program? I would guess it makes up less than .1% of the bytes of the module? Maybe I'm wildly off here though. If it's only a few bytes total one way or the other it seems kinda silly worry about it for space reasons.

sunfishcode:
My memory of the discussion @binji mentioned above was that "prefix plus LEB128" was meant to be a rule that all prefixes could follow, for the sake of consistency, rather than having each prefix have its own opcode encoding scheme.

Fortunately, I think the LLVM issue is fixable (and have commented in the bug).

tlively:
[...]

PR #209, opened 2020-04-01
Opcode renumbering

tlively:
This PR is for discussion only and should not be merged. Once we have
reached consensus on the renumbering, the existing documents should be
updated to reflect the new opcodes.

The goals of this renumbering are:

 1. To maintain consistency with the ordering of MVP operations

 2. To organize sets of similar operations into tables such that the offset
 from an operation for one type to the same operation for the next
 type is constant for all operations in that set.

 3. To use round hexadecimal numbers for offsets where not too wasteful.

[Rendered](https://github.com/WebAssembly/simd/blob/new-opcodes/proposals/simd/NewOpcodes.md)


proposals/simd/NewOpcodes.md
```diff
@@ -0,0 +1,124 @@
...
+| i8x16.shr_s          | 0x6c   | i16x8.shr_s              | 0x8c   | i32x4.shr_s              | 0xac   | i64x2.shr_s | 0xcc   |
+| i8x16.shr_u          | 0x6d   | i16x8.shr_u              | 0x8d   | i32x4.shr_u              | 0xad   | i64x2.shr_u | 0xcd   |
+| i8x16.add            | 0x6e   | i16x8.add                | 0x8e   | i32x4.add                | 0xae   | i64x2.add   | 0xce   |
+| i8x16.add_saturate_s | 0x6f   | i16x8.add_saturate_s     | 0x8f   | ---- add_sat ----        | 0xaf   | ----        | 0xcf   |
+| i8x16.add_saturate_u | 0x70   | i16x8.add_saturate_u     | 0x90   | ---- add_sat ----        | 0xb0   | ----        | 0xd0   |
+| i8x16.sub            | 0x71   | i16x8.sub                | 0x91   | i32x4.sub                | 0xb1   | i64x2.sub   | 0xd1   |
+| i8x16.sub_saturate_s | 0x72   | i16x8.sub_saturate_s     | 0x92   | ---- sub_sat ----        | 0xb2   | ----        | 0xd2   |
+| i8x16.sub_saturate_u | 0x73   | i16x8.sub_saturate_u     | 0x93   | ---- sub_sat ----        | 0xb3   | ----        | 0xd3   |
+| ---- dot ----        | 0x74   | ---- dot ----            | 0x94   | i32x4.dot_i16x8_s        | 0xb4   | ----        | 0xd4   |
```
ngzhian, line 95: What is this op? Is it #20  ?

proposals/simd/NewOpcodes.md
```diff
@@ -0,0 +1,124 @@
...
+| f32x4.mul       | 0xe6   | f64x2.mul       | 0xf6   |
+| f32x4.div       | 0xe7   | f64x2.div       | 0xf7   |
+| f32x4.min       | 0xe8   | f64x2.min       | 0xf8   |
+| f32x4.max       | 0xe9   | f64x2.max       | 0xf9   |
+| ---- pmin ----  | 0xea   | ---- pmin ----  | 0xfa   |
+| ---- pmax ----  | 0xeb   | ---- pmax ----  | 0xfb   |
+
+| Conversion Op           | opcode |
+| ----------------------- | ------ |
+| i32x4.trunc_sat_f32x4_s | 0x100   |
```
ngzhian, line 121: nit: can this be written as 0x0100? It makes it slightly clearer that it's 2 bytes.

proposals/simd/NewOpcodes.md
tlively: #127

lars-t-hansen:
It's unclear to me what utility this renumbering has; the goals that are stated have uncertain value and are to some extent circular.  Would anyone be using a decoder based on these constant offsets?  Would anyone do a reverse lookup from opcode into a table to find the meaning of the opcode instead of just searching a text file for the opcode value?

zeux:
Here's the practical problems that I foresee with this:

- Existing binary blobs will have to be recompiled
- For a period of a month or more WASM SIMD will be very hard to use because of the mismatch in tools. This may be exacerbated by the lag time in delivery due to COVID on the Chrome side
- Existing WASM feature detectors will likely misfire - most of them use some simple opcode like a load that will likely still validate (I haven't verified this with the proposed encoding scheme, but this is probable).

Here's an alternative possibility:

- Renumber the shuffle back to 0x03 because toolchains use 0x03 anyhow
- Possibly find new numbers for opcodes that aren't stable yet, e.g. dot, bitmask, pmin/pmax, if the existing numbers are suboptimal and there are gaps to fill

It would be good to understand the pros that this numbering has - my mental model has certainly been that the encoding adjustments would be very minor. If the decoder needs to classify opcodes, it's straightforward to build a table (right now the table can be of size 256!) that maps opcode to any information desired by the encoder, which can be arbitrarily complex and not limited by the proposed grouping.

binji:
Is there a concern about renumbering in general, or about this specific renumbering? At the very least I'd like to see some opcode grouping in the encoding, if only because it's easy to miss which instructions are even there in the current ordering. For example, I'd like to have the load splat and load extend instructions grouped with `v128.load`. It would be nice to group `v128.andnot` with the other `v128.*` instructions too.

I agree it's not important for decoders (we'll all just use a switch statement or table), but I think there is value for people understanding the spec. If only for folks looking at opcode tables (which I do fairly often, but perhaps I'm an outlier :smile:)

lars-t-hansen:
I think my concern is entirely general - there doesn't seem to be any good justification for doing this.  Clearly for documentation & exposition we can group operations logically, as the existing overview does very well; the mapping to the encoding doesn't seem to matter though.  Having had a lot to do with the encoding of the existing non-SIMD instructions all I can say is that any logical grouping of the opcodes is not something I pay any attention to.

zeux:
In terms of understanding the spec, I feel like we can just sort the opcode table *not* according to the opcode number, but according to the implied easy-to-understand order. You could imagine multiple orders that make sense and it's not clear to me which one is best - for example, is it more useful to group by type or group by operation? (do you consolidate all shifts for all types or not) Because it's hard to pick one, it's not clear to me that we need to reflect this order in the opcode sort.

There's something to be said about consistency of course, but it just doesn't seem that opcode numbers are very important? x64 has a crazy encoding but it's not because the numbers aren't sequenced, it's because there's a billion hard to understand modes, and it doesn't seem to be truly important whether 0xE9 is a CALL or a JUMP.

zeux:
Just to be clear, my main concern with this is that this is going to cause churn for everyone involved - browser vendors, toolchain providers, application developers (early adopters) - with potentially hard to understand validation errors, and I'm expecting that it will take a few months to fully resolve this, because of delivery lag times in various components.

binji:
Those are fair arguments, and if we stay with the current opcode ordering it probably won't affect much.

Though I will say we have some precedent for a change like this. We renumbered the opcodes in the 0xd version of wasm too: see https://github.com/WebAssembly/design/pull/826, which references a [spreadsheet describing the ordering](https://docs.google.com/spreadsheets/d/1RglIQM97cP83Sy-J3BT3D7IxdcH3QCKftiM1bjOL2vc/edit#gid=0).

> You could imagine multiple orders that make sense and it's not clear to me which one is best - for example, is it more useful to group by type or group by operation?

We would probably want to base this on the organization of the v1 opcode encoding. In particular, the instructions are categorized by kind (control, parametric, binary, etc.), then type (i32, f32, etc.) then operation.

> Just to be clear, my main concern with this is that this is going to cause churn for everyone involved 

Agreed, but this was churn that was originally planned back in October, in issue #129. It didn't seem like there was much disagreement then, are the Chrome release plans (and COVID stuff) the main thing that has changed?

I was at least under the assumption that newer opcodes were added a bit more haphazardly since we knew that we'd reorder them later anyway. It seems a shame to leave them as-is given that.

tlively:
There is also precedent for this in the SIMD proposal itself, particularly in #48. When we did that reorganization we were careful to leave holes for opcodes that had not yet been merged into the proposal. We thought that that would be the final renumbering, and we did not anticipate all the new opcodes that have been added in the last year and a half.

> I was at least under the assumption that newer opcodes were added a bit more haphazardly since we knew that we'd reorder them later anyway. It seems a shame to leave them as-is given that.

I strongly agree with this. We've been tacking opcodes onto the end in the order they are merged to the proposal, resulting in a an ordering that is completely inconsistent with the ordering of preexisting opcodes. If we hadn't been planning on a final renumbering, we would have done this differently by leaving lots of opcode space available for future opcodes to fit into the existing organization.

> Just to be clear, my main concern with this is that this is going to cause churn for everyone involved... and I'm expecting that it will take a few months to fully resolve this

As described in #129, the plan is to give implementors enough time to implement the renumbering then try to synchronize the update across the ecosystem. There will be lag from projects having multiple distribution channels that update at different times, but I believe we will be able to synchronize the release of many projects on their primary early-adopter distribution channels to keep the churn to about a week.

> I feel like we can just sort the opcode table *not* according to the opcode number

> the mapping to the encoding doesn't seem to matter though

I personally care about the numbering to the extent that having constant offsets between the same operations on different types makes the LLVM instruction selection code waaaay simpler. I would also be sad if the opcodes were not in order in the spec as an aesthetic matter.

zeux:
> Agreed, but this was churn that was originally planned back in October, in issue #129. It didn't seem like there was much disagreement then

It wasn't clear from that issue what the extent of renumbering was, I guess. It could be as innocuous as "keep status quo" and as extensive as, well, this proposal :) We can still decide it's the right thing to do but I feel like we should be very explicit about what we gain and why it's not a documentation-only issue.

> but I believe we will be able to synchronize the release of many projects on their primary early-adopter distribution channels to keep the churn to about a week.

If we can synchronize the release window to be short then I don't have any other concerns. But I'm wondering how practical this is - e.g. node.js ships a stale version of v8 even in the nightlies.

jgravelle-google:
> I was at least under the assumption that newer opcodes were added a bit more haphazardly since we knew that we'd reorder them later anyway. It seems a shame to leave them as-is given that.

Strongly agree with this. I think that as a general approach for standardizing binary formats (thinking ahead to what we'll do for Interface Types), we should incrementally add opcodes as a proposal evolves. This will mean things are scattered and unordered, but should reduce incremental churn. Then, before finalizing the proposal, renumber things to be clean, organized, and non-ugly. This matches what we did for wasm itself.

The core question is: does it matter that the opcodes be numbered in a "non-ugly" way? I don't feel strongly about that as an absolute good, but I see it as a matter of quality. This is, necessarily, a highly non-technical argument. I think we should want proposal creators to feel a strong sense of pride in their work. And I think that necessitates leaving some room for aesthetics and taste. Technical concerns trump style, but I think this is a place where any encoding is equally good technically, so we might as well do something pleasant here.

I am also of the opinion that anyone using experimental features does so at their own peril. We can't break the open web. If we can't break experimental features past a certain point, that puts undue pressure on people to get things right ahead of time, rather than getting something working and iterate. Concretely, I can imagine people taking the time to fully map out all their possible opcodes and coming up with a sensible scheme for them, and leaving room for things they plan to add later. Which is doing more work up-front that would be better served by being informed by end-to-end practical experience.

> we would have done this differently by leaving lots of opcode space available for future opcodes to fit into the existing organization

Oh hey this was commented while I was writing it up. Which, yeah that.

zeux:
> I was at least under the assumption that newer opcodes were added a bit more haphazardly since we knew that we'd reorder them later anyway. It seems a shame to leave them as-is given that.

That's a good point. What's the plan for future additions, though? By necessity I'd expect we'll need to evolve the spec over the next few years, by shipping extra additions e.g. fast instructions - is that going to use a separate prefix, or just appending to the existing set?

tlively:
> But I'm wondering how practical this is - e.g. node.js ships a stale version of v8 even in the nightlies.

Projects like node that have long release cycles are always somewhat stale compared to upstream V8, but that's already a problem for using newly-implemented instructions, so I don't think the extra pain from the renumbering is a huge issue there.

> What's the plan for future additions, though?

I would expect future additions would either use a new prefix or use subsequent opcode space with the same prefix. That choice would depend on the actual content of the future proposals. But I'm not worried about trying to consider future proposals in this ordering; I expect the contents of future proposals would be different enough from the contents of this proposal that they would warrant their own groupings.

zeux:
> I am also of the opinion that anyone using experimental features does so at their own peril. 

While this is certainly true, I just shudder at the implications of this. Maybe that's what it is.

Today I can use Wasm simd with Chrome, node.js and Emscripten. If there's a period of time where Chrome Canary / node.js nightly / Emscripten don't agree on the encoding subset, it's going to be impossible to do any SIMD related work. If this period is going to last a month or more, it feels like a significant step back from the status quo where a large subset of the spec works, can be tested, etc.

I'm fine with using latest nightly builds of various tools to get stuff done, but I'm worried about having to wait for months for everything to get back to where we are today (which is "things basically work").

zeux:
> I don't think the extra pain from the renumbering is a huge issue there.

The entire -msimd subset works on node today as far as I'm aware, as well as a few instructions that are in -munimplemented-simd. So this is currently not a big problem in practice.

I guess it would be feasible for existing users (like, uh, me, which is why I'm worried!) to keep using the older Emscripten version until node/Chrome catch up.

proposals/simd/NewOpcodes.md
```diff
@@ -0,0 +1,124 @@
...
+| f32x4.max       | 0xe9   | f64x2.max       | 0xf9   |
+| ---- pmin ----  | 0xea   | ---- pmin ----  | 0xfa   |
+| ---- pmax ----  | 0xeb   | ---- pmax ----  | 0xfb   |
+
+| Conversion Op           | opcode   |
+| ----------------------- | -------- |
+| i32x4.trunc_sat_f32x4_s | 0x0100   |
+| i32x4.trunc_sat_f32x4_u | 0x0101   |
+| f32x4.convert_i32x4_s   | 0x0102   |
+| f32x4.convert_i32x4_u   | 0x0103   |
```
zeux, line 124: Is there any way to put this section somewhere else to avoid double-byte encoding? It feels like it would make encoding more uniform, would make opcode-based lookup tables have a small fixed size without extra range checks, and would make it possible to patch the bytecode in place vs the old encoding for compatibility purposes.

zeux: Ohh, or is it the case that right now the opcode is actually not encoded as a byte, since LEB128 encoding uses two bytes for any opcodes >=0x80? :-/

tlively: Yes, all the bytecodes are LEB128 so we already have opcodes that take 2 bytes in the current numbering.

dtig:
Maybe this has already been mentioned before, but one of the motivating reasons was that while this has no bearing on applications, and engines having consistent offsets makes the tools implementation non-ideal. Perhaps @tlively already mentioned this and I missed it. Especially as SIMD has many regular operation spread across different types, making this coherent and consistent was also a factor here. 

We could have done this from the beginning, but given that there were many applications relying on experimental support, to get quick feedback at the time it made most sense to tack on operations at the end so applications aren't affected by toolchain/engine changes as they just won't be using those opcodes. Also, renumbering opcodes at different intervals seemed like unnecessary churn (ex: swizzle opcode) when we could do it in a more planned way when we have a good idea of the operations being introduced, and frozen addition of new opcodes so we can renumber them with a one-time overhead. Any future additions as a part of a separate proposal will have their own prefix, so shouldn't affect this set of opcodes.

This PR is mainly proposing the new opcodes, but the actual implementation can be co-ordinated with dates we all agree on, post COVID release freezes to minimize any churn that applications have to deal with. Very specifically for Chrome Canary / node.js nightly / Emscripten - the plan on our end is to co-ordinate this tightly so, the engine change will land in V8 as soon as the Emscripten change lands - Chrome canary and Emscripten should be in sync within 48 hours, for node.js nightly I'm not very clear on the schedule for different platforms, so that's a little uncertain but hopefully should have an answer for this soon. Even though this is still in experimental, the intent was to do this in a way that we can plan it in conjunction with other tools/engines to minimize user pain.

proposals/simd/NewOpcodes.md
zeux: This is unfortunate. The existing Wasm spec doesn't do this as far as I understand. We have 170 opcodes right now. I'm not sure why the LEB128 encoding was chosen, is there a precedent for this elsewhere in Wasm?

zeux:
I'm for this if we can minimize the churn 👍

proposals/simd/NewOpcodes.md
tlively: I had the same question in #46. TL;DR: this is how all proposals work since 2017, but SIMD is the only one where you can tell the difference. Notably, V8 was mistakenly using bytes instead of LEB128s until a couple weeks ago: https://bugs.chromium.org/p/v8/issues/detail?id=10258.

zeux: So to clarify, the baseline wasm spec doesn't use LEB128 for instructions but all extensions do? A bit confused as to how v8 managed to work with byte encoding if there are a lot of opcodes today over 0x80?

zeux: Also I'm a bit sad that we're doing LEB128. It hurts code size (maybe in a minor way), it makes decoding slightly more involved (why?).

If we introduce *more* opcodes, we need to be introducing them in groups as we discussed in prior calls, because just adding new opcodes incrementally is impossible to deal with from the application side.

So on the encoding perspective, it seems straightforward to dedicate, say, 0xf0-0xff to prefixes for future extensions. This can work well in that when we introduce a fast SIMD subset, it's just going to take 0xf1 or smth and evolve that independently from any other subsets.

Of course there's a risk of prefix-based approach eventually evolving into something odd, and the parsing of extensions gets more uniform if we just go with LEB128. Still, it's odd to see this diverging from the Wasm spec approach.

zeux: I'm still trying to understand this (sorry for the spam). You're saying that the spec says that opcodes should be encoded with LEB128 but the *entire ecosystem* chose to use bytes right now? That is - this isn't just a v8 issue, it's also an issue on LLVM side, wabt, et al - *all* tools that interact with SIMD must be assuming bytes?

zeux: Dear god. Thanks, this makes more sense now!

zeux:
I wrote a polyfill for this :D https://gist.github.com/zeux/10fa3a3c85262ac87979887cb02e0c66

(the polyfill doesn't have real remapping tables yet, but it should be straightforward to fix this code to either remap old=>new or new=>old by just implementing the remap)

proposals/simd/NewOpcodes.md
lars-t-hansen: > I had the same question in #46. TL;DR: this is how all proposals work since 2017, but SIMD is the only one where you can tell the difference. Notably, V8 was mistakenly using bytes instead of LEB128s until a couple weeks ago: https://bugs.chromium.org/p/v8/issues/detail?id=10258.

Same bug in Firefox, found it when I implemented SIMD :-}

proposals/simd/NewOpcodes.md
```diff
@@ -0,0 +1,124 @@
...
+| v32x4.load_splat   | 0x09   |
+| v64x2.load_splat   | 0x0a   |
+| v128.store         | 0x0b   |
+
+| Basic instruction    | opcode |
+| -------------------- | ------ |
+| v128.const           | 0x0c   |
+| v8x16.shuffle        | 0x0d   |
+| v8x16.swizzle        | 0x0e   |
+| i8x16.splat          | 0x0f   |
```
zeux, line 21: The ordering here is mildly unsatisfying for decoding purposes. Similarly to the existing spec, this interleaves splat/extract/replace lane, and they have different encoding formats. Thoughts on grouping these by instruction type instead? (splats followed by lane operations)

tlively: Yeah I think that sounds reasonable 👍

tlively:
@zeux is the latest commit the change you had in mind?

zeux:
> @zeux is the latest commit the change you had in mind?

Yup 👍

dtig:
We are getting back to a (somewhat) normal Chrome release cycle going forward, so I'm hoping we can find a good way to coordinate landing, and propagating this to different implementations. 

Are there any outstanding concerns with the renumbering process? If so, please respond so we can address them. Of the current in progress implementations I know of that this will affect, most of the folks involved have responded to this issue. The one thing I'm uncertain of is the Rust toolchain - @tlively perhaps you may know if this would be a breaking change for them, or is there someone working on the SIMD parts that we can alert? 

The current plan I'd like to suggest for Chrome/V8, toolchain to minimize breakage time for applications using Chrome canary for prototyping is - 
1. V8 opcode renumbering lands on a Friday, so by Monday the Dev/Canary branches for most platforms us updates.
2. LLVM renumbering lands on the Monday after.

penzn:
> 1. V8 opcode renumbering lands on a Friday, so by Monday the Dev/Canary branches for most platforms us updates.
> 2. LLVM renumbering lands on the Monday after.

What about wabt?

tlively:
cc @alexcrichton for insight into how this will affect Rust tools.

cc @binji for how it makes sense to update WABT.

dtig:
> > 1. V8 opcode renumbering lands on a Friday, so by Monday the Dev/Canary branches for most platforms us updates.
> > 2. LLVM renumbering lands on the Monday after.
> 
> What about wabt?

Defer to @binji, but it would make sense for updating WABT possibly at the same time as LLVM. Happy to submit a PR to have it ready to go depending on which week we decide to push for this.

zeux:
Would be good to also have a binary Emscripten release soon after the LLVM update, so that Emscripten users don't need to build LLVM themselves.

alexcrichton:
Thanks for the ping! Is there a preview LLVM patch to poke around? For the Rust toolchain we do not follow LLVM tip-of-tree (but rather releases), so if we want functionality from tip-of-tree we need to backport it. We are currently using LLVM 9 and are trying to upgrade to LLVM 10, but ideally the patch would be back-portable to LLVM 9 for us to use. (wanted to poke around and see if we can handle this)

dtig:
Perhaps @kripken can comment on the possibility of an Emscripten release after the LLVM update?

tlively:
@alexcrichton There's not LLVM patch yet, but I can take care of preparing a Rust-LLVM PR as soon as it lands upstream to minimize the lag time. Will there be issues with other tools like wasm-pack, etc. which use non-LLVM libraries?

kripken:
@dtig 

> Perhaps @kripken can comment on the possibility of an Emscripten release after the LLVM update?

Sure, please ping me at that time. Should be no problem to do that very soon after LLVM (and Binaryen?) update.

alexcrichton:
@tlively it'll take some time to update everything, and we're probably in for a week or so of breakage, but that's probably ok. We don't have too too many libraries to update so it'll mostly just take time for the changes to propagate.

tlively:
It probably makes sense to push to LLVM upstream on the Friday so we can do an Emscripten release on Monday and get it into Rust's nightly build by Monday as well.

@sunfishcode will wasi-sdk be updated to reflect the SIMD opcode renumbering or will it pick it up with the next LLVM release?

binji:
Should be a pretty straightforward change for wabt, we should be able to land it anytime since IIUC it's primarily used by developers anyway.

tlively:
@binji is there a WABT release process we can try to line up or do folks generally build from source to get the latest version?

binji:
Often folks build from source, but I occasionally cut a new release too. It's much more convenient for users with a release, since they can just download the binaries from GitHub. It's been a few months since the last release, so I can cut a new one after we make the change.

proposals/simd/NewOpcodes.md
dtig: Looking at this again as I'm updating engine opcodes, I would also like to express a preference for the decoded opcodes to be 1-byte because we have the space. The reason is that on the engine side, this will add some bloat to opcode/signature tables which assume 1-byte opcodes. It's not terrible to update, but considering we have the space, and it's just the 4 opcodes that are spilling over I'd like to avoid it as it looks like it's possible in this case.

tlively:
I've just renumbered the float and conversion opcodes to make sure all opcodes stay below 0xff. This will keep opcode decoding tables smaller in some engines. Thanks for the suggestion, @dtig!

tlively:
LLVM patch at https://reviews.llvm.org/D79224

binji:
I noticed that `simd_const.wast` uses the `v128.const` encoding directly (as `0xfd 0x02`), you may want to fix that in this PR too.

tlively:
Thanks, I added a note about that to the [tracking issue](https://github.com/WebAssembly/simd/issues/216)

lars-t-hansen:
Since `i32x4.dot_i16x8_s` is not approved yet, and in keeping with the situation for the bitmask opcodes which are also pending, it should not be in this table so as to avoid any confusion about what instructions are in the agreed-upon instruction set.

tlively:
@lars-t-hansen Thanks for the suggestion. I've fixed that and also marked the PR as draft to make it more obvious that it won't be merged. The other docs that already exist in the repo will need to be updated instead.

ngzhian:
I think we pretty much have consensus on this renumbering, and a lot of tooling have been updated, can we make the changes to the docs?

tlively:
Sure, I can update them today.

abrown:
I don't see how this updates the spec tests which would be pretty useful (fixing https://github.com/bytecodealliance/wasm-tools/blob/ede81acc5f594b2efb3cd3c62aad63c6d20da023/tests/roundtrip.rs#L652-L661 and https://github.com/bytecodealliance/wasmtime/blob/master/build.rs#L203). Is there some magic to get the opcodes out of the tables or the spec tests need to be changed as well?

alexcrichton:
As an example, [this test](https://github.com/WebAssembly/simd/blob/bbf08e4cda457395a630df1b2222d636deacbf9c/test/core/simd/simd_const.wast#L1566-L1580) still uses the old encoding of `v128.const` which was `0xfd 0x02` (same as a number of tests below that)

tlively:
Oh right, I only updated the documents. The spec tests are being developed and validated by @Honry upstream in WAVM and imported here, so I think it would make sense to update them in a separate PR by that same process. @Honry does that sound good to you?

dtig:
Sorry, missed the last update on this - I think #233 should fix it - I don't see other places that this is hardcoded in. @tlively The PR is on this branch, so maybe we can combine that with this PR, and follow up with @Honry if other tests need to be fixed as well, wdyt?

tlively:
@dtig sounds good. Want to push that commit to this branch?

dtig:
Yeah, I'd like to do that - mostly because a few tools pull spec tests directly from this repository so having it updated here hopefully reduces some of that confusion and a WAVM update can follow.

proposals/simd/BinarySIMD.md
```diff
@@ -28,177 +28,177 @@ The `v8x16.shuffle` instruction has 16 bytes after `simdop`.
...
+| `i32x4.load16x4_u`         |    `0x04`| m:memarg           |
+| `i64x2.load32x2_s`         |    `0x05`| m:memarg           |
+| `i64x2.load32x2_u`         |    `0x06`| m:memarg           |
+| `v8x16.load_splat`         |    `0x07`| m:memarg           |
+| `v16x8.load_splat`         |    `0x08`| m:memarg           |
+| `v32x4.load_splat`         |    `0x09`| m:memarg           |
+| `v64x2.load_splat`         |    `0x0a`| m:memarg           |
+| `v128.store`               |    `0x0b`| m:memarg           |
+| `v128.const`               |    `0x0c`| i:ImmByte[16]      |
+| `v8x16.swizzle`            |    `0x0d`| -                  |
```
dtig, line 43: It looks like the shuffle/swizzle opcodes have swapped opcode numbers.

tlively: Oops! Thanks for catching that :)

PR #215, opened 2020-04-29
Fix string comparison

ngzhian:
Classic newbie mistake of using != on strings. Plus I got the
conditional wrong - it should error if s is none of the valid simd
shapes.

Also have the error message a bit more verbose.


ngzhian:
Decided to revert the error message back, don't want to cause unnecessary churn to error messages since this test suite is pulled into other packages, and the increase in verbosity isn't of *too* much help anyway.

PR #217, opened 2020-05-01
Parse v128.const of nans in result_list of assert_return with a recursive result' type

ngzhian:
Add a new `result'` case, `SimdF32x4Result`, that contains 4 `result'`, representing values in the lane, which can be a `F32` literal, or a `Nan`. This makes `result'` a recursive type.

I had to move the parsing of v128_const rule out of const, to avoid a shift reduce conflict in the `result :` rule.


ngzhian:
Went with #218 instead.

PR #218, opened 2020-05-01
Parse v128.const of nans in result_list of assert_return by wrapping existing result type

ngzhian:
Wrap existing `LitResult` and `NanResult` in a `ValueResult`, and create another case `SimdF32x4Result` to represent a f32x4 result. This allows us to parse v128.const containing literals and/or nans without any recursive types.

This is in contrast to #217 which makes `result'` recursive, and so some functions in `run.ml` have to be changed to recursive.

I had to move the parsing of v128_const rule out of const, to avoid a shift reduce conflict in the `result :` rule.

### DIFF
--- a/interpreter/script/run.ml
+++ b/interpreter/script/run.ml
@@ -250,16 +250,23 @@ let string_of_nan = function
 
 let type_of_result r =
   match r with
-  | LitResult v -> Values.type_of v.it
-  | NanResult n -> Values.type_of n.it
+  | ValueResult (LitResult v) -> Values.type_of v.it
+  | ValueResult (NanResult n) -> Values.type_of n.it
+  | SimdF32x4Result (_, _, _, _) -> let open Types in V128Type
 
-let string_of_result r =
+let string_of_result_pat (r : result_pat) =
   match r with
-  | LitResult v -> Values.string_of_value v.it
-  | NanResult nanop ->
+  | (LitResult v) -> Values.string_of_value v.it
+  | (NanResult nanop) ->
     match nanop.it with
     | Values.I32 _ | Values.I64 _ | Values.V128 _ -> assert false
     | Values.F32 n | Values.F64 n -> string_of_nan n
+
+let string_of_result r =
+  match r with
+  | ValueResult v -> string_of_result_pat v
+  | SimdF32x4Result (s1, s2, s3, s4) ->
+    String.concat " " (List.map string_of_result_pat [s1; s2; s3; s4])
 
 let string_of_results = function
   | [r] -> string_of_result r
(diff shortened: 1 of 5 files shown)


ngzhian:
@rossberg This patch (and #217) implements parsing of v128.const containing nans, inside of `assert_return`. I found that there are two approaches
- add new type to wrap `LitResult` and `NanResult` (this patch)
- make `result'` recursive (#217)
I'm hoping to get your advice on which is a better solution.

In both cases I had to tweak the parser a bit, to move `v128_const` parsing out of the `const` rule, this is to avoid a shift/reduce conflict in the `result` rule.

interpreter/text/parser.mly
```diff
@@ -846,18 +860,28 @@ meta :
...
+  | v128const const_list { $1 :: $2 }
+
+numpat :
+  | literal { fun s -> simd_lane_lit s $1.it $1.at }
+  | NAN { fun s -> simd_lane_nan s $1 (ati 3) }
 
 result :
-  | const { LitResult $1 @@ at () }
-  | LPAR CONST NAN RPAR { NanResult (nanop $2 ($3 @@ ati 3)) @@ at () }
+  | LPAR V128_CONST SIMD_SHAPE numpat numpat numpat numpat RPAR {
```
rossberg, line 880: Nit: Move this case after NAN case.

interpreter/text/parser.mly
```diff
@@ -846,18 +860,28 @@ meta :
   | LPAR OUTPUT script_var_opt RPAR { Output ($3, None) @@ at () }
 
 const :
-  | LPAR CONST literal RPAR { snd (literal $2 $3) @@ ati 3 }
+  | LPAR CONST literal RPAR { print_endline "const"; snd (literal $2 $3) @@ ati 3 }
```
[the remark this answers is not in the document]
ngzhian: Yea, removed.

interpreter/script/script.ml
```diff
@@ -17,6 +17,9 @@ and nan = CanonicalNan | ArithmeticNan
 
 type result = result' Source.phrase
 and result' =
+  | ValueResult of result_pat
+  | SimdF32x4Result of result_pat * result_pat * result_pat * result_pat
+and result_pat =
   | LitResult of Ast.literal
   | NanResult of nanop
```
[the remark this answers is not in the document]
ngzhian: Latest commit makes this change.
I find that I have to do a nested pattern match on LitPat/NanPat in multiple places, since `Ast.literal = Values.value Source.phrase`, everytime I have a `num_pat`, I have to dig into `num_pat.it`, then match on that to get `LitPat`, then do `.it` again.
I wonder if we can have
```
and result' =
  | NumResult of num_pat'
```
Because for NumResult, the source of the pattern is the same.
It makes the type definition look a bit asymmetrical though.

interpreter/text/parser.mly
ngzhian: Done.

interpreter/script/run.ml
```diff
@@ -348,11 +348,11 @@ let run_action act : Values.value list =
...
 
 
-let assert_result_pat at v r =
+let assert_num_pat at v r =
   let open Values in
   match r with
-    | LitResult v' -> v <> v'.it
-    | NanResult nanop ->
+    | { it = LitPat v'; _ } -> v <> v'.it
+    | { it = NanPat nanop; _ } ->
```
rossberg, line 355: You can match on `r.it` to avoid these.

interpreter/text/arrange.ml
```diff
@@ -444,9 +444,9 @@ let nan = function
 
 let result res =
   match res.it with
-  | ValueResult (LitResult lit) -> literal lit
-  | SimdF32x4Result _ -> failwith "unimplemented"
-  | ValueResult (NanResult nanop) ->
+  | NumResult { it = LitPat lit; _ } -> literal lit
+  | SimdResult _ -> failwith "unimplemented"
+  | NumResult { it = NanPat nanop; _ } ->
```
rossberg, line 449: Here I would actually create a separate function for printing num_pat's.

interpreter/script/run.ml
```diff
@@ -250,23 +250,23 @@ let string_of_nan = function
...
+  | NumResult { it = NanPat v ; _ } -> Values.type_of v.it
+  | SimdResult (_, _) -> let open Types in V128Type
 
-let string_of_result_pat (r : result_pat) =
+let string_of_num_pat (r : num_pat) =
   match r with
-  | (LitResult v) -> Values.string_of_value v.it
-  | (NanResult nanop) ->
+  | { it = LitPat v; _ } -> Values.string_of_value v.it
+  | { it = NanPat nanop; _ } ->
```
rossberg, line 260: You can match on `r.it` here.

interpreter/script/run.ml
```diff
@@ -250,23 +250,23 @@ let string_of_nan = function
...
   match r with
-  | ValueResult (LitResult v) -> Values.type_of v.it
-  | ValueResult (NanResult n) -> Values.type_of n.it
-  | SimdF32x4Result (_, _, _, _) -> let open Types in V128Type
+  | NumResult { it = LitPat v ; _ } -> Values.type_of v.it
+  | NumResult { it = NanPat v ; _ } -> Values.type_of v.it
+  | SimdResult (_, _) -> let open Types in V128Type
 
-let string_of_result_pat (r : result_pat) =
+let string_of_num_pat (r : num_pat) =
```
rossberg, line 257: Nit: rename `r` to `p`

interpreter/script/run.ml
```diff
@@ -348,11 +348,11 @@ let run_action act : Values.value list =
     )
 
 
-let assert_result_pat at v r =
+let assert_num_pat at v r =
```
rossberg, line 351: Nit: rename `r` to `p`

interpreter/script/script.ml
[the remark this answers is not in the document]
rossberg: I think the match on `it` is fine. In some cases you can actually avoid it, see my comments. In some others it might be necessary to factor out handling num_pat into a separate function anyway once you implement the SIMD case.

interpreter/script/run.ml
ngzhian: Done

interpreter/script/run.ml
ngzhian: Done

interpreter/script/run.ml
ngzhian: Good point, fixed!

interpreter/text/arrange.ml
ngzhian: Done.

interpreter/script/run.ml
ngzhian: Done.

ngzhian:
Thanks for the review, Andreas. Merging.

PR #219, opened 2020-05-04
Implement v128.const parsing for f64x2 with nans

ngzhian:
And enough operations to pass the simd_f64x2.wast test.

### DIFF
--- a/interpreter/syntax/ast.ml
+++ b/interpreter/syntax/ast.ml
@@ -47,8 +47,8 @@ end
 (* FIXME *)
 module SimdOp =
 struct
-  type unop = F32x4Abs
-  type binop = F32x4Min | F32x4Max
+  type unop = F32x4Abs | F64x2Abs
+  type binop = F32x4Min | F32x4Max | F64x2Min | F64x2Max
   type testop = TodoTestOp
   type relop = TodoRelOp
   type cvtop = TodoCvtOp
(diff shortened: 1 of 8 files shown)


interpreter/syntax/ast.ml
```diff
@@ -47,8 +47,8 @@ end
 (* FIXME *)
 module SimdOp =
 struct
-  type unop = F32x4Abs
-  type binop = F32x4Min | F32x4Max
+  type unop = F32x4Abs | F64x2Abs
+  type binop = F32x4Min | F32x4Max | F64x2Min | F64x2Max
```
rossberg, line 51: Given that most of the regular SIMD operators are symmetrically available for each lane count, wouldn't it be nicer to factor this? For example:
```
type iunop = ...
type ibinop = ...
type funop = Abs | ...
type fbinop = Min | Max | ...
...
type unop = Simd.shape * (iunop, iunop, funop, funop) op
type binop = Simd.shape * (ibinop, ibinop, fbinop, fbinop, unit) op
...
```
Or something like this?

(Similarly, instead of a flat list, the `Simd` module could have symmetric submodules for the ops on different shapes, which all share the same signature.)

ngzhian: Agree, I can do that in a follow up. I can't really imagine what the "symmetric submodules" look like exactly yet, I'll probably need to write something up locally to see how it all fits in. Are you thinking of something like:

```
module type S =
sig
  val min
  val max
  ...
end

module F32x4 : S = struct
  let min = ?
...
end

(* repeat for each SIMD shape *)
```

rossberg: Yes, exactly. Sorry for being terse.

PR #222, opened 2020-05-06
WIP V128 refactor simd submods

ngzhian:


interpreter/exec/simd.ml
```diff
@@ -72,26 +79,49 @@ struct
...
-    of_f64x2 (List.map f (to_f64x2 x))
-  let f64x2_binop f x y =
-    of_f64x2 (List.map2 f (to_f64x2 x) (to_f64x2 y))
 
   let f64x2_extract_lane i x = List.nth (to_f64x2 x) i
-  let f64x2_min = f64x2_binop F64.min
-  let f64x2_max = f64x2_binop F64.max
-  let f64x2_abs x = f64x2_unop F64.abs x
+
+  module type Shape =
```
ngzhian, line 89: @rossberg does this look like something you had in mind?
I had to duplicate this `Shape` sig, not sure how else to do it.
And the other problem I had with this nested module is that. `Shape.u` is now opaque, so I couldn't really implement extract_lane
```
module F32x4 : Shape = struct
type u = F32.t
val extract_lane (i : int) (x : t) : u = implementation
```
Then callers will get a `Simd.F32x4.u` instead of a `F32.t`.

ngzhian: Found a way around this in the latest commit.

interpreter/exec/simd.ml
```diff
@@ -45,15 +57,35 @@ sig
...
+(* The base type of a SIMD shape, e.g. for F32x4, the ElementType is F32 *)
+module type ElementType =
+sig
+  type t
+  val abs : t -> t
+  val min : t -> t -> t
+  val max : t -> t -> t
+end
+
+module MakeSimdShape (B : ElementType) (Rep : sig
```
ngzhian, line 75: I guess I could move this into `module Make` below, then I don't need to have the `Rep` parameter.

interpreter/exec/simd.ml
```diff
@@ -33,6 +33,18 @@ sig
   val of_f64x2 : F64.t list -> t
 end
 
+(* This signature defines the types and operations different SIMD shapes can expose. *)
+module type SimdShape =
```
rossberg, line 37: I'd rename this to `Float`, since there will be a different one for ints.

interpreter/exec/simd.ml
```diff
@@ -33,6 +33,18 @@ sig
...
+sig
+  type t
+  type u (* underlying type *)
```
rossberg, line 40: Rename to `lane`?

interpreter/exec/simd.ml
rossberg: Makes sense. Same comment about renaming this to `MakeFloat`.

interpreter/exec/simd.ml
```diff
@@ -78,50 +103,18 @@ struct
...
-    val min : t -> t -> t
-    val max : t -> t -> t
-  end
-
-  module F32x4 : Shape =
-  struct
-      type u = F32.t
-
+  module F32x4 : SimdShape with type t = t and type u = F32.t =
+    MakeSimdShape (F32) (struct
```
rossberg, line 107: No need to repeat the signature here.
```suggestion
  module F32x4 = MakeSimdShape (F32) (struct
```

interpreter/exec/simd.ml
```diff
@@ -78,50 +103,18 @@ struct
...
-      let max = binop F32.max
-  end
-
-  module F64x2 : Shape =
-  struct
-      type u = F64.t
+    end)
 
+  module F64x2 : SimdShape with type t = t and type u = F64.t =
+    MakeSimdShape (F64) (struct
```
rossberg, line 114:
```suggestion
  module F64x2 = MakeSimdShape (F64) (struct
```

interpreter/exec/simd.ml
ngzhian: I also changed ElementType to be Float.S.

PR #223, opened 2020-05-06
Update ImplementationStatus to reflect status quo

zeux:
Updates based on latest LLVM & latest v8 (earlier v8 versions aren't
useful because of renumbering anyway):

- v8 implements load_splat and load with zero/sign extension
- v8 implements i64x2_mul
- v8 implements swizzle and andnot
- LLVM supports integer abs instructions and intrinsics
- Removing note about shuffle opcode - irrelevant post-renumbering

Notable remaining omissions:

- v8 doesn't implement v128.const
- LLVM doesn't implement (or at least doesn't expose an intrinsic for)
i64x2_mul
- A lot of instructions are still part of unimplemented subset in LLVM


ngzhian:
V8 side LGTM, thank you!

binji:
@lars-t-hansen It'd be great to get the spidermonkey status too!

lars-t-hansen:
> @lars-t-hansen It'd be great to get the spidermonkey status too!

In due time :-)  We've landed in baseline (x64 only) and preffed-off, but once we land in the optimizing compiler I'll post a PR for implementation status.  Hoping this will happen in May.

PR #231, opened 2020-05-18
Create a functor MakeInt (similar to MakeFloat)

ngzhian:
This does little right now, but will make it easier when we add more
SIMD int ops later (similar to MakeFloat).


ngzhian:
@rossberg ping

ngzhian:
Hi @rossberg ptal when you get a chance, thanks!!

Issue #203, opened 2020-03-09
Soft Freeze on new operation additions post Phase 3


dtig:
As the SIMD proposal is now at [Phase 3](https://github.com/WebAssembly/meetings/blob/master/process/phases.md#3-implementation-phase-community--working-group), and getting closer to standardization, I'd like to propose a soft freeze on operation additions so that implementations, and toolchains have the opportunity to  implement and optimize the existing operations without too much churn. 

The freeze does not include existing issues/PRs, and is only for new proposals. The rationale is that due to the nature of the SIMD proposal, it's possible to keep adding new operations and given that there is reasonable consensus (barring open issues) on the current proposal, it will be hard to progress in standardizing the proposal without some sort of a freeze. 

This is a soft freeze as it is possible with more implementations getting up to speed, and more applications are using the current SIMD proposal that there are must-have operations, or semantics need to be tweaked. But the bar for these would be higher. 

Some open questions -

1. Is a deadline for a soft freeze useful (as opposed to going in effect immediately)? I suspect not, as the proposal has already been under discussion for some time now and as it's not a hard freeze deadlines may be arbitrary and not useful. 
2. What should the bar be? This seems hard to codify in very concrete terms, but my thoughts are that the bar here should be that this would be very hard to emulate without an addition to the Spec, and the alternatives if any will cause performance cliffs. Addition of operations would be discussed on a case-by-case basis. 

I'll be pinning this issue as suggested in a previous meeting for visibility. Any thoughts/concerns about this approach?

dtig:
This would also make it easier for having a consistent set of opcode numbers as proposed in #129 going forward.

jan-wassenberg:
@dtig makes sense. I see no pressing need for a deadline/extension. Hard/expensive to emulate without spec change sounds like a reasonable bar to me.

zeux:
I would propose something like:

- hard/expensive to emulate in absence of spec change
- well supported on 2+ major archs
- useful in 2+ important use cases

Obviously all of this is vague (what are important use cases? what's the threshold for emulation complexity? etc.) but I think the goal is to think very carefully about any new instructions and default to maintaining the current proposal since solidifying it is (arguably) more important than adding a new instruction. Agreed that the deadline doesn't seem necessary.

tlively:
I like @zeux's idea for the threshold because it shares its axes of criteria with the current threshold and is simply stricter.

dtig:
[...]

PR #237, opened 2020-06-02
v128.load32_zero and v128.load64_zero instructions

Maratyszcza:
Introduction
=========

This PR introduce two new variants of load instructions which load a **single** 32-bit or 64-bit element into the lowest part of 128-bit SIMD vector and zero-extend it to full 128 bits. These instructions natively map to SSE2 and ARM64 NEON instruction, and have two broad use-cases:
1. Non-contiguous loads, when we need to combine elements from disjoint locations in a single SIMD vector. Non-contiguous loads are commonly emulated by doing loads a single elements and then combining the values through shuffles. While is it possible to do through a combination of scalar loads and `v128.replace_lane` instructions, the resulting code would use be inefficient in using too many general-purpose registers, producing an overly long dependency chain (every `v128.replace_lane` depends on the previous one), and hitting the long-latency/low-throughput instructions to copy from general-purpose registers to SIMD registers. Non-contiguous loads using the proposed `v128.load32_zero` and `v128.load64_zero` instructions avoid all these bottlenecks.
2. Processing fewer than 128 bits of data. Sometimes the algorithm or data structures just don't expose enough data to utilize all 128 bits of a SIMD vector, but would nevertheless benefit from processing fewer elements in parallel (e.g. adding 8 bytes in one SIMD instruction rather than eight scalar instructions).

Applications
=========

- [libgav1 (AV1 decoder)](https://chromium.googlesource.com/codecs/libgav1/+/df75e6a9e8e2032c9291d9f0aa2c7c8519adfbfe/src/dsp/x86/common_sse4.h#103)
- [x264 encoder: sum of transformed differences (processing less than 16 elements)](https://github.com/mirror/x264/blob/33f9e1474613f59392be5ab6a7e7abf60fa63622/common/x86/pixel-32.asm#L86-L89)
- [x264 encoder: deblocking (non-contiguous load)](https://github.com/mirror/x264/blob/33f9e1474613f59392be5ab6a7e7abf60fa63622/common/x86/deblock-a.asm#L157-L164)
- [QDPXX/libintrin: matrix-vector multiplication (non-contiguous load)](https://github.com/usqcd-software/libintrin/blob/6e429e4c32b5099546dce833294781255b4c54a7/lib/sse_mult_adj_su3_mat_4vec.c)
- [QNNPACK: depthwise convolution (processing less than 8 elements)](https://github.com/pytorch/QNNPACK/blob/901e9d40aeedb2c1b212b28f022fb099ffe617c2/src/q8dwconv/mp8x25-sse2.c#L49)
- [cpp Wilson Dslash (non-contiguous load)](https://github.com/JeffersonLab/cpp_wilson_dslash/blob/56a4abf64c4d586a73bc2fd7747067f6e92329c9/include/cpp_dslash_parscalar_mvv_recons_32bit_sse2.h#L45-L53)
- [GROMACS: matrix transpose (non-contiguous load)](https://github.com/gromacs/gromacs/blob/a68cc20fd27a3cbebbc7c7cf563fcfb558f1feca/src/gromacs/simd/impl_x86_sse2/impl_x86_sse2_util_double.h#L101-L102)
- [OpenBLAS: matrix-matrix multiplication](https://github.com/xianyi/OpenBLAS/blob/ce3651516f12079f3ca2418aa85b9ad571c3a391/kernel/x86_64/cgemm_kernel_4x8_sandy.S#L158-L159)

===========================

- **v128.load32_zero**
  - `v = v128.load32_zero(mem)` is lowered to `VMOVSS xmm_v, [mem]`
- **v128.load64_zero**
  - `v = v128.load64_zero(mem)` is lowered to `VMOVSD xmm_v, [mem]`

- **v128.load32_zero**
  - `v = v128.load32_zero(mem)` is lowered to `MOVSS xmm_v, [mem]`
- **v128.load64_zero**
  - `v = v128.load64_zero(mem)` is lowered to `MOVSD xmm_v, [mem]`

ARM64 processors

- **v128.load32_zero**
  - `v = v128.load32_zero(mem)` is lowered to `LDR Sv, [mem]`
- **v128.load64_zero**
  - `v = v128.load64_zero(mem)` is lowered to `LDR Dv, [mem]`

- **v128.load32_zero**
  - `v = v128.load32_zero(mem)` is lowered to `VMOV.I32 Qv, 0` + `VLD1.32 {Dv_lo[0]}, [mem]`
- **v128.load64_zero**
  - `v = v128.load64_zero(mem)` is lowered to `VMOV.I32 Dv_hi, 0` + `VLD1.32 {Dv_lo}, [mem]`

### DIFF
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -222,3 +222,5 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
 | `i32x4.trunc_sat_f32x4_u`  |    `0xf9`| -                  |
 | `f32x4.convert_i32x4_s`    |    `0xfa`| -                  |
 | `f32x4.convert_i32x4_u`    |    `0xfb`| -                  |
+| `v128.load32_zero`         |    `0xfc`| -                  |
+| `v128.load64_zero`         |    `0xfd`| -                  |
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -190,6 +190,8 @@
 | `i32x4.trunc_sat_f32x4_u`  |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `f32x4.convert_i32x4_s`    |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `f32x4.convert_i32x4_u`    |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
+| `v128.load32_zero`         |                           |                    |                    |                    |                    |
+| `v128.load64_zero`         |                           |                    |                    |                    |                    |
 
 [1] Tip of tree LLVM as of May 20, 2020
 
--- a/proposals/simd/NewOpcodes.md
+++ b/proposals/simd/NewOpcodes.md
@@ -12,6 +12,8 @@
 | v128.load32_splat  | 0x09   |
 | v128.load64_splat  | 0x0a   |
 | v128.store         | 0x0b   |
+| v128.load32_zero   | 0xfc   |
+| v128.load64_zero   | 0xfd   |
 
 | Basic operation | opcode |
 | ----------------| ------ |
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -766,6 +766,24 @@ def S.load(memarg):
     return S.from_bytes(memory[memarg.offset:memarg.offset + 16])
 ```
 
+### Load and Zero-Pad
+
+* `v128.load32_zero(memarg) -> v128`
+* `v128.load64_zero(memarg) -> v128`
+
+Load a single 32-bit or 64-bit element into the lowest bits of a `v128` vector,
+and initialize all other bits of the `v128` vector to zero.
+
+```python
+def S.load32_zero(memarg):
+    return S.from_bytes(memory[memarg.offset:memarg.offset + 4])
+```
+
+```python
+def S.load64_zero(memarg):
+    return S.from_bytes(memory[memarg.offset:memarg.offset + 8])
+```
+
 ### Load and Splat
 
 * `v128.load8_splat(memarg) -> v128`


tlively:
Thanks for the suggestion, @Maratyszcza! Now that we are in phase 3, we have [stricter guidelines](https://github.com/WebAssembly/simd/issues/203#issuecomment-597440438) on adding new instructions. It sounds like these instructions are well supported on multiple architectures, but we need to agree that they are used in multiple important use cases and that they would be expensive to emulate. Can you point to real-world uses of this pattern that we could adapt as benchmarks to determine how much of a benefit these instructions would be?

Maratyszcza:
@tlively Added examples of applications using these instructions

Maratyszcza:
XNNPACK has SIMD table-based exp and sigmoid implementations that could be used for evaluation

jan-wassenberg:
@tlively I agree these would be helpful. Another expensive to emulate use case is when you have an existing data structure of 1-2 floats and we can't be sure 4 floats are accessible. Or the much more common case of remainder handling - using the same code as the main loop, but with 32-bit loads/stores going one element at a time. JPEG XL has [several examples](https://gitlab.com/search?utf8=%E2%9C%93&search=HWY_CAPPED%28float%2C+1%29&group_id=2658731&project_id=10573324&scope=&search_code=true&snippets=false&repository_ref=master&nav_source=navbar) of this.

lemaitre:
Maybe the more general `vld1q_lane` from Neon might be desirable (https://static.docs.arm.com/den0018/a/DEN0018A_neon_programmers_guide_en.pdf#G15.1154120)

Basically, it can loads a single element into any lane, not just the first one, and leaves the other lanes untouched.

The problem I see is that it would need some kind pattern matching to make this the most efficient on x86 where we only have "load into the first lane and put the rest at zero".
But we can envision that a sequence of `load_lane` could be converted into shuffles in SSE and even a (masked) gather in AVX2.

If pattern recognition fails (or is disabled), the generated code for a single `load_lane` would still be faster than `scalar load + insert_lane` as it would be converted into  `loadl + shuffle` and stay in the SIMD register space.
And the shuffle can be easily eliminated if the WASM runtime detect that the lane index is 0 and the input vector already contains zeros.

---
The store counterpart might also be interesting.
However, the store variant will not be able to handle 8- and 16-bit types efficiently on x86.
We can stay with 32- and 64-bit types, as proposed here, though.

tlively:
For consistency with the load_splat instructions, these instructions should probably have `v32x4` and `v64x2` prefixes. More descriptive names might be `v32x4.load_lane` and `v64x2.load_lane`.

Maratyszcza:
IMO it is best to save `load_lane` for (future) variants which load a single lane while leaving other unchanged (i.e. analogs of `vld1q_lane_XX` in ARM NEON and `_mm_insert_epiXX` on x86).

We could rename these instructions to `v128.load32_u` and `v128.load64_u` for consistency with `v128.load32x2_u` and other zero-extending instructions.

tlively:
Saving `load_lane` for potential future instructions makes sense to me. How about `v32x4.load32` and `v64x2.load64`? The `_u` suffix doesn't seem necessary because there is no sign interpretation happening. I still think it makes sense to use the prefixes for hinting at the lane interpretation, but I could probably be convinced otherwise as well.

On a different note, prototypes of these instructions have been merged to both LLVM and Binaryen and will be available in the next version of Emscripten via the builtin functions `__builtin_wasm_load32_zero` and `__builtin_wasm_load64_zero`.

ngzhian:
I think the memory instructions should all start with `v128.` (Ref: mvp instructions are all of the form `<type>.load[<n>_<sx>]`.)

The shape prefix suggests how the operands are treated, which doesn't apply for loads, since the operands are all memargs. This *might* be a point of confusion. Making everything start with `v128.load_` will help categorize all these variants of load as: "load from memory to get a v128", i.e. these are all the ways you can load something from memory to get a v128. Then the format becomes:

`v128.load_<splat/extend/zero/others>_<numberofbytesloaded>_<sign extension>"`

For load_splat we might even consider change it to: `v128.load_splat8`, similar to how we have `i32.load8_s`.

So maybe load zeroes can be: `v128.load_zero32`.

I think this has a (imo nice) side effect of making the spec text a bit clearer, because you can now say, all instructions that start with the shape prefix describe how they treat their operands (and you don't have to say "except for memory instructions").

tlively:
@ngzhian That seems reasonable and consistent to me. We would want to use the `v128` prefix for load-extend operations as well. I think it would look more consistent with MVP if we put `<numberofbytesloaded>` after `load`, like in `v128.load8_splat` or `v128.load32_zero`. WDYT?

ngzhian:
Yea that looks good to me. It becomes really clear from the name that v128 is the return type, and how many bytes will be loaded. The remaining portion will be to tell us how to get from bytes to v128, of which we can have many different wants.

tlively:
@Maratyszcza we briefly discussed this in the sync meeting today, and there is general support for these instructions, but we still need benchmarking data to make the case for including them. Would you be able to get performance numbers for these?

Maratyszcza:
I evaluated the performance effect of these instructions by modifying WebAssembly SIMD microkernels for Sigmoid operator in [XNNPACK library](https://github.com/google/XNNPACK) of neural network operators to use `v128.load32_zero` instruction, and the results are below:

Operator | Performance with WAsm SIMD | Performance with WAsm SIMD + `v128.load32_zero` | Speedup |
-- | -- | -- | --
Sigmoid | 5.09 GB/s | 5.82 GB/s | 14% |

These speedups were measured on ThinkStation P520, and the code modifications can be seen in google/XNNPACK#1016

ngzhian:
@billbudge brought up a good point on the name of these instructions, can we rename this to: `v128.load32_u` and `v128.load64_u`? It is essentially loading 32 bits, then unsigned extending to v128. This is symmetrical with the load instructions in the spec.

(Though there is a bit of an asymmetry since we don't have the signed versions of these, but it doesn't make sense in the context of SIMD.)

tlively:
Fine with me!

Maratyszcza:
Additional Sigmoid benchmarks on ARM64 phones:

Phone | Performance with WAsm SIMD | Performance with WAsm SIMD + `v128.load32_zero` | Speedup |
-- | -- | -- | --
LG G8 ThinQ | 3.47 GB/s | 3.60 GB/s | 4% |
Galaxy S10 (Exynos) | 3.82 GB/s | 3.96 GB/s | 4% |

omnisip:
Why doesn't this instruction set have a non-memory load?  E.g. to load into the lower part without pulling from memory?

tlively:
@omnisip that sounds like v128.const, but leaving some lanes undefined. Unfortunately we can't have instructions that leave parts of values undefined in WebAssembly because we can't have undefined behavior and we don't want to make the type system so complex that it can track undefined lanes statically.

omnisip:
In this case, it would be the same operation with a register instead of a
memory argument. There's no undefined behavior.

tlively:
How is that different from `v128.const` then?

omnisip:
The value is calculated at runtime but isn't a pointer in memory.

tlively:
Then it sounds like a `i64x2.replace_lane 0` I guess that requires a full vector to be materialized first, though.

In general, the answer to "why doesn't the instruction set have X" is some combination of "X is not portable enough," "X is not useful enough,"  or "no one has suggested X yet". You're totally welcome to suggest new instructions when you identify deficiencies in the instruction set. Useful information to include is how the instruction would be lowered on Intel and ARM ISAs, applications that would benefit from the instruction, and any estimates you have for the performance improvement that instruction can bring. See many of @Maratyszcza's PRs for good examples of new instruction proposals.

omnisip:
In this case, it would belong with this PR because it would still be the same underlying instructions with a different argument. @Maratyszcza do you want me to make a patch or do you want to add this yourself?

tlively:
WebAssembly doesn't overload instructions with different kinds of arguments like that, so it will have to be a new instruction proposal.

omnisip:
There are three identical implementations for this functionality that are
useful:

1) from memory like this currently is
2) from 32 or 64 bit register
3) from another v128 vector

I'll put the other two in my proposal so they can be evaluated together
when nomenclature is discussed.

penzn:
> Why doesn't this instruction set have a non-memory load? E.g. to load into the lower part without pulling from memory?

Memory operations either go from memory to a virtual register (load) or from virtual register to memory (store). There are no other possible semantics for memory operations in WebAssembly.

omnisip:
@penzn If I understand your comments correctly, this doesn't affect my proposal in #373, correct?

ngzhian:
In the sync today we agreed to add these 2 instructions.
@Maratyszcza can you please update https://github.com/WebAssembly/simd/blob/master/proposals/simd/NewOpcodes.md as well? Put it in the table of memory instructions.

Maratyszcza:
@ngzhian Done

ngzhian:
Thanks, LGTM

PR #238, opened 2020-06-02
Implement i32x4 neg add sub mul

ngzhian:
This required exposing neg from int.ml, to be used by simd.ml

Fixes to parsing of v128.const in assert_returns, to handle i32x4.

Handle matching i32x4 in script/run.ml.

With this, test/core/simd/simd_i32x4_arith.wast passes.

### DIFF
--- a/interpreter/script/run.ml
+++ b/interpreter/script/run.ml
@@ -375,6 +375,14 @@ let assert_result at got expect =
             let open Values in
             let open Simd in
             match shape, v with
+            | I32x4, V128 v ->
+                    let l0 = I32 (V128.I32x4.extract_lane 0 v) in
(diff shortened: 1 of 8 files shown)


interpreter/script/run.ml
```diff
@@ -375,6 +375,14 @@ let assert_result at got expect =
             let open Values in
             let open Simd in
             match shape, v with
+            | I32x4, V128 v ->
+                    let l0 = I32 (V128.I32x4.extract_lane 0 v) in
```
rossberg, line 379: Nit: odd indentation?

ngzhian: Yea, might as well fix the indentation below too!

PR #239, opened 2020-06-02
Fix: assert_return should use canonical NaN for comparison

abrown:
In running these tests, I discovered that they expected a specific NaN, not a canonical NaN, which causes errors if we always return a canonical NaN of a different sign. As @ngzhian pointed out in a separate conversation, the spec expects the returned value to be a canonical NaN: https://webassembly.github.io/spec/core/bikeshed/index.html#nan-propagation%E2%91%A0. 

@Honry, I don't know if any meta machinery needs to change to generate these files correctly but let me know and I can update the PR.

cc: @arunetm

### DIFF
--- a/test/core/simd/meta/simd_f64x2.py
+++ b/test/core/simd/meta/simd_f64x2.py
@@ -144,13 +144,13 @@ class Simdf64x2Case(Simdf32x4Case):
             [
                 'f64x2.min',
                 [['nan', '0'], ['0', '1']],
-                [['nan', '0']],
+                [['nan:canonical', '0']],
                 ['f64x2', 'f64x2', 'f64x2']
             ],
             [
                 'f64x2.min',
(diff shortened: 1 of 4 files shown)


ngzhian:
Thanks for making these changes! LGTM, I'm also not sure if these were generated from somewhere, if so, that should change too.

Honry:
@abrown, good catch! Please help update these files: https://github.com/WebAssembly/simd/blob/master/test/core/simd/meta/simd_f64x2.py#L147,
https://github.com/WebAssembly/simd/blob/master/test/core/simd/meta/simd_f64x2.py#L153,
https://github.com/WebAssembly/simd/blob/master/test/core/simd/meta/simd_f64x2_arith.py#L121

abrown:
Ok, I updated the meta code as well. @arunetm can we approve and merge?

arunetm:
LGTM

PR #240, opened 2020-06-03
Implement i32x4 abs min_s min_u max_s max_u

ngzhian:


interpreter/text/lexer.mll
```diff
@@ -400,7 +412,7 @@ rule token = parse
       BINARY (simdop s unreachable unreachable unreachable unreachable f32x4_max f64x2_max) }
   | (simd_shape as s)".abs"
     { if s = "i64x2" then error lexbuf "unknown operator";
```
rossberg, line 414: Is this condition out of date?

Issue #162, opened 2019-12-19
Alignment of `loadAxB` and `load_splat`


abrown:
We have been going back and forth about this in e-mail chains and [PRs](https://github.com/bytecodealliance/wasmparser/pull/158#discussion_r358455367): what are the allowed values for the WAT `align` modifier when using the SIMD `loadAxB` (load extend) and `load_splat` instructions?

The [SIMD document](https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#integer-to-integer-narrowing) doesn't clearly specify the valid values and we should remedy this. The [current spec does](https://webassembly.github.io/spec/core/bikeshed/index.html#-tmathsfhrefsyntax-instr-memorymathsfloadnmathsf_hrefsyntax-sxmathitsxhrefsyntax-memargmathitmemarg) for existing instructions and reading `loadN` would lead me to assume the following for `loadAxB`:
 - we ignore the destination size (128 bits) and focus exclusively on the source size of the data--the value to be loaded
 - we need to get the byte-width of the value to be loaded (e.g. for `Load8x8`: `8x8/8 = 64/8 = 8`)
 - we need to satisfy `2^memarg.align <= [the byte width]` (e.g. for `Load8x8`: `memarg.align` is in `[0, 1, 2, 3]`)
 - this means that in WAT, `align` can have the values of `2^memarg.align` (e.g. for `Load8x8`: `2^0 = 1`, `2^1 = 2`, `2^2 = 4`, and `2^3 = 8`)

Is this understanding correct? I think it intuitively makes sense to have alignment of 1, 2, 4, and 8 bytes for 64-bit values (e.g. `Load8x8`) because those are the log_2 ways of aligning the bytes; similarly, 128-bit values would have alignments of 1, 2, 4, 8, and 16 bytes. But I want to make this more explicit in the spec or correct my understanding before I go much further. I would also like to confirm my assumption that the Wasm binary encoding will use `memarg.align` and that the Wasm textual encoding will use `2^memarg.align`.

penzn:
This is a very good point. Normally alignment should follow the source size - as it is supposed to match the store that originally produced the values - but there is no "partial size" SIMD store to match it. It would still make sense to follow the source size though, even if store is a full SIMD store aligned on 16 bytes, taking parts of the value would be aligned on respective size (4 byte lane would be aligned on 4 bytes, 8 byte chunk for load-and-extend would be aligned on 8 bytes and so on).

[...]

PR #244, opened 2020-06-05
Specify the alignment of splatting and extending loads

### LINKED ISSUE #230: Load and extend natural alignment
LLVM currently emits load and extend instructions with alignment parameters up to 3 (i.e. 8-byte aligned), but V8 is throwing errors saying the expected maximum alignment for `i16x8.load8x8_s` is 0 (i.e. 1-byte alignment). The overview document does not specify either the natural alignment or the maximum, so we should figure out what they should be.

IMO it makes sense for 8-byte alignment to be the natural alignment because these instructions are loading 8 bytes at a time.

tlively:
Clarify that the natural alignment is always the size of the loaded
data.

Resolves #230 and resolves and #162.

### DIFF
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -773,7 +773,8 @@ def S.load(memarg):
 * `v32x4.load_splat(memarg) -> v128`
 * `v64x2.load_splat(memarg) -> v128`
 
-Load a single element and splat to all lanes of a `v128` vector.
+Load a single element and splat to all lanes of a `v128` vector. The natural
+alignment is the size of the element loaded.
 
 ```python
 def S.load_splat(memarg):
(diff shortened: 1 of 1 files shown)


proposals/simd/SIMD.md
```diff
@@ -790,7 +791,8 @@ def S.load_splat(memarg):
 * `i64x2.load32x2_s(memarg) -> v128`: load two 32-bit integers and sign extend each one to a 64-bit lane
 * `i64x2.load32x2_u(memarg) -> v128`: load two 32-bit integers and zero extend each one to a 64-bit lane
 
-Fetch consecutive integers up to 32-bit wide and produce a vector with lanes up to 64 bits.
+Fetch consecutive integers up to 32-bit wide and produce a vector with lanes up
+to 64 bits. The natural alignment is 8.
```
ngzhian, line 795: I think we should specify "bytes" here, the memarg in spec uses log2, so it can be a bit confusing.

PR #243, opened 2020-06-04
Add f32x4 add sub mul div neg sqrt

ngzhian:
This makes simd_f32x4_arith.wast pass.

PR #245, opened 2020-06-05
Fix lexer

ngzhian:
In #243 I accepted some changes without pulling and running locally. My
laziness is rewarded with compilation errors. This fixes the issue.


ngzhian:
Going to be a rebel and merge without Andreas's +1 :)

rossberg:
:)

PR #246, opened 2020-06-05
Add f64x2 add sub mul div neg sqrt

ngzhian:
This passes f64x2_arith.wast


interpreter/text/lexer.mll
```diff
@@ -389,17 +389,15 @@ rule token = parse
   | "output" { OUTPUT }
 
   | (simd_shape as s)".neg"
-    { if s <> "i32x4" && s <> "f32x4" then error lexbuf "unknown operator";
-      UNARY (simdop s unreachable unreachable i32x4_neg unreachable f32x4_neg unreachable) }
-  | (simd_float_shape as s)".sqrt"
-    { if s <> "f32x4" then error lexbuf "unknown operator";
-      UNARY (simd_float_op s f32x4_sqrt unreachable) }
+    { if s <> "i32x4" && s <> "f32x4" && s <> "f64x2" then error lexbuf "unknown operator";
```
rossberg, line 392: These checks are getting very repetitive. How about introducing a little helper, like:
```
let only shapes s lexbuf =
  if not (List.mem s shapes) then
    error lexbuf "instruction not yet implemented";
```
Then e.g. the current line could become
```
only ["i32x4"; "f32x4"; "f64x2"] s lexbuf;
```

ngzhian: Make sense, I will make that change in a follow up.

PR #249, opened 2020-06-08
Create a helper to check valid simd operations

ngzhian:
The "i32x4" case looks trivial now, but we will be adding i64x2 later.


ngzhian:
Addressing feedback on https://github.com/WebAssembly/simd/pull/246/files/73b22cfe4887b24bbe7635d2c76d597828f52db0#r436483845

PR #250, opened 2020-06-08
Refactor simd operations out of eval_numeric into eval_simd

ngzhian:
@rossberg does this look like what you had in mind when you suggested to pull out SIMD related functionality out of eval_numeric into its own eval_simd?
I couldn't figure out how to deal with `exception TypeError`, I couldn't import it from eval_numeric, since that would be circular imports.
Should I define `exception TypeError` in another file, then import it into both eval_numeric and eval_simd?


ngzhian:
Np, since i'm on an instructions-adding mode right now, I'll leave this as it is, and when things are more stable I will pick up this refactoring again (will have less merge conflicts then). Thanks!

PR #254, opened 2020-06-11
Consolidate parsing of v128.const

ngzhian:


ngzhian:
@rossberg ptal, thanks!

PR #258, opened 2020-06-17
Support i16x8 and implement neg, add, sub, mul

ngzhian:
Create a new module I16, which uses Int.Make, and is backed by Int32. It
reuses a bunch of logic in Int. It stores 16-bit integers sign-extended
in Int32. This means that -1 (0xFFFF) is stored as 0xFFFFFFFF, rather
than 0x0000FFFF. All the bytes decode/encode logic is also done using
the signed form.

### DIFF
--- a/interpreter/exec/int.ml
+++ b/interpreter/exec/int.ml
@@ -237,11 +237,9 @@ struct
     | 'A' .. 'F' as c ->  0xa + Char.code c - Char.code 'A'
     | _ ->  failwith "of_string"
 
-  let max_upper, max_lower =
-    let mup, mlow = divrem_u Rep.minus_one ten in
-    mup, mlow
+  let max_upper, max_lower = divrem_u Rep.minus_one ten
 
-  let maybe_sign_extend i =
+  let sign_extend i =
     (* This module is used with I32 and I64, but the bitwidth can be less
      * than that, e.g. for I16. When used for smaller integers, the stored value
      * needs to be signed extened, e.g. parsing -1 into a I16 (backed by Int32)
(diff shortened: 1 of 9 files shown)


interpreter/exec/int.ml
```diff
@@ -237,7 +237,26 @@ struct
     | 'A' .. 'F' as c ->  0xa + Char.code c - Char.code 'A'
     | _ ->  failwith "of_string"
 
-  let max_upper, max_lower = divrem_u Rep.minus_one ten
+  let max_upper, max_lower =
+    let mup, mlow = divrem_u Rep.minus_one ten in
+    mup, mlow
```
rossberg, line 242: Is this change some debugging leftover?

interpreter/exec/int.ml
```diff
@@ -237,7 +237,26 @@ struct
...
+
+  let maybe_sign_extend i =
```
rossberg, line 244: Nit: I'd name this just `sign_extend`

interpreter/exec/int.ml
```diff
@@ -237,7 +237,26 @@ struct
...
+    (* This module is used with I32 and I64, but the bitwidth can be less
+     * than that, e.g. for I16. When used for smaller integers, the stored value
+     * needs to be signed extened, e.g. parsing -1 into a I16 (backed by Int32)
+     * shoud have all high bits set. We can do that by logor with a mask,
+     * where the mask is minus_one left shifted by bitwidth. But if bitwidth
+     * matches the number of bits of Rep, the shift will be incorrect.
+     *   -1 (Int32) << 32 = -1
+     * Then the logor will be also wrong. So we check and bail out early.
+     * *)
+    if Rep.bitwidth >= 32 then i else
```
rossberg, line 254: I'd make this check a little more generic, e.g., by checking whether max_int + 1 = min_int.

interpreter/exec/int.ml
ngzhian: Yes :)

PR #259, opened 2020-06-24
Implement i16x8 operations

ngzhian:
Abs, MinS, MinU, MaxS, MaxU, AvgrU.

Remove max_int and min_int from I16. The interaction with Int was
tricky:

1. Int relies on max_int and min_int being the real max_int and min_int
of Rep, i.e. max_int has top bit unset and rest of the bit set, and
min_int the opposite. cmp_u relies on this for correct overflowing.

2. sign_extend relied on max_int + one == min_int for checking if we are
implementing smaller ints inside of Int32. This only worked if we
defined max_int and min_int in I16 to the correct 16-bit values.

1 and 2 conflicted, so I choose to tweak the way we check 2. We check
that 1 << (bitwidth - 1) == min_int. This should be true for integers
that are fully utilizing the representative types, and false for i16
(and i8 in the future).

Amongst the operations, AvgrU was the more involved one. I chose to
implement it in I16 and I32, since it requires treating the inputs as
unsigned, widening, then addition. So in I16 we only needed to mask top
bits (since it was already stored in Int32), and in I32 we widen to I64.


ngzhian:
Argh, pushed to wrong branch, will send out another PR.

PR #260, opened 2020-06-24
Implement i16x8 operations

ngzhian:
Abs, MinS, MinU, MaxS, MaxU, AvgrU.

Remove max_int and min_int from I16. The interaction with Int was
tricky:

1. Int relies on max_int and min_int being the real max_int and min_int
of Rep, i.e. max_int has top bit unset and rest of the bit set, and
min_int the opposite. cmp_u relies on this for correct overflowing.

2. sign_extend relied on max_int + one == min_int for checking if we are
implementing smaller ints inside of Int32. This only worked if we
defined max_int and min_int in I16 to the correct 16-bit values.

1 and 2 conflicted, so I choose to tweak the way we check 2. We check
that 1 << (bitwidth - 1) == min_int. This should be true for integers
that are fully utilizing the representative types, and false for i16
(and i8 in the future).

Amongst the operations, AvgrU was the more involved one. I chose to
implement it in I16 and I32, since it requires treating the inputs as
unsigned, widening, then addition. So in I16 we only needed to mask top
bits (since it was already stored in Int32), and in I32 we widen to I64.


interpreter/exec/i16.ml
```diff
@@ -6,6 +6,10 @@ include Int.Make (struct
...
   let bitwidth = 16
 
-  let min_int = Int32.of_int (-32768)
-  let max_int = Int32.of_int 32767
+  let avgr_u x y =
+    (* Mask top bits to treat the value as unsigned. *)
+    let mask = Int32.of_int 0xffff in
+    let x_u16 = logand mask x in
+    let y_u16 = logand mask y in
+    Int32.div (Int32.add (Int32.add x_u16 y_u16) one) (Int32.of_int 2)
```
rossberg, line 14:
```suggestion
    Int32.(div (add (add x_u16 y_u16) one) (of_int 2))
```

interpreter/exec/i32.ml
```diff
@@ -4,4 +4,11 @@ include Int.Make
   (struct
     include Int32
     let bitwidth = 32
+
+    let avgr_u x y =
+      let open Int64 in
+      let mask = of_int (-1) in
+      let x64 = logand mask (of_int32 x) in
+      let y64 = logand mask (of_int32 y) in
+      to_int32 (div (add (add x64 y64) one) (of_int 2))
```
rossberg, line 13: Shouldn't this be implemented generically in the Int functor instead of repeating it for each type?

interpreter/exec/i16.ml
ngzhian: No needed anymore since the implementation is moved into Int functor.

interpreter/exec/i32.ml
ngzhian: I thought of that initially but had trouble, since we extend i16 to int32, and i32 to i64.
But we can use i64 for all cases. The additional requirement then is for Rep to provide {to,of}_int64, in order to extend to Int64.t and narrow back down into Rep.t. I added this to RepType, and implemented them in the obvious way in i16.ml, i32.ml, and i64.ml. Let me know if that looks good.

(I thought of int, but I think it will be problematic on 32-bit platforms, since not all values of int32_t will be representable if we convert to int, I think we will have some problems)

ngzhian:
(Rebased and force pushed to incorporate latest changes on master)

ngzhian:
@rossberg moved the avgr_u implementation into the function for sharing, ptal, thanks!

ngzhian:
Thanks for catching those spelling mistakes and formatting issues. Fixed in latest commit (found another typo too).

PR #267, opened 2020-07-14
Implement v128 bitwise operations

ngzhian:
These are implemented in terms of I32x4 operations, since they are
shape agnostic.

Not, and, or, xor, andnot.

### DIFF
--- a/interpreter/exec/simd.ml
+++ b/interpreter/exec/simd.ml
@@ -47,6 +47,11 @@ sig
   type lane
 
   val extract_lane : int -> t -> lane
+  val lognot : t -> t
+  val logand : t -> t -> t
+  val or_ : t -> t -> t
+  val xor : t -> t -> t
+  val andnot : t -> t -> t
   val abs : t -> t
   val neg : t -> t
   val add : t -> t -> t
@@ -143,6 +148,11 @@ struct
     let extract_lane i s = List.nth (Convert.to_shape s) i
     let unop f x = Convert.of_shape (List.map f (Convert.to_shape x))
     let binop f x y = Convert.of_shape (List.map2 f (Convert.to_shape x) (Convert.to_shape y))
+    let lognot = unop Int.lognot
+    let logand = binop Int.and_
+    let or_ = binop Int.or_
+    let xor = binop Int.xor
+    let andnot = binop (fun x y -> Int.and_ x (Int.lognot y))
     let abs = unop Int.abs
     let neg = unop Int.neg
     let add = binop Int.add
--- a/interpreter/exec/simd.ml
+++ b/interpreter/exec/simd.ml
@@ -108,7 +108,7 @@ sig
   module I64x2 : Int with type t = t and type lane = I64.t
   module F32x4 : Float with type t = t and type lane = F32.t
   module F64x2 : Float with type t = t and type lane = F64.t
-  module V128x1 : Vec with type t = t
+  module V128 : Vec with type t = t
 end
 
 module Make (Rep : RepType) : S with type bits = Rep.t =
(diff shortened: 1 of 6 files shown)


ngzhian:
Agree I32x4 is definitely arbitrary, since the lane size you choose won't matter for these operations. If you want best performance I64x2 is probably the choice, there will be less Byte operations for marshalling. I8x16 makes the most sense semantics wise, because that's the smallest "bitwise" operation you can perform. But I8x16 uses Int32 under the hood (it's really Int32 x 16).

I'm not sure what you mean by representing it differently, something like V128x1?

binji:
> I'm not sure what you mean by representing it differently, something like V128x1?

That's what I was thinking, yeah. But I don't know if that's reasonable. In any case, i32x4 seems like a fine choice. Maybe @rossberg has thoughts though.

rossberg:
Agreed with @binji: it's a bit odd to stuff these into an arbitrary int shape. Instead, I would suggest to add `vunop` and `vbinop` to `SimdOp` for ops that are shapeless, and respectively, some `Vec` submodule complementing `Int` and `Float`.

ngzhian:
@rossberg would you like to take another look at this? created the v128x1 module.
@binji for a second look too. Thanks for the review!

interpreter/exec/simd.ml
```diff
@@ -97,6 +109,7 @@ sig
   module I64x2 : Int with type t = t and type lane = I64.t
   module F32x4 : Float with type t = t and type lane = F32.t
   module F64x2 : Float with type t = t and type lane = F64.t
+  module V128x1 : Vec with type t = t and type lane = I64.t
```
rossberg, line 112: Is there a reason for not sticking to the name used in the instructions and calling this plain V128?

interpreter/exec/simd.ml
```diff
@@ -77,6 +77,18 @@ sig
   val extract_lane : int -> t -> lane
 end
 
+module type Vec =
+sig
+  type t
+  type lane
```
rossberg, line 83: Nit: remove this type. It isn't used, and v128 doesn't have lanes.

interpreter/exec/simd.ml
```diff
@@ -112,6 +125,27 @@ struct
   let to_i16x8 = Rep.to_i16x8
   let to_i32x4 = Rep.to_i32x4
 
+  module MakeVec (Int : Int.S) (Convert : sig
```
rossberg, line 128: Do we actually need to have this be a functor?

interpreter/syntax/ast.ml
```diff
@@ -52,17 +52,20 @@ struct
   type ibinop = Add | Sub | MinS | MinU | MaxS | MaxU | Mul | AvgrU
   type funop = Abs | Neg | Sqrt
   type fbinop = Add | Sub | Mul | Div | Min | Max
+  type vunop = Not
+  type vbinop = And | Or | Xor | Andnot
```
rossberg, line 56: Nit: AndNot

interpreter/exec/simd.ml
ngzhian: Because we have v128.ml, I thought that Simd.V128 might be a bit confusing.

interpreter/exec/simd.ml
ngzhian: Nope, a bit of consistency with the rest, but I can change this to be a plain module :)

interpreter/exec/simd.ml
ngzhian: And V128x1 has symmetry with the rest of the module names here. I will switch it to V128 if you like, no strong preferences either way.

interpreter/exec/simd.ml
ngzhian: Changed to a module.

interpreter/exec/simd.ml
rossberg: Yes, I think V128 is preferable, since V128x1 is not a notion the language has, AFAIU.

ngzhian: Sg, done!

PR #269, opened 2020-07-21
Add SIMD type and value

ngzhian:
This is meant to replace #193 .
Address the feedback given there:

- Issue with using the term "packed integers" in defining the value of 128-bit SIMD. Use i128 instead. Later on, when introducing SIMD instructions, we can add pseudo instructions to convert i128 into the expected lane shapes
- use v128 instead of s128, v128 is used in multiple places, BinarySIMD.md, SIMD.md, reference interpreter, plus s128 already means signed 128-bit integer.

I cherry-picked Deepti's changes from #193, excluding the merge commits, since this PR is rebased on top of master and so already have those changes. The additions/modifications I made are in the latest two commits:
- 4bd26e3ba9d08b6448e0ecdc5b967a948e31b38e
- 8e1bda60cec910c48e1de09c8ca923380226fc01

Looking at only those two might make the review easier. (using the commit tab in this PR)

### DIFF
--- a/document/core/syntax/types.rst
+++ b/document/core/syntax/types.rst
@@ -22,7 +22,7 @@ Value Types
 .. math::
    \begin{array}{llll}
    \production{value type} & \valtype &::=&
-     \I32 ~|~ \I64 ~|~ \F32 ~|~ \F64 ~|~ \S128 \\
+     \I32 ~|~ \I64 ~|~ \F32 ~|~ \F64 ~|~ \V128 \\
    \end{array}
 
 The types |I32| and |I64| classify 32 and 64 bit integers, respectively.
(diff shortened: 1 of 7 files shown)


document/core/syntax/types.rst
```diff
@@ -31,13 +31,17 @@ Integers are not inherently signed or unsigned, their interpretation is determin
 The types |F32| and |F64| classify 32 and 64 bit floating-point data, respectively.
 They correspond to the respective binary floating-point representations, also known as *single* and *double* precision, as defined by the |IEEE754|_ standard (Section 3.3).
 
+The type |V128| corresponds to 128 bit packed integer or floating-point data. The packed data
```
rossberg, line 34:
```suggestion
The type |V128| corresponds to a 128 bit vector of packed integer or floating-point data. The packed data
```

ngzhian:
@binji I think all your comments from the previous PR have been addressed. Do you mind taking a quick look again? Thanks!

PR #270, opened 2020-07-22
Implement all_true, any_true, bitselect

ngzhian:
Depends on #267 , please review bcd4fe49b2c88a136c7530cc4f535ccbb2b8f0e7 for the changes.


ngzhian:
Bitselect is the only ternary operation in SIMD.
I was deciding between making and AST that is specifically Bitselect, but think ternary is a bit cleaner (if we ever add more ternary instructions).

ngzhian:
FYI: force push is a rebase, since the depended PR was squashed before merging.

PR #271, opened 2020-07-23
Add SIMD instructions to syntax

ngzhian:
Main changes are defining:

- ixx for i8x16, i16x8, i32x4, i64x2
- fxx for f32x4, f64x2
- vxx for ixx and fxx
- the the rest of the SIMD instructions are defined based on these

Unfortunately, the definition of instructions aren't as clean as the existing ones, due to high asymmetry (see https://github.com/WebAssembly/simd/blob/master/proposals/simd/NewOpcodes.md for an overview).

This leads to a lot of specific i8x16, i16x8, i32x4 instructions defined (since i64x2 has to be left out).

I also tweaked the paragraph describing comparisons, to indicate that comparisons can return a boolean, or return a value of the same type, because the SIMD comparisons return S128. I'm not sure if this tweak should stick, we can also classify SIMD comparisons as Binary instructions, but that's confusing too. I guess I don't know if these subcategories are meant to classify instructions by "the kind of things they do" or "the type signature"?

See https://www.ngzhian.com/simd/core/syntax/instructions.html for a preview.

### DIFF
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -210,7 +210,7 @@ SIMD instructions provide basic operations over :ref:`values <syntax-value>` of 
      \K{v128.}\vsunop \\&&|&
      \K{v128.}\vsbinop \\&&|&
      \K{v128.}\vsternop \\&&|&
-     \K{v8x16.}\SHUFFLE~\laneidx^{16} ~|~ \K{v8x16.}\SWIZZLE \\&&|&
+     \K{i8x16.}\SHUFFLE~\laneidx^{16} ~|~ \K{i8x16.}\SWIZZLE \\&&|&
      \X{vxx}\K{.}\SPLAT \\&&|&
      \K{i8x16.}\EXTRACTLANE\K{\_}\sx~\laneidx ~|~
      \K{i16x8.}\EXTRACTLANE\K{\_}\sx~\laneidx \\&&|&
@@ -424,10 +424,13 @@ Instructions in this group are concerned with linear :ref:`memory <syntax-mem>`.
      \K{i}\X{nn}\K{.}\STORE\K{8}~\memarg ~|~
      \K{i}\X{nn}\K{.}\STORE\K{16}~\memarg ~|~
      \K{i64.}\STORE\K{32}~\memarg \\&&|&
-     \K{i16x8.}\LOAD\K{8x8}\_\sx~\memarg ~|~
-     \K{i32x4.}\LOAD\K{16x4}\_\sx~\memarg ~|~
-     \K{i64x2.}\LOAD\K{32x2}\_\sx~\memarg \\&&|&
-     \X{vxx}\K{.}\LOAD\K{\_splat}~\memarg \\&&|&
+     \K{v128.}\LOAD\K{8x8}\_\sx~\memarg ~|~
+     \K{v128.}\LOAD\K{16x4}\_\sx~\memarg ~|~
+     \K{v128.}\LOAD\K{32x2}\_\sx~\memarg \\&&|&
+     \K{v128.}\LOAD\K{8\_splat}~\memarg ~|~
+     \K{v128.}\LOAD\K{16\_splat}~\memarg \\&&|&
+     \K{v128.}\LOAD\K{32\_splat}~\memarg ~|~
+     \K{v128.}\LOAD\K{64\_splat}~\memarg \\&&|&
      \MEMORYSIZE \\&&|&
      \MEMORYGROW \\
    \end{array}
@@ -437,7 +440,7 @@ They all take a *memory immediate* |memarg| that contains an address *offset* an
 Integer loads and stores can optionally specify a *storage size* that is smaller than the :ref:`bit width <syntax-valtype>` of the respective value type.
 In the case of loads, a sign extension mode |sx| is then required to select appropriate behavior.
 
-SIMD loads can specify a shape that is half the :ref:`bit width <syntax-valtype>` of |V128|. Each lane is half its usual size, and the sign extension mode |sx| then specifies how the smaller lane is extended to the larger lane. Alternatively, SIMD loads can perform a *splat*, such that only a single lane of the respective shape is loaded, and the result is duplicated to all other lanes.
+SIMD loads can specify a shape that is half the :ref:`bit width <syntax-valtype>` of |V128|. Each lane is half its usual size, and the sign extension mode |sx| then specifies how the smaller lane is extended to the larger lane. Alternatively, SIMD loads can perform a *splat*, such that only a single lane of the specified storage size is loaded, and the result is duplicated to all lanes.
 
 The static address offset is added to the dynamic address operand, yielding a 33 bit *effective address* that is the zero-based index at which the memory is accessed.
 All values are read and written in |LittleEndian|_ byte order.
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -437,7 +437,7 @@ They all take a *memory immediate* |memarg| that contains an address *offset* an
 Integer loads and stores can optionally specify a *storage size* that is smaller than the :ref:`bit width <syntax-valtype>` of the respective value type.
 In the case of loads, a sign extension mode |sx| is then required to select appropriate behavior.
 
-SIMD loads can specify a shape that is half the :ref:`bit width <syntax-valtype>` of |V128|. Each lane is half its usual size, and the sign extension mode |sx| then specifies how the smaller lane is extended to the larger lane. SIMD loads can be annotated with *splat*, to indicate that only a single lane of the respective shape is loaded, and the result is duplicated to all other lanes.
+SIMD loads can specify a shape that is half the :ref:`bit width <syntax-valtype>` of |V128|. Each lane is half its usual size, and the sign extension mode |sx| then specifies how the smaller lane is extended to the larger lane. Alternatively, SIMD loads can perform a *splat*, such that only a single lane of the respective shape is loaded, and the result is duplicated to all other lanes.
 
 The static address offset is added to the dynamic address operand, yielding a 33 bit *effective address* that is the zero-based index at which the memory is accessed.
 All values are read and written in |LittleEndian|_ byte order.
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -172,6 +172,7 @@ Occasionally, it is convenient to group operators together according to the foll
 
 .. index:: ! simd instruction, fixed-width simd, value, value type
    pair: abstract syntax; instruction
+.. _syntax-laneidx:
 .. _syntax-vunop:
 .. _syntax-vbinop:
 .. _syntax-vsunop:
@@ -202,20 +203,21 @@ SIMD instructions provide basic operations over :ref:`values <syntax-value>` of 
      \K{f32x4} ~|~ \K{f64x2} \\
    \production{vshape} & \X{vxx} &::=&
      \X{ixx} ~|~ \X{fxx} \\
+   \production{lane index} & \laneidx &::=& \byte \\
    \production{instruction} & \instr &::=&
      \dots \\&&|&
      \K{v128.}\CONST~\xref{syntax/values}{syntax-simd}{\vX{\X{nnn}}} \\&&|&
      \K{v128.}\vsunop \\&&|&
      \K{v128.}\vsbinop \\&&|&
      \K{v128.}\vsternop \\&&|&
-     \K{v8x16.}\SHUFFLE ~|~ \K{v8x16.}\SWIZZLE \\&&|&
+     \K{v8x16.}\SHUFFLE~\laneidx^{16} ~|~ \K{v8x16.}\SWIZZLE \\&&|&
      \X{vxx}\K{.}\SPLAT \\&&|&
-     \K{i8x16.}\EXTRACTLANE\K{\_}\sx ~|~
-     \K{i16x8.}\EXTRACTLANE\K{\_}\sx \\&&|&
-     \K{i32x4.}\EXTRACTLANE ~|~
-     \K{i64x2.}\EXTRACTLANE \\&&|&
-     \X{fxx}\K{.}\EXTRACTLANE \\&&|&
-     \X{vxx}\K{.}\REPLACELANE \\&&|&
+     \K{i8x16.}\EXTRACTLANE\K{\_}\sx~\laneidx ~|~
+     \K{i16x8.}\EXTRACTLANE\K{\_}\sx~\laneidx \\&&|&
+     \K{i32x4.}\EXTRACTLANE~\laneidx ~|~
+     \K{i64x2.}\EXTRACTLANE~\laneidx \\&&|&
+     \X{fxx}\K{.}\EXTRACTLANE~\laneidx \\&&|&
+     \X{vxx}\K{.}\REPLACELANE~\laneidx \\&&|&
      \X{ixx}\K{.}\virelop \\&&|&
      \X{fxx}\K{.}\vfrelop \\&&|&
      \K{i8x16.}\viunop ~|~
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -208,7 +208,8 @@ SIMD instructions provide basic operations over :ref:`values <syntax-value>` of 
      \K{v128.}\vsunop \\&&|&
      \K{v128.}\vsbinop \\&&|&
      \K{v128.}\vsternop \\&&|&
-     \K{i8x16.}\SHUFFLE~\laneidx^{16} ~|~ \K{i8x16.}\SWIZZLE \\&&|&
+     \K{i8x16.}\SHUFFLE~\laneidx^{16} \\&&|&
+     \K{i8x16.}\SWIZZLE \\&&|&
      \X{vxx}\K{.}\SPLAT \\&&|&
      \K{i8x16.}\EXTRACTLANE\K{\_}\sx~\laneidx ~|~
      \K{i16x8.}\EXTRACTLANE\K{\_}\sx~\laneidx \\&&|&
@@ -318,9 +319,6 @@ Operations are performed point-wise on the values of each lane.
    The bitwidth of the numeric type :math:`t` times :math:`N` always is 128.
 
 Instructions prefixed with :math:`\K{v128}` do not involve a specific interpretation, and treat the |V128| as an |i128| value or a vector of 128 individual bits.
-
-.. todo::
-  write up runtime interpretation for the lane shapes
 
 SIMD instructions can be grouped into several subcategories:
 
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -336,11 +336,11 @@ SIMD instructions can be grouped into several subcategories:
 
 * *Shifts*: consume a |v128| operand and a |i32| operand, producing one |V128| result.
 
-* *Splats*: consume a value of the integer or floating-point type and produce a |V128| result of a specified shape.
+* *Splats*: consume a value of numeric type and produce a |V128| result of a specified shape.
 
-* *Extract lanes*: consume a |V128| operand and an immediate byte specifying the lane index and produce a result of the element type.
+* *Extract lanes*: consume a |V128| operand and return the numeric value in a given lane.
 
-* *Replace lanes*: consume a |V128| operand, an immediate byte specifying the lane index, and a value of the element type, and produce a |V128| result.
+* *Replace lanes*: consume a |V128| operand and a numeric value for a given lane, and produce a |V128| result.
 
 Some SIMD instructions have a signedness annotation |sx| which distinguishes whether the elements in the operands are to be :ref:`interpreted <aux-signed>` as :ref:`unsigned <syntax-uint>` or :ref:`signed <syntax-sint>` integers.
 For the other SIMD instructions, the use of two's complement for the signed interpretation means that they behave the same regardless of signedness.
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -181,6 +181,7 @@ Occasionally, it is convenient to group operators together according to the foll
 .. _syntax-vshiftop:
 .. _syntax-viunop:
 .. _syntax-vibinop:
+.. _syntax-viminmaxop:
 .. _syntax-vsatbinop:
 .. _syntax-vfunop:
 .. _syntax-vfbinop:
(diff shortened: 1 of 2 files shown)


document/core/syntax/instructions.rst
```diff
@@ -71,6 +93,49 @@ These operations closely match respective operations available in hardware.
...
+     \K{i32x4.}\EXTRACTLANE ~|~
+     \K{i64x2.}\EXTRACTLANE \\&&|&
+     \X{fxx}\K{.}\EXTRACTLANE \\&&|&
+     \X{vxx}\K{.}\REPLACELANE \\&&|&
+     \X{ixx}\K{.}\irelop \\&&|&
+     \X{fxx}\K{.}\frelop \\&&|&
+     \K{i8x16.}\viunop ~|~
+     \K{i16x8.}\viunop ~|~
+     \K{i32x4.}\viunop \\&&|&
+     \K{i64x2.}\NEG \\&&|&
```
binji, line 109: I wonder if it would be clearer to use patterns everywhere and exclude forms that are not allowed; there doesn't seem to be precedent for that, however.

document/core/syntax/instructions.rst
```diff
@@ -105,8 +170,20 @@ These operations closely match respective operations available in hardware.
...
      \K{max} ~|~
      \K{copysign} \\
+   \production{SIMD unary operator} & \vunop &::=&
+     \K{not} \\
+   \production{SIMD binary operator} & \vbinop &::=&
+     \K{and} ~|~
+     \K{andnot} ~|~
+     \K{or} ~|~
+     \K{xor} \\
+   \production{SIMD unary operator} & \vternop &::=&
```
binji, line 180: ternary

document/core/syntax/instructions.rst
```diff
@@ -134,15 +237,25 @@ For each type, several subcategories can be distinguished:
...
 * *Tests*: consume one operand of the respective type and produce a Boolean integer result.
 
-* *Comparisons*: consume two operands of the respective type and produce a Boolean integer result.
+* *Comparisons*: consume two operands of the respective type and produce a Boolean integer result or a result of the respective type.
 
 * *Conversions*: consume a value of one type and produce a result of another
   (the source type of the conversion is the one after the ":math:`\K{\_}`").
 
+.. todo::
+  Do these subcategories have to cover every instruction? E.g. simd shifts don't fit anywhere here, since they take 128-bit int and a 32-bit int.
```
binji, line 246: I think it might be clearer to group the SIMD operations separately.

document/core/syntax/instructions.rst
```diff
@@ -154,9 +267,22 @@ Occasionally, it is convenient to group operators together according to the foll
...
      \iunop ~|~
      \funop ~|~
+     \vunop ~|~
+     \viunop ~|~
+     \vfunop ~|~
      \EXTEND{N}\K{\_s} \\
-   \production{binary operator} & \binop &::=& \ibinop ~|~ \fbinop \\
-   \production{test operator} & \testop &::=& \itestop \\
+   \production{binary operator} & \binop &::=&
+   \ibinop ~|~
```
binji, line 275: nit: indent

document/core/syntax/instructions.rst
ngzhian: That will be kinda neat, but the asymmetry is pretty bad, such an exclusion list might end up being unwieldy too. Probably a short form for i8x16, i16x8, i32x4 will get us part way there (what would a name for that be small_ixx?)

document/core/syntax/instructions.rst
binji: Yeah, I think that would be OK. But I'd prefer separating out comparisons too, since those work pretty differently with SIMD.

document/core/syntax/instructions.rst
binji: Yeah, not sure the best name to use. Thoughts on this, @rossberg?

document/core/syntax/instructions.rst
```diff
@@ -47,15 +57,27 @@ These operations closely match respective operations available in hardware.
    \begin{array}{llcl}
    \production{width} & \X{nn}, \X{mm} &::=&
      \K{32} ~|~ \K{64} \\
+   \production{simdwidth} & \X{sss} &::=&
```
rossberg, line 60: This should be `nnn`. But is it even worth introducing it as long as we only have one SIMD type?

document/core/syntax/instructions.rst
```diff
@@ -35,6 +35,16 @@ The following sections group instructions into a number of different categories.
...
+.. _syntax-vunop:
+.. _syntax-vbinop:
+.. _syntax-vternop:
+.. _syntax-vtestop:
+.. _syntax-vshiftop:
+.. _syntax-viunop:
+.. _syntax-vibinop:
+.. _syntax-vsatbinop:
+.. _syntax-vfunop:
+.. _syntax-vfbinop:
```
rossberg, line 47: I think it's nicer if we put all the SIMD instructions into their own new section.

document/core/syntax/instructions.rst
rossberg: In the context of memory access I've called the smaller int types "packed", but does that really fit here? Also, it looks like the irregularity is high enough that we need at least two subsets. And then multiple sets of viunop/vibinop. It may not be worth it.

document/core/syntax/instructions.rst
```diff
@@ -71,6 +93,49 @@ These operations closely match respective operations available in hardware.
...
+     \K{i16x8.}\WIDEN\K{\_low}\K{\_i8x16\_}\sx ~|~
+     \K{i32x4.}\WIDEN\K{\_low}\K{\_i16x8\_}\sx \\&&|&
+     \K{i16x8.}\WIDEN\K{\_high}\K{\_i8x16\_}\sx ~|~
+     \K{i32x4.}\WIDEN\K{\_high}\K{\_i16x8\_}\sx \\&&|&
+     \X{ixx}\K{.}\vshiftop \\&&|&
+     \K{i8x16.}\vibinop ~|~
+     \K{i16x8.}\vibinop ~|~
+     \K{i32x4.}\vibinop \\&&|&
+     \K{i64x2.}\ADD ~|~
+     \K{i64x2.}\SUB \\&&|&
```
rossberg, line 128: I'm confused, why are these not binops?

document/core/syntax/instructions.rst
rossberg: As noted in my comment above, I think we should not lump SIMD under numeric instructions. They are different beasts. And then we can define different subcategories for them.

document/core/syntax/instructions.rst
ngzhian: These are binops, but saying `\K{i64x2.}\vibinop` here will not work, since vibinop includes `min_{sx}` and `max_{sx}`, which is not available on i64x2.

Or do you mean, why these 2 instructions are not in the `binops` part of the "Conventions" section below? If this is the case, I probably missed them. I can add them there.

document/core/syntax/instructions.rst
ngzhian: Changed to nnn. Probably not worth it, inlined it to be v128 in all the necessary instructions.

document/core/syntax/instructions.rst
ngzhian: Sg, added a new section: SIMD instructions.

document/core/syntax/instructions.rst
ngzhian: Moved SIMD instructions into their own section. Called out "shift" instructions in the subcategory.
I did not add a "convention" section to add `vbinop` as a shorthand for `vibinop` and `vfbinop` yet, I think we can add those in as needed when we get to the validation chapter.

document/core/syntax/instructions.rst
ngzhian: Yea, packed doesn't fit as well here. We kinda use "pack" to mean packing integers/floats side by side. I agree that it might not be worth it. Let's keep this for now, as I work on this, I'll stare at this and see if I can come up with something better. (And I think future work on the validation/execution portion will help us see if there is a better way to organize this.)

document/core/syntax/instructions.rst
```diff
@@ -235,15 +410,21 @@ Instructions in this group are concerned with linear :ref:`memory <syntax-mem>`.
...
-     \K{f}\X{nn}\K{.}\STORE~\memarg \\&&|&
+     \K{f}\X{nn}\K{.}\STORE~\memarg ~|~
+     \K{v128.}\STORE~\memarg \\&&|&
      \K{i}\X{nn}\K{.}\LOAD\K{8\_}\sx~\memarg ~|~
      \K{i}\X{nn}\K{.}\LOAD\K{16\_}\sx~\memarg ~|~
      \K{i64.}\LOAD\K{32\_}\sx~\memarg \\&&|&
      \K{i}\X{nn}\K{.}\STORE\K{8}~\memarg ~|~
      \K{i}\X{nn}\K{.}\STORE\K{16}~\memarg ~|~
      \K{i64.}\STORE\K{32}~\memarg \\&&|&
+     \K{i16x8.}\LOAD\K{8x8}\_\sx ~|~
```
binji, line 424: I think we'll want to explain the widening and splat load instructions below too.

document/core/syntax/instructions.rst
```diff
@@ -170,6 +170,181 @@ Occasionally, it is convenient to group operators together according to the foll
...
+* *Shifts*: consume a |v128| operand and a |i32| operand, producing one |V128| result.
+
+* *Extract lanes*: consume a |V128| operand and an immediate byte specifying the lane index and produce a result of the element type.
+
+* *Replace lanes*: consume a |V128| operand, an immediate byte specifying the lane index, and a value of the element type, and produce a |V128| result.
+
+* *Conversions/Splats*: consume a value of the integer or floating-point type and produce a |V128| result of a specified shape.
+
+.. todo::
+   should comparisons be called out in a separate subcategory? they are essentially the same as binary operations
```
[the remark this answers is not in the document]
ngzhian: Conversion (in the sense of i32x4 <-> f32x4) can probably go into vbinop (same signature).
I meant this bullet point to be about splats, which are conversions in the same sense as the numeric conversions which convert between types.
I think calling it "Conversions/Splats" confuses things, I'll simply call it "Conversions" (and the instructions that make up the conversions are splats).

document/core/syntax/instructions.rst
```diff
@@ -170,6 +170,181 @@ Occasionally, it is convenient to group operators together according to the foll
...
+.. todo::
+  write up runtime interpretation for the lane shapes
+
+SIMD instructions can be grouped into several subcategories:
+
+* *Constants*: return a static constant.
+
+* *Unary Operations*: consume one |V128| operand and produce one |V128| result.
+
+* *Binary Operations*: consume two |V128| operands and produce one |V128| result.
```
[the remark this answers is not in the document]
ngzhian: Good point, done!

document/core/syntax/instructions.rst
ngzhian: Added a paragraph on this, ptal at 1a8af6d.

document/core/syntax/instructions.rst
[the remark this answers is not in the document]
ngzhian: Fixed in de59417

document/core/syntax/instructions.rst
```diff
@@ -170,6 +170,176 @@ Occasionally, it is convenient to group operators together according to the foll
...
+     \K{mul} ~|~
+     \K{div} ~|~
+     \K{min} ~|~
+     \K{max} \\
+   \end{array}
+
+SIMD instructions have a naming convention that
+determines how their operands will be interpreted. An instruction beginning with :math:`\K{i32x4}`
+will interpret its operands as four |i32|, packed side-by-side into a |i128|.
+This prefix, :math:`\K{i32x4}`, is known as the *shape* of the type, and is made up of the underlying element type, :math:`\K{i32}`, and the number of elements or *lanes*, :math:`\K{4}`. Operations are performed lane-wise on each element.
```
rossberg, line 312:
```suggestion
SIMD instructions have a naming convention involving a prefix that
determines how their operands will be interpreted.
This prefix describes the *shape* of the operand,
written :math:`t\K{x}N`, and consisting of a packed numeric type :math:`t` and the number of *lanes* :math:`N` of that type.
Operations are performed point-wise on the values of each lane.

.. note::
   For example, the shape :math:`\K{i32x4}` interprets the operand
as four |i32| values, packed into an |i128|.
   The bitwidth of the numeric type :math:`t` times :math:`N` always is 128.
```

document/core/syntax/instructions.rst
```diff
@@ -170,6 +170,176 @@ Occasionally, it is convenient to group operators together according to the foll
...
+
+* *Binary Operations*: consume two |V128| operands and produce one |V128| result.
+
+* *Ternary Operations*: consume three |V128| operands and produce one |V128| result.
+
+* *Tests*: consume one |V128| operand and produce a Boolean integer result.
+
+* *Shifts*: consume a |v128| operand and a |i32| operand, producing one |V128| result.
+
+* *Splats*: consume a value of the integer or floating-point type and produce a |V128| result of a specified shape.
```
rossberg, line 333:
```suggestion
* *Splats*: consume a value of numeric type and produce a |V128| result of a specified shape.
```

document/core/syntax/instructions.rst
```diff
@@ -170,6 +170,176 @@ Occasionally, it is convenient to group operators together according to the foll
...
+
+* *Extract lanes*: consume a |V128| operand and an immediate byte specifying the lane index and produce a result of the element type.
```
rossberg, line 335:
```suggestion
* *Extract lanes*: consume a |V128| operand and return the numeric value in a given lane.
```

document/core/syntax/instructions.rst
```diff
@@ -170,6 +170,176 @@ Occasionally, it is convenient to group operators together according to the foll
...
+
+* *Replace lanes*: consume a |V128| operand, an immediate byte specifying the lane index, and a value of the element type, and produce a |V128| result.
```
rossberg, line 337:
```suggestion
* *Replace lanes*: consume a |V128| operand and a numeric value for a given lane, and produce a |V128| result.
```

document/core/syntax/instructions.rst
ngzhian: This list contains all the operations, so there will not be more ops (binops) that will need to be enumerated.
Yes I can definitely take min/max out (not sure what I would call them though, let me come up with something and you can comment on my poor names :) )

2 reasons:
- we didn't see any usage of these during the discussion and benchmarking phase when re-considering i64x2 instructions
- difficult to implement these in a performant manner, especially on ia32 and arm v7

document/core/syntax/instructions.rst
```diff
@@ -170,6 +170,176 @@ Occasionally, it is convenient to group operators together according to the foll
...
+     \K{i16x8.}\viunop ~|~
+     \K{i32x4.}\viunop \\&&|&
+     \K{i64x2.}\NEG \\&&|&
+     \X{fxx.}\vfunop \\&&|&
+     \K{i8x16.}\vtestop ~|~
+     \K{i16x8.}\vtestop ~|~
+     \K{i32x4.}\vtestop \\&&|&
+     \K{i8x16.}\BITMASK ~|~
+     \K{i16x8.}\BITMASK ~|~
+     \K{i32x4.}\BITMASK \\&&|&
```
ngzhian, line 230: What do you think about merging BITMASK into vtestop. Their signatures are the same `v128 -> i32`. Though semantically Bitmask is not a "test".

document/core/syntax/instructions.rst
ngzhian: Made `viminmaxop` in cbb383ee5e41e58a43c75544c33c86eab514554c

document/core/syntax/instructions.rst
```diff
@@ -170,6 +170,182 @@ Occasionally, it is convenient to group operators together according to the foll
...
+     \K{f32x4} ~|~ \K{f64x2} \\
+   \production{vshape} & \X{vxx} &::=&
+     \X{ixx} ~|~ \X{fxx} \\
+   \production{instruction} & \instr &::=&
+     \dots \\&&|&
+     \K{v128.}\CONST~\xref{syntax/values}{syntax-simd}{\vX{\X{nnn}}} \\&&|&
+     \K{v128.}\vsunop \\&&|&
+     \K{v128.}\vsbinop \\&&|&
+     \K{v128.}\vsternop \\&&|&
+     \K{v8x16.}\SHUFFLE ~|~ \K{v8x16.}\SWIZZLE \\&&|&
```
ngzhian, line 211: Do we need to talk about `v8x16` below? Or perhaps renaming this to `i8x16`.
The reason `v8x16` was chosen is that the operation doesn't care about the type, just that it is an 8-bit value.
But `v8` isn't really defined anywhere (and is not a "packed numeric type" like the description of *shape* below mentions).

ngzhian:
I noticed that some instructions that take immediates were not properly specified:
- v8x16.shuffle
- extract_lane
- replace_lane

Introduced a new production `laneidx`, which is `byte`, and added it to the appropriate instructions. see 18049c7.

document/core/syntax/instructions.rst
```diff
@@ -235,15 +413,21 @@ Instructions in this group are concerned with linear :ref:`memory <syntax-mem>`.
...
      \K{i}\X{nn}\K{.}\LOAD\K{8\_}\sx~\memarg ~|~
      \K{i}\X{nn}\K{.}\LOAD\K{16\_}\sx~\memarg ~|~
      \K{i64.}\LOAD\K{32\_}\sx~\memarg \\&&|&
      \K{i}\X{nn}\K{.}\STORE\K{8}~\memarg ~|~
      \K{i}\X{nn}\K{.}\STORE\K{16}~\memarg ~|~
      \K{i64.}\STORE\K{32}~\memarg \\&&|&
+     \K{i16x8.}\LOAD\K{8x8}\_\sx~\memarg ~|~
+     \K{i32x4.}\LOAD\K{16x4}\_\sx~\memarg ~|~
+     \K{i64x2.}\LOAD\K{32x2}\_\sx~\memarg \\&&|&
+     \X{vxx}\K{.}\LOAD\K{\_splat}~\memarg \\&&|&
```
ngzhian, line 430: This is incorrect. vxx represents ixx and fxx (i8x16... f64x2). It should be `v8x16`, `v16x8`, `v32x4`, `v64x2`, might need another production for here.

document/core/syntax/instructions.rst
rossberg: Seems like a stretch, I think I'd leave it separate.

document/core/syntax/instructions.rst
rossberg: Is there a chance we can just rename all the `vNxM` prefixes to `iNxM`? I understand the reason is that it doesn't matter that they are ints, but the same is true for, say, plain `i32.and` or `i32.load`.

document/core/syntax/instructions.rst
```diff
@@ -253,6 +437,8 @@ They all take a *memory immediate* |memarg| that contains an address *offset* an
 Integer loads and stores can optionally specify a *storage size* that is smaller than the :ref:`bit width <syntax-valtype>` of the respective value type.
 In the case of loads, a sign extension mode |sx| is then required to select appropriate behavior.
 
+SIMD loads can specify a shape that is half the :ref:`bit width <syntax-valtype>` of |V128|. Each lane is half its usual size, and the sign extension mode |sx| then specifies how the smaller lane is extended to the larger lane. SIMD loads can be annotated with *splat*, to indicate that only a single lane of the respective shape is loaded, and the result is duplicated to all other lanes.
```
rossberg, line 440:
```suggestion
SIMD loads can specify a shape that is half the :ref:`bit width <syntax-valtype>` of |V128|. Each lane is half its usual size, and the sign extension mode |sx| then specifies how the smaller lane is extended to the larger lane. Alternatively, SIMD loads can perform a *splat*, such that only a single lane of the respective shape is loaded, and the result is duplicated to all other lanes.
```

document/core/syntax/instructions.rst
ngzhian: The difference is that `i32.load`  returns `i32` and we have `f32.load` if we want a `f32`, whereas these all return a `v128`. Naming it with a `i` prefix suggests that it is implemented using integer loads, where it is not always the case, and also might cause confusion because there is no `f` counterpart for loading floats.

That said, I don't have a strong preference either way. I can file an issue and see if anyone has comments.
@tlively if you have comments on this re: toolchain or initial rationale for naming loads.

tlively: I actually would prefer using `i` prefixes for these as well because the `v` prefixes are used on few enough instructions that they seem surprising and because they do not correspond to anything in MVP. I recall [discussing](https://github.com/WebAssembly/simd/pull/82#issuecomment-519893850) this for specific instructions in the past, but the `v` prefix was [present](https://github.com/WebAssembly/simd/pull/1/files#diff-28263aca205221ef5f3c6b4a232654c3R281-R285) in the original proposal writeup, so I don't think we've discussed making this change for all instructions before.

There's no toolchain-specific reason to make this change, though.

binji: How much simpler is the definition here if we make the change? I have to admit I'm a bit hesitant to suggest changing the instruction names for the simplicity of the spec document.

ngzhian: > How much simpler is the definition here if we make the change?

This definition of load_splats will become `ixx.load_splat`.

If we stick to `v8x16` et al, we will need to:

1. Define `vX`. Right now we rely on the definition of `iN` to describe the *shape* of SIMD, using `vX` means we need to somehow describe `vX`, but in a way that makes it not part of the syntax
2. After `vX` we can talk about `vXxN`

It won't be a lot more text, but it will be very specialized to these 6 instructions, which (like Thomas says) does not have any analogues in MVP.

We *could* define `f32x4.load_splat` and `f64x2.load_splat`, but have them be "aliases", so share the same opcode and semantics as the integer counterparts...

tlively: I would also be fine just having `i32x4.load_splat` and friends and not having `f32x4.load_splat`. We could also avoid the problem (for these instructions) by renaming them to `v128.load<width>_splat` (and following suit for the load-and-zero instructions).

ngzhian: Yea I quite like the `v128.load<width>_splat` notation we were discussing in #237 , then we can side-step this issue for load_splat, and rename swizzle and shuffle to `i8x16`.

ngzhian: Filed #297 to discuss this more, and see if anyone supports/objects.

ngzhian: #297 hasn't had too many comments outside of the people already involved here, but I think it will be a fine change. The only other change is `v8x16.shuffle` and `v8x16.swizzle`, which I have filed #316 for.

ngzhian: I have optimistically made the naming changes (of load splats, load extends, shuffle, swizze) in the latest commit. With that, we no longer have special `v...` prefix, so no need to define them, and loads all start with `v128`.

document/core/syntax/instructions.rst
ngzhian: No valid anymore, these have been renamed to use the `i8x16` prefix.

document/core/syntax/instructions.rst
```diff
@@ -210,7 +208,7 @@ SIMD instructions provide basic operations over :ref:`values <syntax-value>` of
      \K{v128.}\vsunop \\&&|&
      \K{v128.}\vsbinop \\&&|&
      \K{v128.}\vsternop \\&&|&
-     \K{v8x16.}\SHUFFLE~\laneidx^{16} ~|~ \K{v8x16.}\SWIZZLE \\&&|&
+     \K{i8x16.}\SHUFFLE~\laneidx^{16} ~|~ \K{i8x16.}\SWIZZLE \\&&|&
```
rossberg, line 211: Nit: two lines, since they are different ops

rossberg (approved):

PR #272, opened 2020-07-24
Implement shl and shr_s for v128

ngzhian:
shr_u is a bit more tricky due for I8 and I16. E.g. -128 in I8 has all
top bits set (sign extended) in an int32, and shr_u using Int's
implementation will result in 0xffffffc0, instead of 0x000000c0. But we
can't simply change the implementation, since parsing relies on this
behavior. I'll figure that out in a later patch.


interpreter/syntax/operators.ml
```diff
@@ -227,6 +227,8 @@ let v128_bitselect = Ternary (V128Op.Bitselect)
 let i8x16_neg = Unary (V128 (V128Op.I8x16 V128Op.Neg))
 let i8x16_any_true = Test (V128 (V128Op.I8x16 V128Op.AnyTrue))
 let i8x16_all_true = Test (V128 (V128Op.I8x16 V128Op.AllTrue))
+let i8x16_shl = Shift (V128Op.I8x16 V128Op.Shl)
+let i8x16_shr_s = Shift (V128Op.I8x16 V128Op.ShrS)
```
rossberg, line 231: Nit: could shorten this to `Shift V128Op.(I8x16 Shl)`. Similarly many of the other cases.

interpreter/syntax/ast.ml
```diff
@@ -135,6 +138,7 @@ and instr' =
   | Ternary of ternop                 (* ternary numeric operator *)
   | Convert of cvtop                  (* conversion *)
   | ExtractLane of extractop          (* extract lane from v128 value *)
+  | Shift of shiftop                  (* shifts for v128 value *)
```
rossberg, line 141: Perhaps rename this to SimdShift for clarity? And symmetrically, ExtractLane -> SimdLane?

interpreter/syntax/operators.ml
ngzhian: Good tip, I'll change only this for now, then make the change for the rest (to reduce some local merge conflicts).

interpreter/syntax/ast.ml
ngzhian: I'll rename Shift here, and extractlane can go in another PR i have for implementing more of extract lane, thanks!

ngzhian:
Force push due to rebebasing.

PR #273, opened 2020-07-24
Implement SIMD splats

ngzhian:
i8x16, i16x8, i32x4, i64x2, f32x4, f64x2 splats.

Splats can really be implemented with the Convert ast nodes too. I'm not
sure which one is better, tried it out in #274.


interpreter/syntax/ast.ml
```diff
@@ -134,6 +137,7 @@ and instr' =
   | Binary of binop                   (* binary numeric operator *)
   | Ternary of ternop                 (* ternary numeric operator *)
   | Convert of cvtop                  (* conversion *)
+  | Splat of splatop                (* duplicate value into a v128 *)
```
binji, line 140: nit: align comment

ngzhian:
Abandoned in favor of #274.

PR #274, opened 2020-07-24
Implement SIMD splats using Convert

ngzhian:
i8x16, i16x8, i32x4, i64x2, f32x4, f64x2 splats.

Implemented using Convert ast as a splat essentially takes a value of
every other type to V128.

Uploaded as a PR for comparison with the other implementation using a new Splat AST node see #273


ngzhian:
Sg, thanks both for the review! Latest commit strips out some comment out code I forgot to remove.

PR #277, opened 2020-07-27
Fix parsing of splat

ngzhian:
Lost the token in the parser during rebases.


ngzhian:
Thanks for the quick review, Ben!

PR #279, opened 2020-07-27
Set OCaml version to 4.08.1

ngzhian:
The Bytes [0] functions we use require 4.08.1. This should only be set
on SIMD repo, since bytes is only used for SIMD implementation. By the
time we would like to merge this back into main spec, 4.08.1 should
hopefully be widespread enough for us to bump minimum version of OCaml
for spec to 4.08.1 too.

[0] https://caml.inria.fr/pub/docs/manual-ocaml/libref/Bytes.html see
"Binary encoding/decoding of integers"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,7 +16,7 @@ addons:
       - yarn
 
 install:
-  - opam init --auto-setup --compiler=4.07.1
+  - opam init --auto-setup --compiler=4.08.1
   - eval $(opam env)
   - opam --version
   - ocaml --version


ngzhian:
@rossberg fyi, we had discussions about 4.08 requirements before. This change should be fine in SIMD branch.

rossberg:
Sure, sounds good to me. Since we decided to drop support for 4.02 on the main repo I think there is little reason to stay far behind on OCaml versions. 4.11 is about to be released, so 4.08.X seems pretty moderate.

PR #281, opened 2020-07-27
Implement extract lane for the remaining shapes

ngzhian:
i8x16, i16x8, i64x2, f64x2.

Created a new extractop that is based on v128op. Use extension type to
determine signed or unsigned extract (only used for i8x16 and i16x8).

Had to introduce an unsigned_mask in simd.ml, because i8 and i16 are stored in i32, singed, so negative values would be extracted incorrect (-1 of type i8 is stored as 0xFFFFFFFF, extract unsigned should be 0x000000FF).


interpreter/exec/simd.ml
```diff
@@ -48,6 +48,7 @@ sig
 
   val splat : lane -> t
   val extract_lane : int -> t -> lane
+  val extract_lane_u : int -> t -> lane
```
rossberg, line 51: If this is named `_u`, shouldn't the other be named `_s`?

interpreter/exec/simd.ml
```diff
@@ -200,12 +201,18 @@ struct
       val to_shape : Rep.t -> Int.t list
       val of_shape : Int.t list -> Rep.t
       val num_lanes : int
+      (* For I8 and I16, since they are stored signed in an int32, this mask
+       * is used for unsigned extracts, to get the right unsinged value. *)
+      val unsigned_mask : Int.t
```
rossberg, line 206: I'm not sure I understand why this needs to leak here. Shouldn't that be encapsulated in the Int8/Int16 modules?

interpreter/valid/valid.ml
```diff
@@ -169,6 +177,16 @@ let check_memop (c : context) (memop : 'a memop) get_sz at =
   require (1 lsl memop.align <= size) at
     "alignment must not be larger than natural"
 
+let check_simd_lane_idx op at =
+  let max, idx = match op with
+  | V128Op.I8x16 (_, idx) -> 16, idx
```
rossberg, line 182: Nit: indent += 2

interpreter/exec/simd.ml
ngzhian: Move this into Int.ml, created an `as_unsigned` method there.

interpreter/exec/simd.ml
ngzhian: Good point, done!

interpreter/syntax/ast.ml
```diff
@@ -74,7 +74,10 @@ struct
   type relop = TodoRelOp
   type vcvtop = Splat
   type cvtop = (vcvtop, vcvtop, vcvtop, vcvtop, vcvtop, vcvtop, vcvtop) v128op
-  type extractop = I32x4ExtractLane of int | F32x4ExtractLane of int
+  type e =
```
binji, line 77: better name here?

ngzhian: Done!

PR #282, opened 2020-07-27
v128 load and store

ngzhian:
Depends on #281. Please review commit 2e5c9d4 for the changes.

I couldn't figure out a cleaner way to write `load_value`, since `loadn` and `storen` both support 64 bit ints.


ngzhian:
Force push after rebase on merged #281 .

PR #283, opened 2020-07-28
Implement v8x16.swizzle

ngzhian:
Decided to cheat a little bit, and squeeze v8x16.swizzle into the i8x16
ast node, so that we don't have to create a new v8x16 data type.
Semantically it means the same thing, v8x16 is called that way since it
doesn't treat the underlying data as any particular type, so treating it
as int will work too.

Based on #282 (load and store), review 30a97f8 for changes.

### DIFF
--- a/interpreter/exec/eval_numeric.ml
+++ b/interpreter/exec/eval_numeric.ml
@@ -146,6 +146,7 @@ struct
 
   let binop (op : binop) =
     let f = match op with
+      | I8x16 Swizzle -> SXX.V8x16.swizzle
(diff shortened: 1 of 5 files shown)


binji:
This seems to be based on top of the extract change, is that right?

ngzhian:
Oops forgot to mention, based on #282 (load and store), review 30a97f8 for changes.

ngzhian:
Force push after rebase on merged #282, and took the chance to clean up superfluous commits.

Issue #287, opened 2020-08-04
Text format for extract_lane/replace_lane literals


ngzhian:
For extract_lane (and similarly replace_lane or any instructions that require specifying a lane literal), the test cases allow for specifying it in terms of an `int` or `nat`:

https://github.com/WebAssembly/simd/blob/master/test/core/simd/simd_lane.wast#L889

I'm not super sure if we want this capability. Is such flexibility useful in specifying lane literals? Can we restrict it to nats? So the sign should never be there. It can only be `+` anyway.

@Honry wdyt?

ngzhian:
Also, we seem to have two syntax errors:

- when the lane index for `i8x16.extract_lane_s` is -1, it is a parse error with "malformed lane index"
- when the lane index for `i8x16.extract_lane_s` is 16, it is a validation error with "invalid lane index"

I think we can try to be a bit more consistent here. Maybe make both parse errors, or both validation errors.
Technically -1 is parsable to be a byte of 0xff, it is then during validation that it should raise.

tlively:
I think we should try to be as consistent as possible with the MVP spec here. My reading is that the immediate index on `local.get`, for example, can be any `u32` and can have a sign, so that suggests that the immediate index for `extract_lane` and `replace_lane` should also be allowed to have a sign. On the other hand, the local index can also be an identifier, so perhaps it is different enough that we should not try to match its rules.

binji:
Yeah, I agree that this should be `nat` instead. I brought it up before, but it may have been missed: https://github.com/WebAssembly/simd/issues/181#issuecomment-597386919

binji:
> My reading is that the immediate index on local.get, for example, can be any u32 and can have a sign,

@tlively I don't think that's right, an [index is either an `id` or  `u32`](https://webassembly.github.io/spec/core/text/modules.html#indices). But a [`u32` cannot have a sign](https://webassembly.github.io/spec/core/text/values.html#text-int).

ngzhian:
> Yeah, I agree that this should be nat instead. I brought it up before, but it may have been missed: #181 (comment)

Oh nice, yea I think it was missed. Seems like we want:

- only `nat` as lane indices
- and treatment of malformed v.s. valid w.r.t https://github.com/WebAssembly/simd/issues/181#issuecomment-576003672

> @tlively I don't think that's right, an index is either an id or u32. But a u32 cannot have a sign.

I tried it out with the interpreter too, it's a parser error if we give it `local.get -1`.

[...]

PR #288, opened 2020-08-04
Implement replace_lane for all shapes

ngzhian:
The error messages don't match the tests yet (simd_lane.wast), that
still needs to be worked out in #287.


interpreter/exec/simd.ml
```diff
@@ -186,7 +188,13 @@ struct
     type t = Rep.t
     type lane = Float.t
     let unop f x = Convert.of_shape (List.map f (Convert.to_shape x))
+    let unopi f x = Convert.of_shape (List.mapi f (Convert.to_shape x))
     let binop f x y = Convert.of_shape (List.map2 f (Convert.to_shape x) (Convert.to_shape y))
+
+    let splat x = Convert.of_shape (List.init Convert.num_lanes (fun i -> x))
+    let extract_lane i s = List.nth (Convert.to_shape s) i
+    let replace_lane v i l = unopi (fun j x -> if j = i then l else x) v
```
binji, line 196: maybe use `x` instead of `l`, since `l` is easily confused with `I` or `1` in some fonts

interpreter/exec/eval_numeric.ml
```diff
@@ -264,6 +264,17 @@ struct
...
 
+  let replaceop op v (r : Values.value) =
+    let v128 = of_value 1 v in
+    match op, r with
+    | I8x16 (imm), I32 r -> to_value ((SXX.I8x16.replace_lane v128 imm r))
+    | I16x8 (imm), I32 r -> to_value ((SXX.I16x8.replace_lane v128 imm r))
+    | I32x4 (imm), I32 r -> to_value ((SXX.I32x4.replace_lane v128 imm r))
+    | I64x2 (imm), I64 r -> to_value ((SXX.I64x2.replace_lane v128 imm r))
+    | F32x4 (imm), F32 r -> to_value ((SXX.F32x4.replace_lane v128 imm r))
+    | F64x2 (imm), F64 r -> to_value ((SXX.F64x2.replace_lane v128 imm r))
```
rossberg, line 275: Nit: you don't need the parens around the imm

interpreter/exec/simd.ml
```diff
@@ -186,7 +188,13 @@ struct
     type t = Rep.t
     type lane = Float.t
     let unop f x = Convert.of_shape (List.map f (Convert.to_shape x))
+    let unopi f x = Convert.of_shape (List.mapi f (Convert.to_shape x))
     let binop f x y = Convert.of_shape (List.map2 f (Convert.to_shape x) (Convert.to_shape y))
+
+    let splat x = Convert.of_shape (List.init Convert.num_lanes (fun i -> x))
+    let extract_lane i s = List.nth (Convert.to_shape s) i
+    let replace_lane v lane r = unopi (fun i x -> if i = lane then r else x) v
```
rossberg, line 196: Nit: Can we use similar naming for similar things? For example, v for SIMD value, i,j for lane indices, x,y for lane values?

I'd also suggest tweaking the argument orders of extract and replace to be consistent with each other.

interpreter/exec/simd.ml
```diff
@@ -216,11 +222,15 @@ struct
   struct
     type t = Rep.t
     type lane = Int.t
+    let unop f x = Convert.of_shape (List.map f (Convert.to_shape x))
+    let unopi f x = Convert.of_shape (List.mapi f (Convert.to_shape x))
+    let replace_lane v lane r = unopi (fun i x -> if i = lane then r else x) v
```
rossberg, line 227: Move this down to extract, as in MakeFloat?

interpreter/exec/simd.ml
ngzhian: Yea good point, done!

interpreter/exec/simd.ml
ngzhian: Used i/j and x/y as suggested, thanks!

interpreter/exec/simd.ml
ngzhian: Btw, made them both take the imm as first argument: `extract_lane imm v`, `replace_lane imm v x`

PR #289, opened 2020-08-05
Update lane index errors

### LINKED ISSUE #287: Text format for extract_lane/replace_lane literals

ngzhian:
- lane indices are u8 (nats), negative values are now parse errors ("unexpected token"), as are values with a positive sign `+0x1`
- changed a bunch of "expected i8 literal" to "unexpected token", in order to differentiate these syntax errors we probably need to change the parser to match on a bunch of `EXTRACT_LANE <token we don't care about>`
- some "type mismatch" are now unexpected token

Fixed #287

### DIFF
--- a/test/core/simd/simd_lane.wast
+++ b/test/core/simd/simd_lane.wast
@@ -393,35 +393,38 @@
   (v128.const i64x2 01_234_567_890_123_456_789_0 0x0_1234_5678_90AB_cdef))
   (v128.const i32x4 0xeb1f_0ad2 0xab54_a98c 0x90ab_cdef 0x1234_5678))
 
+;; Syntax errors for negative values
+
+(assert_malformed (module quote "(func (result i32) (i8x16.extract_lane_s  -1 (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)))") "unexpected token")
+(assert_malformed (module quote "(func (result i32) (i8x16.extract_lane_u  -1 (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0)))") "unexpected token")
+(assert_malformed (module quote "(func (result i32) (i16x8.extract_lane_s  -1 (v128.const i16x8 0 0 0 0 0 0 0 0)))") "unexpected token")
+(assert_malformed (module quote "(func (result i32) (i16x8.extract_lane_u  -1 (v128.const i16x8 0 0 0 0 0 0 0 0)))") "unexpected token")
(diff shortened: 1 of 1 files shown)


test/core/simd/simd_lane.wast
```diff
@@ -882,29 +876,29 @@
...
+(assert_malformed (module quote "(func (result i64) (i64x2.extract_lane +1 (v128.const i64x2 0 0)))") "unexpected token")
+(assert_malformed (module quote "(func (result v128) (i8x16.replace_lane +015 (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0) (i32.const 1)))") "unexpected token")
+(assert_malformed (module quote "(func (result v128) (i16x8.replace_lane +0x7 (v128.const i16x8 0 0 0 0 0 0 0 0) (i32.const 1)))") "unexpected token")
+(assert_malformed (module quote "(func (result v128) (i32x4.replace_lane +3 (v128.const i32x4 0 0 0 0) (i32.const 1)))") "unexpected token")
+(assert_malformed (module quote "(func (result v128) (f64x2.replace_lane +0x01 (v128.const f64x2 0 0) (f64.const 1.0)))") "unexpected token")
+
 ;; Lane index literal
 
 (module (func (result i32) (i8x16.extract_lane_s 0x0f (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0))))
-(module (func (result i32) (i8x16.extract_lane_u +0x0f (v128.const i8x16 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0))))
```
binji, line 888: Can you add tests that a leading `+` is malformed too?

ngzhian: These are just above. The diff is slightly confusing, but these deleted heres on the left were changed into `assert_malformed`, and then I moved them above, to a ";; Non-nat lane index", to consolidate all the leading `+` checks.

ngzhian:
Yea a lot of the arithmetic ones are, simd_lane.wast isn't, I looked at the list of files in https://github.com/WebAssembly/simd/blob/master/test/core/simd/meta/README.md.

test/core/simd/simd_lane.wast
binji: Ah, I see. Thanks!

PR #291, opened 2020-08-05
Validate lane index

ngzhian:


interpreter/valid/valid.ml
```diff
@@ -177,15 +177,26 @@ let check_memop (c : context) (memop : 'a memop) get_sz at =
   require (1 lsl memop.align <= size) at
     "alignment must not be larger than natural"
 
-let check_simd_lane_idx op at =
+let check_simd_extract_lane_index op at =
```
ngzhian, line 180: This duplication is slightly annoying, due to the chosen representation for `extractop` and `replaceop`. Is there way to align them? Maybe using a record or something?

ngzhian:
Force push after #288 was squash and merged.

interpreter/valid/valid.ml
binji: Agree, it would be nicer to combine these. I don't know ocaml well enough to say how that could be done, though

rossberg: It should work to make it a higher-order function taking an extractor function:
```
let check_simd_lane_index f op at =
  let max, op' = match op with
    | V128Op.I8x16 op' -> 16, op'
    | ...
  in require (f op' < max) at "invalid lane index"
```
and call it with `check_simd_lane_index Fun.id` vs `check_simd_lane_index snd`.

ngzhian: Great! Made that change, thanks!

interpreter/text/parser.mly
```diff
@@ -65,6 +65,12 @@ let simd_lane_lit shape l at =
   | F32x4 -> LitPat (Values.F32 (F32.of_string l) @@ at) @@ at
   | F64x2 -> LitPat (Values.F64 (F64.of_string l) @@ at) @@ at
 
+let simd_lane_index s at =
+  try
+    let n = int_of_string s in
+    if n >= 0 && n < 256 then n else raise (Failure "")
+  with Failure _ -> error at "malformed lane index"
```
rossberg, line 72: Nit: this could be written more elegantly:
```suggestion
   match int_of_string s with
   | n when 0 <= n && n < 256 -> n
   | _ | exception Failure _ -> error at "malformed lane index"
```

ngzhian: Neat! I didn't know about pattern matching on exception, thanks, please see #301 .

PR #292, opened 2020-08-05
Implement shuffle

ngzhian:
Depends on #288 and #291, review ac03a3d for changes.

### DIFF
--- a/interpreter/text/parser.mly
+++ b/interpreter/text/parser.mly
@@ -71,6 +71,11 @@ let simd_lane_index s at =
     if n >= 0 && n < 256 then n else raise (Failure "")
   with Failure _ -> error at "malformed lane index"
 
+let shuffle_literal ss at =
+  if not (List.length ss = 16) then
+    error at "invalid lane length";
+  List.map (fun s -> simd_lane_index s.it s.at) ss
+
 let nanop f nan =
   let open Source in
   let open Values in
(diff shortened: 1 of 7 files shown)


interpreter/text/parser.mly
```diff
@@ -85,6 +85,17 @@ let nat32 s at =
 let name s at =
   try Utf8.decode s with Utf8.Utf8 -> error at "malformed UTF-8 encoding"
 
+let shuffle_literal ( ss : string Source.phrase list) at : int list =
+  if not (List.length ss = 16) then
+    error at "invalid lane length";
+  let to_index s at =
+    let n = nat s at in
+    if n > 31 then
```
binji, line 93: This should be like extract/replace and allow up to 0xff, right? In general, can this share behavior from those instructions?

ngzhian: Sorry I made this draft because I ran to lunch before rebasing this properly on another PR, I'll fix them up, it should share the same logic as extract replace like you said.

ngzhian: Rebased, and fixed.

binji:
Still lgtm

interpreter/valid/valid.ml
```diff
@@ -165,6 +165,12 @@ let check_unop unop at =
     check_pack sz (Values.type_of unop) at
   | _ -> ()
 
+let check_binop binop at =
+  match binop with
+  | Values.V128 V128Op.(I8x16 (Shuffle imms)) ->
+      if List.for_all ((>) 32) imms then () else error at "invalid lane index"
```
rossberg, line 171:
```suggestion
    if List.exists ((<=) 32) imms then
      error at  "invalid lane index"
```

ngzhian: Thanks, ptal #301 .

PR #298, opened 2020-08-11
Implement some binary<-> text support for SIMD

ngzhian:
This is sufficient to pass test/core/simd/simd_i32x4_arith.wast.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -102,6 +102,7 @@ let vs33 s = I32_convert.wrap_i64 (vsN 33 s)
 let vs64 s = vsN 64 s
 let f32 s = F32.of_bits (u32 s)
 let f64 s = F64.of_bits (u64 s)
+let v128 s = V128.of_bits (get_string (Types.size Types.V128Type) s)
 
 let len32 s =
   let pos = pos s in
@@ -138,6 +139,7 @@ let value_type s =
   | -0x02 -> I64Type
   | -0x03 -> F32Type
   | -0x04 -> F64Type
+  | -0x05 -> V128Type
   | _ -> error s (pos s - 1) "malformed value type"
 
 let elem_type s =
@@ -214,6 +216,16 @@ let math_prefix s =
   | 0x05l -> i64_trunc_sat_f32_u
   | 0x06l -> i64_trunc_sat_f64_s
   | 0x07l -> i64_trunc_sat_f64_u
+  | n -> illegal s pos (I32.to_int_u n)
+
+let simd_prefix s =
+  let pos = pos s in
+  match vu32 s with
+  | 0x0cl -> v128_const (at v128 s)
+  | 0xa1l -> i32x4_neg
+  | 0xael -> i32x4_add
+  | 0xb1l -> i32x4_sub
+  | 0xb5l -> i32x4_mul
   | n -> illegal s pos (I32.to_int_u n)
 
 let rec instr s =
@@ -455,6 +467,7 @@ let rec instr s =
   | 0xc4 -> i64_extend32_s
 
   | 0xfc -> math_prefix s
+  | 0xfd -> simd_prefix s
 
   | b -> illegal s pos b
 
--- a/interpreter/text/arrange.ml
+++ b/interpreter/text/arrange.ml
@@ -286,7 +286,6 @@ let rec instr e =
     | Store op -> storeop op, []
     | MemorySize -> "memory.size", []
     | MemoryGrow -> "memory.grow", []
-    (* Special case for V128, which has a shape after the type *)
     | Const lit -> constop lit ^ " " ^ value lit, []
     | Test op -> testop op, []
     | Compare op -> relop op, []
--- a/interpreter/text/arrange.ml
+++ b/interpreter/text/arrange.ml
@@ -248,7 +248,11 @@ let storeop op =
 
 let var x = nat32 x.it
 let value v = string_of_value v.it
-let constop v = value_type (type_of v.it) ^ ".const"
+let constop v =
+  let typename = value_type (type_of v.it) in
+  match v.it with
+  | V128 _ -> typename ^ ".const i32x4"
+  | _ -> typename ^ ".const"
 
 let block_type = function
   | VarBlockType x -> [Node ("type " ^ var x, [])]
@@ -283,7 +287,6 @@ let rec instr e =
     | MemorySize -> "memory.size", []
     | MemoryGrow -> "memory.grow", []
     (* Special case for V128, which has a shape after the type *)
-    | Const (({it=V128(_); at}) as lit) -> constop lit ^ " i32x4 " ^ value lit, []
     | Const lit -> constop lit ^ " " ^ value lit, []
     | Test op -> testop op, []
     | Compare op -> relop op, []
@@ -428,11 +431,8 @@ let literal mode lit =
 
 (* Converts a literal into a constant instruction. *)
 let constant mode lit =
-  let type_name = string_of_value_type (type_of lit.it) in
   let lit_string = literal mode lit in
-  match lit.it with
-  | Values.V128 v -> Node ("v128.const i32x4 " ^ lit_string, [])
-  | _ -> Node (type_name ^ ".const " ^ lit_string, [])
+  Node (constop lit ^ lit_string, [])
 
 let definition mode x_opt def =
   try
--- a/interpreter/text/arrange.ml
+++ b/interpreter/text/arrange.ml
@@ -250,10 +250,10 @@ let storeop op =
 let var x = nat32 x.it
 let value v = string_of_value v.it
 let constop v =
-  let typename = value_type (type_of v.it) in
-  match v.it with
-  | V128 _ -> typename ^ ".const i32x4 "
-  | _ -> typename ^ ".const "
+  let shape = match v.it with
+    | V128 _ -> "i32x4 "
+    | _ -> ""
+  in value_type (type_of v.it) ^ ".const " ^ shape
 
 let block_type = function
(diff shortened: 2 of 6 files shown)


interpreter/binary/decode.ml
```diff
@@ -216,6 +218,16 @@ let math_prefix s =
   | 0x07l -> i64_trunc_sat_f64_u
   | n -> illegal s pos (I32.to_int_u n)
 
+let simd_prefix s =
+  let pos = pos s in
+  match vu32 s with
+  | 0x0cl -> v128_const (at v128 s)
```
binji, line 224: missing load/store

interpreter/text/arrange.ml
```diff
@@ -405,23 +416,23 @@ let module_ = module_with_var_opt None
...
+  | Values.F32 z -> (choose_mode F32.to_hex_string F32.to_string) z
+  | Values.F64 z -> (choose_mode F64.to_hex_string F64.to_string) z
+  | Values.V128 v -> (choose_mode V128.to_hex_string V128.to_string) v
+
+(* Converts a literal into a constant instruction. *)
+let constant mode lit =
+  let type_name = string_of_value_type (type_of lit.it) in
+  let lit_string = literal mode lit in
+  match lit.it with
+  | Values.V128 v -> Node ("v128.const i32x4 " ^ lit_string, [])
```
binji, line 434: This logic to display " i32x4 " comes up a few times, is it possible to use the SIMD shape representation here? Perhaps as `shape option` to handle non-SIMD types?

interpreter/text/arrange.ml
```diff
@@ -185,38 +185,47 @@ struct
...
+  | V128 o -> ""
+  | _ -> value_type (type_of op) ^ "." in
+  let ops = match op with
   | I32 o -> intop "32" o
   | I64 o -> intop "64" o
   | F32 o -> floatop "32" o
   | F64 o -> floatop "64" o
-  | V128 o -> vectop "128" o
-  )
+  | V128 o ->vectop "128" o
```
binji, line 221: `vectop` should probably be renamed too? Fine for another PR, though

interpreter/binary/decode.ml
ngzhian: If you don't mind, those will come in a future PR, this commit is sufficient to run simd_i32x4_arith.wast, which does not have load/stores.

interpreter/text/arrange.ml
ngzhian: Not sure if we want the "shape" logic inside of all the numeric const nodes, we might have to come up with our own SimdConst node, then it can have a Shape. I can probably try that out as a follow up in order to have more consistent translation between text and binary, and keep the shape information.
Does Wabt do this? (shape option in Const AST nodes)

interpreter/text/arrange.ml
ngzhian: Good point, I can make it in this pr, shouldn't have any conflicts...

interpreter/binary/decode.ml
binji: Yep, that's OK. I was just surprised to see them below in the encoder, but not here.

interpreter/text/arrange.ml
binji: I was thinking that the shape would be a parameter, not necessarily stored in the node. But yeah, doesn't have to be this PR

ngzhian: Ah okay, I don't think it's possible here, because constant here is called by instructions that don't have the knowledge of the shape.
If you look below, `let result mode res =` on line 498, we branch on `SimdResult` or `NumResult`, we only have the shape in the first case, in the second case we end up with a V128 value, and have no information on the shape.

interpreter/binary/decode.ml
ngzhian: Oh yes, I probably messed up `git add -p` and `rebase` somewhere in the flow, thanks!

interpreter/text/arrange.ml
binji: Right, I guess I mean for clarity so there doesn't have to be an explicit match -- so the shape could still be a constant `I32X4`, but the conversion to text is moved to a function.

ngzhian: I moved this check into `constop` above, so this "i32x4" lives in a single place now.

interpreter/text/arrange.ml
```diff
@@ -185,38 +185,47 @@ struct
...
-let oper (intop, floatop, vectop) op =
-  value_type (type_of op) ^ "." ^
-  (match op with
+let oper (intop, floatop, simdop) op =
+  (* v128 operations don't need to be prefixed by the type,
+   * each instruction will specify their prefix (shape).
+   *)
+  let prefix = match op with
+  | V128 o -> ""
+  | _ -> value_type (type_of op) ^ "." in
```
rossberg, line 215:
```suggestion
  let prefix = match op with
    | V128 o -> ""
    | _ -> value_type (type_of op) ^ "."
  in
```

interpreter/text/arrange.ml
```diff
@@ -185,38 +185,47 @@ struct
...
+  let ops = match op with
   | I32 o -> intop "32" o
   | I64 o -> intop "64" o
   | F32 o -> floatop "32" o
   | F64 o -> floatop "64" o
-  | V128 o -> vectop "128" o
-  )
+  | V128 o -> simdop "128" o
```
rossberg, line 221: Nit: indent + 2

interpreter/text/arrange.ml
```diff
@@ -239,7 +248,11 @@ let storeop op =
 
 let var x = nat32 x.it
 let value v = string_of_value v.it
-let constop v = value_type (type_of v.it) ^ ".const"
+let constop v =
+  let typename = value_type (type_of v.it) in
+  match v.it with
+  | V128 _ -> typename ^ ".const i32x4 "
+  | _ -> typename ^ ".const "
```
rossberg, line 255:
```suggestion
  let shape = match v.it with
    | V128 _ -> "i32x4 "
    | _ -> ""
  in value_type (type_of v.it) ^ ".const " ^ shape
```

interpreter/text/arrange.ml
```diff
@@ -405,23 +418,20 @@ let module_ = module_with_var_opt None
...
-    let f = if mode = `Binary then F64.to_hex_string else F64.to_string in
-    Node ("f64.const " ^ f z, [])
-  | Values.V128 v -> 
-    let f = if mode = `Binary then (failwith "unimplemented v128 binary mode") else V128.to_string in
-    Node ("v128.const " ^ f v, [])
+  | Values.I32 i -> (choose_mode I32.to_hex_string I32.to_string_s) i
+  | Values.I64 i -> (choose_mode I64.to_hex_string I64.to_string_s) i
+  | Values.F32 z -> (choose_mode F32.to_hex_string F32.to_string) z
+  | Values.F64 z -> (choose_mode F64.to_hex_string F64.to_string) z
+  | Values.V128 v -> (choose_mode V128.to_hex_string V128.to_string) v
```
rossberg, line 429:
```suggestion
  | Values.I32 i -> choose_mode I32.to_hex_string I32.to_string_s i
  | Values.I64 i -> choose_mode I64.to_hex_string I64.to_string_s i
  | Values.F32 z -> choose_mode F32.to_hex_string F32.to_string z
  | Values.F64 z -> choose_mode F64.to_hex_string F64.to_string z
  | Values.V128 v -> choose_mode V128.to_hex_string V128.to_string v
```

interpreter/text/arrange.ml
```diff
@@ -464,16 +474,29 @@ let nan = function
...
         match nanop.it with
         | Values.I32 _ | Values.I64 _ | Values.V128 _ -> assert false
         | Values.F32 n -> Node ("f32.const " ^ nan n, [])
         | Values.F64 n -> Node ("f64.const " ^ nan n, [])
 
+let result_simd mode res shape pats =
+  (* A different text generation for SIMD, since the literals within
+   * a SimdResult do not need the i32.const instruction *)
+  let num_pat mode res =
+      match res.it with
```
rossberg, line 488: Nit: indent - 2

interpreter/text/arrange.ml
```diff
@@ -464,16 +474,29 @@ let nan = function
...
+      | LitPat lit -> literal mode lit
+      | NanPat {it=Values.F32 n;_}
+      | NanPat {it=Values.F64 n;_} -> nan n
+      | _ -> assert false in
```
rossberg, line 492: Nit: break before `in`

interpreter/text/arrange.ml
```diff
@@ -464,16 +474,29 @@ let nan = function
 
 let result_numpat mode res =
     match res with
-    | LitPat lit -> literal mode lit
+    | LitPat lit -> constant mode lit
     | NanPat nanop ->
         match nanop.it with
```
rossberg, line 479: Nit: old code, but also using inconsistent indent depth

interpreter/text/arrange.ml
```diff
@@ -464,16 +474,29 @@ let nan = function
...
         | Values.F64 n -> Node ("f64.const " ^ nan n, [])
 
+let result_simd mode res shape pats =
+  (* A different text generation for SIMD, since the literals within
+   * a SimdResult do not need the i32.const instruction *)
+  let num_pat mode res =
+      match res.it with
+      | LitPat lit -> literal mode lit
+      | NanPat {it=Values.F32 n;_}
+      | NanPat {it=Values.F64 n;_} -> nan n
```
rossberg, line 491:
```suggestion
      | NanPat {it = Values.F32 n; _}
      | NanPat {it = Values.F64 n; _} -> nan n
```

PR #306, opened 2020-08-15
Implement SIMD load splat and load extend

ngzhian:
Introduce a new SimdLoad AST node to represent SIMD load splat and load
extend. The existing parameters for loadop is not sufficient to
represent all the load splats and extends, as it only has 3 pack sizes,
and 2 sign-extensions, and 1 none case, giving at total of 7 cases:

- sz = (pack_size * extension) option

while we have 6 load extends and 4 load splats (10 in total).

We can add 1 more case to pack_size, Pack64, which will only be used by
SIMD loads, but that still only gives 8 cases. We can also add 1 more
data type to extension, that will give us enough, but it will be useless
for non-v128 loads.

### DIFF
--- a/interpreter/binary/encode.ml
+++ b/interpreter/binary/encode.ml
@@ -200,6 +200,8 @@ let encode m =
       | Load ({ty = V128Type; _} as mo) ->
         simd_op 0x00l; memop mo
 
+      | SimdLoad _ -> failwith "TODO v128 SimdLoad"
+
       | Store ({ty = I32Type; sz = None; _} as mo) -> op 0x36; memop mo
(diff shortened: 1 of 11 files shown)


ngzhian:
This is more for a RFC than a ready for review PR. I wrote down the issues in the description of this PR. I would really like to reuse `loadop`, but it doesn't have enough information right now to discern between all the cases needed for load splats and load extends. So I created a new AST node so I wouldn't have to muck around with the existing `loadop`, adding things that only made sense for `v128`. Suggestions welcome, thanks!

ngzhian:
Closed in favor of #307

PR #307, opened 2020-08-17
Implement SIMD load splat and load extend

ngzhian:
Add more pack types for SIMD load splat and load extends.

This currently uses the old names, after the sync next week we will decide on #322 and I can add a follow-up patch to address the changes.

### DIFF
--- a/interpreter/syntax/types.ml
+++ b/interpreter/syntax/types.ml
@@ -19,8 +19,10 @@ type extern_type =
 type pack_size = Pack8 | Pack16 | Pack32 | Pack64
 type extension = SX | ZX
 type pack_simd =
-  | PackSplat of pack_size
-  | PackExtend of pack_size * extension
+  | PackSplat
+  | Pack8x8 of extension
+  | Pack16x4 of extension
+  | Pack32x2 of extension
 
 (* Attributes *)
 
@@ -34,14 +36,6 @@ let packed_size = function
   | Pack16 -> 2
   | Pack32 -> 4
   | Pack64 -> 8
-
-let packed_simd_size = function
-  | PackSplat sx -> packed_size sx
-  | PackExtend _ -> 8
-
-let pack_size_of_pack_simd = function
-  | PackSplat sx -> sx
-  | PackExtend _ -> Pack64
 
 (* Subtyping *)
 
--- a/interpreter/syntax/types.ml
+++ b/interpreter/syntax/types.ml
@@ -16,7 +16,7 @@ type extern_type =
   | ExternMemoryType of memory_type
   | ExternGlobalType of global_type
 
-type pack_size = Pack8 | Pack16 | Pack32
+type pack_size = Pack8 | Pack16 | Pack32 | Pack64 | Pack8x8 | Pack16x4 | Pack32x2
 type extension = SX | ZX
 
 
@@ -31,7 +31,7 @@ let packed_size = function
   | Pack8 -> 1
   | Pack16 -> 2
   | Pack32 -> 4
-
+  | Pack64 | Pack8x8 | Pack16x4 | Pack32x2 -> 8
 
 (* Subtyping *)
 
--- a/interpreter/syntax/types.ml
+++ b/interpreter/syntax/types.ml
@@ -16,9 +16,11 @@ type extern_type =
   | ExternMemoryType of memory_type
   | ExternGlobalType of global_type
 
-type pack_size = Pack8 | Pack16 | Pack32 | Pack64 | Pack8x8 | Pack16x4 | Pack32x2
+type pack_size = Pack8 | Pack16 | Pack32 | Pack64
 type extension = SX | ZX
-
+type pack_simd =
+  | PackSplat of pack_size
+  | PackExtend of pack_size * extension
 
 (* Attributes *)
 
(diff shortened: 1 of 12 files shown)


ngzhian:
@rossberg this is an alternative to #306. We add more cases to `Types.pack_size` to distinguish between the different kinds of loads. SIMD load extends (load 64 bit representing 8x8, then extend each lane to 16-bits) use `Pack8x8`. These aren't used by the other loads.

I think I like this more, no other types will use the new pack types, but I don't think that's much of a problem - we add an `assert false` to those pattern matches to catch misuse.

interpreter/syntax/types.ml
```diff
@@ -16,7 +16,7 @@ type extern_type =
   | ExternMemoryType of memory_type
   | ExternGlobalType of global_type
 
-type pack_size = Pack8 | Pack16 | Pack32
+type pack_size = Pack8 | Pack16 | Pack32 | Pack64 | Pack8x8 | Pack16x4 | Pack32x2
```
rossberg, line 19: Ah, hm, so the pack_size is supposed to merely encode the size of something and has been reused as that in other contexts such as the new `extend` instructions (see main repo). So I think we need to keep SIMD shapes out of it.

Also, in the reference types proposal, the existing `Load` and `Store` instructions are already restricted to deal with plain numeric types only, since in the presence of reference types they can no longer accept all value types.

To be compatible with these parallel developments, I'd suggest introducing separate `LoadSimd` and `StoreSimd` instructions that carry the following ops:
```
type pack_simd = Pack8x8 | Pack16x4 | Pack32x2 | PackSplat of pack_size
type simd_loadop = (pack_simd * extension) memop
type simd_storeop = [] memop
```
(`[]` is the empty type, such that the memop's `sz` field can only be None for stores.)

interpreter/text/lexer.mll
```diff
@@ -288,6 +288,20 @@ rule token = parse
...
+  { LOAD (fun a o -> (ext s i32x4_load16x4_s i32x4_load16x4_u (opt a 3)) o) }
+  | "i64x2.load32x2_"(sign as s)
+  { LOAD (fun a o -> (ext s i64x2_load32x2_s i64x2_load32x2_u (opt a 3)) o) }
+  | "v8x16.load_splat"
+  { LOAD (fun a o -> (v8x16_load_splat (opt a 0)) o) }
+  | "v16x8.load_splat"
+  { LOAD (fun a o -> (v16x8_load_splat (opt a 1)) o) }
+  | "v32x4.load_splat"
+  { LOAD (fun a o -> (v32x4_load_splat (opt a 2)) o) }
+  | "v64x2.load_splat"
```
rossberg, line 303: Is the plan still to rename these?

ngzhian: Yes that's the plan. We have a SIMD sync next Friday. I was planning to wait for after the sync to do a rename (and also merge outstanding PRs that renamed other places).

interpreter/syntax/types.ml
ngzhian: Got it, in the latest commit I added a new `SimdLoad` ast node, and a `simd_loadop`. Tweaked the type definition slightly from your suggestion here.
I don't think a separate store is required, there is only 1 way of storing a v128, and it can be already be handled correctly.

rossberg: I think a separate Store would be preferable, a) for symmetry, and b) because the pack_size present with numeric stores doesn't make sense for it.

rossberg: (Also, once we merge reference types, the regular Store will only work for num_types anyway.)

ngzhian: Sg, made the change.
I tweak the `simd_loadop` type a little bit as well - since packsize is the size of something loaded from memory, it made the use of it consistent with `loadop`, each `simd_loadop` is a a tuple of `(pack_size * pack_simd)`, the fst encodes the size to be loaded, the snd is a simd specific shape to describe what to do after loading (splat/extend).

I couldn't do `[] memop` as you suggested, it was a syntax error. I searched around and found https://ocaml.org/releases/4.10/htmlman/emptyvariants.html so I created a `type empty = |`. Lmk if I should change it.

test/core/simd/simd_align.wast
```diff
@@ -105,7 +105,7 @@
   (module quote
     "(memory 1) (func (drop (v128.load align=-1 (i32.const 0))))"
   )
-  "alignment must be a power of two"
+  "unknown operator"
```
binji, line 108: Seems like this test should be fixed to use a positive number, non-power-of-two instead.

ngzhian: Ah yes, make sense. Let me review the tests cases to make sure they cover:
- negative -> unknown opreator
- positive non-power-of-two -> alignment error

ngzhian: Hm, we do have test cases for positive non-power-of-two. For v128.load, we test for aligments -1, 0, 7.
I think we can remove the -1 case, or leave it to check for parse errors.

ngzhian: I will leave these negative alignments here, it doesn't looked like we have such test cases anywhere else.

interpreter/syntax/types.ml
rossberg: Ah, I wasn't aware that polymorphic variants can't be empty. Odd. But since we don't actually care it's a variant it's sufficient to define
```
type empty
```
with no RHS.

ngzhian: Good piont, done in #345.

interpreter/text/lexer.mll
ngzhian: Done in #322

PR #311, opened 2020-08-17
Implement SIMD shr_u

ngzhian:
Shifting requires masking i8 and i16 top bits first, since we store them
sign-extended.

Implement `needs_extend` correctly, previously it was implementing the
opposite, but it was used by callers correctly, so change all callers
to use `not needs_extend`.


interpreter/exec/int.ml
```diff
@@ -194,8 +194,15 @@ struct
   let shr_s x y =
     shift Rep.shift_right x y
 
+  (* Check if we are storing smaller ints. *)
+  let needs_extend = (shl one (Rep.of_int (Rep.bitwidth - 1))) <> Rep.min_int
```
rossberg, line 198: Nit:
```suggestion
  let needs_extend = shl one (Rep.of_int (Rep.bitwidth - 1)) <> Rep.min_int
```

interpreter/exec/int.ml
```diff
@@ -194,8 +194,15 @@ struct
...
+
   let shr_u x y =
-    shift Rep.shift_right_logical x y
+    (* If we are storing smaller ints, we need to mask out the high bits. *)
+    let mask = if not needs_extend then Rep.minus_one
+    else Rep.lognot (Rep.shift_left Rep.minus_one (Rep.bitwidth)) in
+    let result = shift Rep.shift_right_logical (Rep.logand x mask) y in
+    result
```
rossberg, line 205:
```suggestion
    let mask =
      if not needs_extend then Rep.minus_one else
      Rep.lognot (Rep.shift_left Rep.minus_one (Rep.bitwidth))
    in shift Rep.shift_right_logical (Rep.logand x mask) y
```

PR #319, opened 2020-08-24
Implement bitmask for SIMD

ngzhian:
i8x16.bitmask, i16x8.bitmask, i32x4.bitmask. Add some tests to
simd_boolean.wast (just a simple correctness check, future work is to
generate the test, move it into a python script.)

This introduces a new AST, SimdBitmask, since it doesn't really fit in
with any existing AST (Test returns a bool).


interpreter/syntax/ast.ml
```diff
@@ -84,6 +84,8 @@ struct
   type replaceop = (int, int, int, int, int, int, int) v128op
   type shift = Shl | ShrS | ShrU
   type shiftop = (shift, shift, shift, shift, shift, shift, shift) v128op
+  type bitmask = Bitmask
+  type bitmaskop = (bitmask, bitmask, bitmask, bitmask, bitmask, bitmask, bitmask) v128op
```
rossberg, line 88: Some food for later thought, but looking at all these extra types, and this one in particular, it seems unnecessary to use v128op in most of these cases. We'd really only need that for heterogeneous cases, where the op set differs across shapes. Otherwise it's merely a convoluted way of expressing the pair (shape, op). Or in the case of bitmask, just shape.

rossberg (approved):

interpreter/syntax/ast.ml
ngzhian: Yea, agree that it's awkward and verbose. I'll think about how to structure it better in the future, thanks for pointing it out.

PR #320, opened 2020-08-24
Fix to_hex_string for I8 and I16

ngzhian:
Print out the full i32 hex string, then extract the lower 2 or 4 chars.


ngzhian:
Yes there is, running `test/core/run.py` on one of the simd wast files (forgot which ones, probably any i8x16 ones) will fail.

Right now we don't run *any* simd test, I will put up a change to change our Makefile/runner to also run those tests.

PR #321, opened 2020-08-24
Rename shuffle and swizzle to i8x16 prefix

### LINKED ISSUE #316: Small rename of shuffle and swizzle
I propose to rename `v8x16.shuffle` and `v8x16.swizzle` to `i8x16.shuffle` and `i8x16.swizzle`, this is a tiny change from `v` prefix, to `i`.

The purpose of the `v` is to indicate that we are not concerned with the underlying type. The rename does not change this fact. The purpose of the rename is to make the spec slightly simpler, and have less special cases.

If #297 happens (likely, given lack of opposition), these 2 instructions will be the only special instructions that begin with `v8x16`. If we can use `i8x16.swizzle` and `i8x16.shuffle`, it will simplify the specification slightly, since there is no need to talk about `v8x16` as a syntax.

Additionally, the rename will be consistent with other instructions, like the horizontal reductions `i32x4.all_true`, which applies the same if the instruction was `f32x4.all_true`.

ngzhian:
v8x16.shuffle and v8x16.swizzle is now i8x16.shuffle and i8x16.swizzle
respectively.

Fixed #316.

### DIFF
--- a/interpreter/binary/decode.ml
+++ b/interpreter/binary/decode.ml
@@ -234,8 +234,8 @@ let simd_prefix s =
   | 0x0al -> let a, o = memop s in v64x2_load_splat a o
   | 0x0bl -> let a, o = memop s in v128_store a o
   | 0x0cl -> v128_const (at v128 s)
-  | 0x0dl -> v8x16_shuffle (List.init 16 (fun x -> u8 s))
-  | 0x0el -> v8x16_swizzle
+  | 0x0dl -> i8x16_shuffle (List.init 16 (fun x -> u8 s))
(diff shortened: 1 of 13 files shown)


ngzhian:
Will hold merging until after next sync (see #323).

ngzhian:
Poll to rename passed, see [WebAssembly SIMD Sync 9/04/2020](https://docs.google.com/document/d/138cF6aOUa9RZC2tOR7AhlIQWdmX5EtpzXRTVDAN3bfo/edit#heading=h.n7rdjckrtpe5) section titled "2. Instruction renaming", merging.

PR #326, opened 2020-08-25
Run SIMD tests

ngzhian:
This enables running all the SIMD tests (in test/core/simd) whenever we
do `make test`.

This should not be submitted yet, it depends on #306 or #307 for the
implementation of load splats and loads extends.

### DIFF
--- a/test/core/run.py
+++ b/test/core/run.py
@@ -22,14 +22,14 @@ parser.add_argument("file", nargs='*')
 arguments = parser.parse_args()
 sys.argv = sys.argv[:1]
 
-mvp_test_files = glob.glob(os.path.join(inputDir, "*.wast"))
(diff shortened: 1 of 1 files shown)


test/core/run.py
```diff
@@ -22,10 +22,14 @@
 arguments = parser.parse_args()
 sys.argv = sys.argv[:1]
 
+mvp_test_files = glob.glob(os.path.join(inputDir, "*.wast"))
```
rossberg, line 25: Nit: "mvp" is a bit misleading here, since the dir also contains post-mvp tests at this point. Maybe rather "main"?

rossberg (approved):

test/core/run.py
ngzhian: Good point, done!

PR #327, opened 2020-08-26
Add validation for SIMD instructions

ngzhian:
Preview at https://www.ngzhian.com/simd/core/valid/instructions.html

There are a bunch of todos:

- where to define v128.const. in syntax/instructions, v128.const is in the SIMD instructions section. in valid/instructions, `t.const` is defined under "numeric instructions", which already covers v128. Do we need to define v128.const again?
- the auxiliary definition of `packed` as a way to map `i8` and `i16` as they appear in the instruction names to `i32` (and the other types to themselves), does that definition look okay?


document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+   single: abstract syntax; instruction
+
+.. _valid-instr-simd:
+
+SIMD Instructions
+~~~~~~~~~~~~~~~~~
+
+
+.. todo::
+   do we need v128.const here again? It's covered above by Numeric Instructions.
```
rossberg, line 168: I'd move v128.const down here.

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+
+.. todo::
+   can we define these instructions that begin with "v128." together with the numerics? The structure, "t.op" matches up. This will require us to specify in the conventions section of syntax/instructions to extend \binop, rather than define a new \vsbinop
```
rossberg, line 171: The separation into sections should follow the structure established in the syntax definition. In general, I think it's preferable to have SIMD stuff in its own section, since it's a rather separate feature, and quite large as well.

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+
+.. _aux-bool:
+
+SIMD instructions can have a prefix to describe the :ref:`shape <syntax-simd-shape>` of the operand. Packed numeric types, |i8| and |i16|, are not :ref:`value type <syntax-valtype>`, we define an auxiliary function to map such packed types into value types:
+
+.. math::
+   \begin{array}{lll@{\qquad}l}
+   \packed(\K{i8x16}) &=& \I32 \\
+   \packed(\K{i16x8}) &=& \I32 \\
+   \packed(t\K{x}N) &=& t \\
```
rossberg, line 181: I think `unpacked` would probably be the more adequate name for this function.

document/core/syntax/instructions.rst
```diff
@@ -344,6 +349,35 @@ Some SIMD instructions have a signedness annotation |sx| which distinguishes whe
...
+     \vibinop ~|~ \vfbinop \\&&|&
+     \virelop ~|~ \vfrelop \\&&|&
+     \viminmaxop ~|~ \visatbinop \\&&|&
+     \SWIZZLE ~|~
+     \NARROW ~|~
+     \MUL ~|~
+     \AVGRU ~|~
+     \TRUNC ~|~
+     \CONVERT \\
+   \production{ternary operator} & \vsternop &::=& \SHUFFLE \\
```
rossberg, line 377: Why is this vsternop while the others are vunop, vbinop (without the s)? Are vsunop, vsbinop defined somewhere else? (They appear in validation rules.)

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+
+:math:`t\K{.}\vsunop`
+.....................
+
+* The instruction is valid with type :math:`[t] \to [t]`.
+
+.. math::
+   \frac{
+   }{
+     C \vdashinstr t\K{.}\vsunop : [t] \to [t]
```
rossberg, line 195: `t` is always `v128` in these rules, right? So there is no point in abstracting it away.

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+.. math::
+   \frac{
+   }{
+     C \vdashinstr t\K{.}\vsternop : [t~t~t] \to [t]
+   }
+
+
+.. _valid-vunop:
+
+:math:`vxx\K{.}\vunop`
```
rossberg, line 229: This should be `\X{vxx}` (here and below). In fact, you'll want to introduce a macro for it, so that it's linked to the definition, like other grammar symbols.

I'd also consider renaming these grammar symbols:
ixx -> ishape
fxx -> fshape
vxx -> shape

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+* The instruction is valid with type :math:`[\V128] \to [\I32]`.
+
+.. math::
+   \frac{
+   }{
+     C \vdashinstr vxx\K{.}\vtestop : [\V128] \to [\I32]
+   }
+
+.. todo::
+   is this definition of splat okay? Or should we split out i8x16 and i16x8, like below.
```
rossberg, line 283: A single rule like this is definitely better.

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+:math:`vxx\K{.}\SPLAT`
+..........................
+
+* Let :math:`t` be :math:`\packed(vxx)`.
+* The instruction is valid with type :math:`[t] \to [\V128]`.
+
+.. math::
+   \frac{
+   }{
+     C \vdashinstr vxx\K{.}\SPLAT : [t] \to [\V128]
```
rossberg, line 297: Need to replace `t` with `\packed(\vxx)`.

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+
+.. math::
+   \frac{
+   }{
+     C \vdashinstr t\K{x}N\K{.}\SPLAT : [t] \to [\V128]
+   }
+
+
+.. todo::
+   develop some sort of shorthand for this i8x16 and i16x8 repetition?
```
rossberg, line 344: Yes, please. :) I think you can just use `\vxx` -- it doesn't matter that this covers additional cases, the grammar definition already narrows the ones that are syntactically allowed.

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+
+:math:`ixx\K{.}\BITMASK`
+............................
+
+* The instruction is valid with type :math:`[\V128] \to [\I32]`.
+
+.. math::
+   \frac{
+   }{
+     C \vdashinstr t\K{x}N\K{.}\BITMASK : [\V128] \to [\I32]
```
rossberg, line 442:
```suggestion
     C \vdashinstr \ixx\K{.}\BITMASK : [\V128] \to [\I32]
```

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,325 @@ Numeric Instructions
...
+
+* For all :math:`\laneidx_i`, in :math:`\laneidx^{16}`, :math:`\laneidx_i` must be smaller than :math:`N`.
+
+* The instruction is valid with type :math:`[\V128~\V128] \to [\V128]`.
+
+.. math::
+   \frac{
+     (\laneidx < N)^{16}
+   }{
+     C \vdashinstr \K{v8x16.}\SHUFFLE : [\V128~\V128] \to [\V128]
```
rossberg, line 472: Immediate is missing here.

rossberg:
Btw, as a heads-up, I'm vanishing off the grid for the next 2 weeks, so this has to be my last review until after.

ngzhian:
No problem, have a great time!

document/core/syntax/instructions.rst
ngzhian: My horrible naming convention is:
- vs* is all the shape-agnostic instructions, like v128.and
- vi* is all the int-shape instructions, like i32x4.add
- vf* is all the float-shape instructions, like f32x4.sqrt
- v* is union of vi* and vf*

vsunop and vsbinop are defined in the production for instructions above, they aren't grouped as part of this "Conventions" section.

I decided to not group them together becase the validation instructions have 2 different matches, `t.op` which is for `vs*` family of instructions, and `vxx.op`, for the vi*, vf*, v* family.(since `t` matches `v128` and `vxx` matches all the shapes.)

document/core/valid/instructions.rst
ngzhian: Sg, I'll keep them separate, thanks!

document/core/valid/instructions.rst
ngzhian: Good idea, made this change (and also added macro to link to the production rules which define them).

document/core/syntax/instructions.rst
rossberg: Hm, but `vsternop` is already defined earlier as
```
vsternop ::= bitselect
```
So there's some clash here. Shouldn't the one defined here rather be named `vternop`?

Also, wouldn't it be more consistent if `vtestop` was `vitestop`?

document/core/syntax/instructions.rst
```diff
@@ -196,12 +200,12 @@ SIMD instructions provide basic operations over :ref:`values <syntax-value>` of
...
-   \production{fshape} & \X{fxx} &::=&
+   \production{fshape} & \fshape &::=&
      \K{f32x4} ~|~ \K{f64x2} \\
-   \production{vshape} & \X{vxx} &::=&
-     \X{ixx} ~|~ \X{fxx} \\
+   \production{shape} & \shape &::=&
+     \ishape ~|~ \fshape \\
    \production{lane index} & \laneidx &::=& \byte \\
    \production{instruction} & \instr &::=&
      \dots \\&&|&
```
rossberg, line 211: [This comment applies to the line below, but I can't put it there.]
Where is `\vX{nnn}` defined? Shouldn't this be `\i128`?

document/core/syntax/instructions.rst
```diff
@@ -196,12 +200,12 @@ SIMD instructions provide basic operations over :ref:`values <syntax-value>` of
...
+   \production{ishape} & \ishape &::=&
      \K{i8x16} ~|~ \K{i16x8} ~|~ \K{i32x4} ~|~ \K{i64x2} \\
-   \production{fshape} & \X{fxx} &::=&
+   \production{fshape} & \fshape &::=&
      \K{f32x4} ~|~ \K{f64x2} \\
-   \production{vshape} & \X{vxx} &::=&
-     \X{ixx} ~|~ \X{fxx} \\
+   \production{shape} & \shape &::=&
+     \ishape ~|~ \fshape \\
    \production{lane index} & \laneidx &::=& \byte \\
```
rossberg, line 209: Just noted this. This should rather be `\i8`, since `byte` is only intended for uninterpreted binary data. In fact, it would make more sense to make it `\i4`, such that the value range is maintained by construction (with `\Bu4` in the binary format, such that bad indices become a decoding error).

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,263 @@ Numeric Instructions
...
+.. math::
+   \frac{
+   }{
+     C \vdashinstr \V128\K{.}\vsternop : [\V128~\V128~\V128] \to [\V128]
+   }
+
+
+.. _valid-simd-shuffle:
+
+:math:`\K{v8x16.}\SHUFFLE~\laneidx^{16}`
```
rossberg, line 246: Has this been renamed?

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,263 @@ Numeric Instructions
...
+
+* For all :math:`\laneidx_i`, in :math:`\laneidx^{16}`, :math:`\laneidx_i` must be smaller than :math:`32`.
+
+* The instruction is valid with type :math:`[\V128~\V128] \to [\V128]`.
+
+.. math::
+   \frac{
+     (\laneidx < 32)^{16}
+   }{
+     C \vdashinstr \K{v8x16.}\SHUFFLE~\laneidx^{16} : [\V128~\V128] \to [\V128]
```
rossberg, line 257: Same here.

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,263 @@ Numeric Instructions
...
+:math:`\K{v8x16.}\SHUFFLE~\laneidx^{16}`
+........................................
+
+* For all :math:`\laneidx_i`, in :math:`\laneidx^{16}`, :math:`\laneidx_i` must be smaller than :math:`32`.
+
+* The instruction is valid with type :math:`[\V128~\V128] \to [\V128]`.
+
+.. math::
+   \frac{
+     (\laneidx < 32)^{16}
```
rossberg, line 255: Why 32, shouldn't it be 16? (See also my suggestion to make laneidx an i4, which would make this condition obsolete.)

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,263 @@ Numeric Instructions
...
+     \laneidx < \lanes(\shape)
+   }{
+     C \vdashinstr t\K{x}N\K{.}\EXTRACTLANE~\laneidx : [\V128] \to [\unpacked(\shape)]
+   }
+
+
+.. _valid-simd-extract-lane-sx:
+
+:math:`\shape\K{.}\EXTRACTLANE\K{\_}\sx~\laneidx`
+..................................................
```
rossberg, line 297:
```suggestion
.................................................
```

document/core/valid/instructions.rst
```diff
@@ -154,6 +154,263 @@ Numeric Instructions
...
+   \frac{
+     \laneidx < \lanes(\shape)
+   }{
+     C \vdashinstr t\K{x}N\K{.}\EXTRACTLANE~\laneidx : [\V128] \to [\unpacked(\shape)]
+   }
+
+
+.. _valid-simd-extract-lane-sx:
+
+:math:`\shape\K{.}\EXTRACTLANE\K{\_}\sx~\laneidx`
```
rossberg, line 296: Can this be merged with the previous rule by writing `\sx^?` and using `unpacked` to determine the result type?

document/core/valid/instructions.rst
```diff
@@ -389,6 +646,48 @@ Memory Instructions
...
+
+* The alignment :math:`2^{\memarg.\ALIGN}` must not be larger than :math:`8`.
+
+* Then the instruction is valid with type :math:`[\I32] \to [\V128]`.
+
+.. math::
+   \frac{
+     C.\CMEMS[0] = \memtype
+     \qquad
+     2^{\memarg.\ALIGN} \leq 8
```
rossberg, line 664: You could make this rule a bit more symmetric with others by saying `\leq M \cdot N / 8`.

document/core/syntax/instructions.rst
ngzhian: Yea you're right, the `vsternop ::= shuffle` here is wrong, shuffle is not a ternop. I'll remove this.
Also will make the change from `vtestop` to `vitestop`.

document/core/syntax/instructions.rst
ngzhian: Ah yes, I think vnnn was defined previously and then remove, `\i128` is better here. Since we won't have any other type of values.

document/core/syntax/instructions.rst
ngzhian: I fixed this in #346 based on your comment in another PR, so I'll leave this here, and rebase after that is merged.

document/core/valid/instructions.rst
ngzhian: Yup, fixed, thanks!

document/core/valid/instructions.rst
ngzhian: For shuffle, indices can go to 31, since `[0,15]` select from the first operand, and `[16,31]` select from the second operand. (see https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#shuffling-using-immediate-indices)

ngzhian: Because of this, I changed #346 to be a `u8` instead of `u4`. As defined in the semantics, `u8` lane indices are accepted (no decoding error), but will be rejected by validation.
(Text format might need some tweaking, since those rejects lane indices that are out of range at the parsing stage. I'll look at those separately)

document/core/valid/instructions.rst
ngzhian: It will be `\leq N \cdot M /8`, since `N` in this case is the number of lanes, and `M` is the size of each lane. I can swap M and N in the instruction, then it will work. Wdyt?

document/core/valid/instructions.rst
rossberg: Hm, okay, I wonder if that's worth keeping that separate? It's a strange form of index that I'd argue isn't really a proper laneidx. OTOH, maybe that's not worth the trouble.

document/core/valid/instructions.rst
rossberg: I don't mind switching, but isn't A*B/C the same as B*A/C?

document/core/syntax/instructions.rst
```diff
@@ -209,7 +209,7 @@ SIMD instructions provide basic operations over :ref:`values <syntax-value>` of
...
    \production{instruction} & \instr &::=&
      \dots \\&&|&
-     \K{v128.}\CONST~\xref{syntax/values}{syntax-simd}{\vX{\X{nnn}}} \\&&|&
+     \K{v128.}\CONST~\i128 \\&&|&
```
rossberg, line 212: Should this be \VCONST?

document/core/valid/instructions.rst
```diff
@@ -383,17 +367,17 @@ We also define an auxiliary function to get number of packed numeric types in a
    }
 
 
-.. _valid-vtestop:
+.. _valid-vitestop:
 
-:math:`\shape\K{.}\vtestop`
+:math:`\shape\K{.}\vitestop`
 ...........................
```
rossberg, line 373: Sphinx is pedantic.
```suggestion
............................
```

document/core/syntax/instructions.rst
ngzhian: Fixed.

document/core/valid/instructions.rst
ngzhian: Sorry, I left out the parentheses there, what i meant was: `\leq N \cdot (m/8)`.

ngzhian: I swapped the N and M, so it is LoadNxM, it looks a bit more symmetric as N has to do with the size of the type. I also used `N/8 \cdot M`.

document/core/valid/instructions.rst
ngzhian: Say we make `\laneidx ::= u4` in the syntax. In the text format, if we specify that lane indices must be `u4` as well, it means that values like 255 will not be accepted in the text format, which I guess is a parse error?

Right now the implementation will accept 255, and reject it during validation. Which requires that in the text format, lane indices be `u8`. But then we won't be able to create the syntax, since it's a `u4`.

Maybe I'm confused, but for the above reason, I think keeping them the same (`u8`) will be better - we can keep the text -> syntax as currently implemented, and instructions can reject out of bounds indices as necessary during validation.

rossberg: Well, yes, if we defined it to be u4, then larger values would have to be parse errors, and the reference implementation would have to be adjusted accordingly. I don't see a problem with that, but probably not worth it, at least for now.

rossberg (approved):

PR #333, opened 2020-08-28
Specify binary format for SIMD

ngzhian:


document/core/binary/instructions.rst
```diff
@@ -403,6 +403,316 @@ whereas the actual opcode is encoded by a variable-length :ref:`unsigned integer
...
+     \hex{FD}~~10{:}\Bu32~~m{:}\Bmemarg &\Rightarrow& \I64X2.\LOAD\K{\_splat}~m \\ &&|&
+     \hex{FD}~~11{:}\Bu32~~m{:}\Bmemarg &\Rightarrow& \V128.\STORE~m \\
+   \end{array}
+
+The |VCONST| instruction is followed by 16 immediate bytes, which are converted into a |i128| in |littleendian| byte order:
+
+.. math::
+   \begin{array}{llclll}
+   \production{instruction} & \Binstr &::=& \dots \\&&|&
+     \hex{FD}~~12{:}\Bu32~~(b{:}\Bbyte)^{16} &\Rightarrow& \V128.\VCONST~
```
binji, line 440: I wonder if this should use a function defined in `values.rst`, like the others? Maybe it's better to keep all SIMD stuff here.

document/core/binary/instructions.rst
```diff
@@ -403,6 +403,316 @@ whereas the actual opcode is encoded by a variable-length :ref:`unsigned integer
...
+     \hex{FD}~~33{:}\Bu32~~b{:}\Bbyte &\Rightarrow& \F64X2.\EXTRACTLANE~b \\ &&|&
+     \hex{FD}~~34{:}\Bu32~~b{:}\Bbyte &\Rightarrow& \F64X2.\REPLACELANE~b \\
+   \end{array}
+
+All other SIMD instructions are plain opcodes without any immediates.
+
+.. math::
+   \begin{array}{llclll}
+   \production{instruction} & \Binstr &::=& \dots && \phantom{simdhaslongerinstructionnames} \\&&|&
+     \hex{FD}~~14{:}\Bu32 &\Rightarrow& \I8X16.\SWIZZLE \\ &&|&
```
binji, line 480: The v1 documentation is able to introduce the instructions by their immediates, but also keep them in order. I'm wondering if it will be better to keep the instructions in numeric order, even if it means that we break that grouping here.

document/core/binary/instructions.rst
ngzhian: I think this should stay here, because this is about how the instruction treats its immediates. `values.rst` describes how integers are encoded in binary, and already covers a 128-bit integer. It might be confusing to describe how v128.const treats the bytes in the value section.

document/core/binary/instructions.rst
ngzhian: Yea, I'm not sure how to introduce each section if we were to keep the numeric order, it might read strange.

binji: Never mind about this, we introduce them out of order in other proposals too.

ngzhian: Sg, thanks for checking. At least in the index it will all be in-order.

ngzhian:
cc @rossberg, i'll merge this in first, have a text format change that depends on this, please leave comments when you get to this and i'll be happy to make edits, thanks!

document/core/binary/instructions.rst
```diff
@@ -403,6 +403,316 @@ whereas the actual opcode is encoded by a variable-length :ref:`unsigned integer
...
 
+.. index:: simd instruction
+   pair: binary format; instruction
+.. _binary-instr-simd:
+
+SIMD Instructions
+~~~~~~~~~~~~~~~~~~~~
+
+All variants of :ref:`SIMD instructions <syntax-instr-simd>` are represented by separate byte codes.
+The all have a one byte prefix, whereas the actual opcode is encoded by a variable-length :ref:`unsigned integer <binary-uint>`.
```
rossberg, line 414: s/The/They

document/core/binary/instructions.rst
```diff
@@ -403,6 +403,316 @@ whereas the actual opcode is encoded by a variable-length :ref:`unsigned integer
...
+.. math::
+   \begin{array}{llclll}
+   \production{instruction} & \Binstr &::=& \dots \\&&|&
+     \hex{FD}~~12{:}\Bu32~~(b{:}\Bbyte)^{16} &\Rightarrow& \V128.\VCONST~
+     bytes_{\K{i128}}^{-1}(b_{0}~\dots~b_{15}) \\
+   \end{array}
+
+.. _binary-vternop:
+
+The |SHUFFLE| instruction is also followed by 16 immediate bytes:
```
rossberg, line 446: As mentioned in another review, I think laneidx would more appropriately be encoded as u4, not byte. In either case, it would be nice to factor out their encoding into a Blaneidx production.

document/core/binary/instructions.rst
ngzhian: #346

document/core/binary/instructions.rst
ngzhian: Agree, factored this out in #346, ptal

PR #336, opened 2020-09-01
Specify text format for SIMD

ngzhian:
Preview this at https://www.ngzhian.com/simd/core/text/instructions.html#simd-instructions

### DIFF
--- a/document/core/exec/numerics.rst
+++ b/document/core/exec/numerics.rst
@@ -126,6 +126,7 @@ When a number is stored into :ref:`memory <syntax-mem>`, it is converted into a 
 
 .. math::
    \begin{array}{lll@{\qquad}l}
+   \bytes_t(i^\ast) &=& (\bytes_t(i))^\ast \\[1ex]
    \bytes_t(i) &=& \littleendian(\bits_t(i)) \\[1ex]
    \littleendian(\epsilon) &=& \epsilon \\
    \littleendian(d^8~{d'}^\ast~) &=& \littleendian({d'}^\ast)~\ibits_8^{-1}(d^8) \\
--- a/document/core/text/instructions.rst
+++ b/document/core/text/instructions.rst
@@ -447,6 +447,304 @@ Numeric Instructions
    \end{array}
 
 
+.. index:: simd instruction
+   pair: text format; instruction
+.. _text-instr-simd:
+
+SIMD Instructions
+~~~~~~~~~~~~~~~~~~~~
+
+SIMD memory instructions have optional offset and alignment immediates, the rules are the same as the :ref:`memory instructions <text-memarg>`.
+
+.. math::
+   \begin{array}{llclll}
+   \production{instruction} & \Tplaininstr_I &::=& \dots \phantom{averylonginstructionnameforsimdtext} && \phantom{simdhasreallylonginstructionnames} \\ &&|&
+     \text{v128.load}~~m{:}\Tmemarg_{16} &\Rightarrow& \V128.\LOAD~m \\ &&|&
+     \text{i16x8.load8x8\_s}~~m{:}\Tmemarg_8 &\Rightarrow& \I16X8.\LOAD\K{8x8\_s}~m \\ &&|&
+     \text{i16x8.load8x8\_u}~~m{:}\Tmemarg_8 &\Rightarrow& \I16X8.\LOAD\K{8x8\_u}~m \\ &&|&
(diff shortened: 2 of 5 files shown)


document/core/exec/numerics.rst
```diff
@@ -126,6 +126,7 @@ When a number is stored into :ref:`memory <syntax-mem>`, it is converted into a
 
 .. math::
    \begin{array}{lll@{\qquad}l}
+   \bytes_t(i^\ast) &=& (\bytes_t(i))^\ast \\[1ex]
```
binji, line 129: Is this necessary, or can you group the `bytes_` cases below using the format of the right-hand side? It seems like either case will be about as clear.

ngzhian: I'm not sure what you mean by "group the bytes_ cases below using the format of the right-hand side?", can you give an example?
I wanted this case to make `bytes` take multiple numbers as arguments, and then convert them each into their bytes format.

binji: I just meant, can you "inline" this definition into the uses below. But honestly, I don't think it's a big deal either way.

ngzhian: I will keep this here for now, I think it is useful for exec semantics too (future patch), if it ends up being unused otherwise, I can inline in. Thanks!

rossberg: Yeah, I don't understand why this definition is needed either instead of inlining the rhs. It doesn't shorten or abstract anything, and for readability you generally want to avoid the cognitive load of extra levels of definitional indirection when they are unnecessary.

document/core/text/instructions.rst
```diff
@@ -447,6 +447,304 @@ Numeric Instructions
...
+   \begin{array}{llclll}
+   \phantom{\production{instruction}} & \phantom{\Tplaininstr_I} &\phantom{::=}& \phantom{averylonginstructionnameforsimdtext} && \phantom{simdhasreallylonginstructionnames} \\[-2ex] &&|&
+     \text{v128.const}~~\text{i8x16}~~(n{:}\Ti8)^{16} &\Rightarrow& \V128.\VCONST~\bytes_{i128}^{-1}(\bytes_{i8}(n^{16})) \\ &&|&
+     \text{v128.const}~~\text{i16x8}~~(n{:}\Ti16)^{8} &\Rightarrow& \V128.\VCONST~\bytes_{i128}^{-1}(\bytes_{i16}(n^8)) \\ &&|&
+     \text{v128.const}~~\text{i32x4}~~(n{:}\Ti32)^{4} &\Rightarrow& \V128.\VCONST~\bytes_{i128}^{-1}(\bytes_{i32}(n^4)) \\ &&|&
+     \text{v128.const}~~\text{i64x2}~~(n{:}\Ti64)^{2} &\Rightarrow& \V128.\VCONST~\bytes_{i128}^{-1}(\bytes_{i64}(n^2)) \\ &&|&
+     \text{v128.const}~~\text{f32x4}~~(z{:}\Tf32)^{4} &\Rightarrow& \V128.\VCONST~\bytes_{i128}^{-1}(\bytes_{f32}(z^4)) \\ &&|&
+     \text{v128.const}~~\text{f64x2}~~(z{:}\Tf64)^{2} &\Rightarrow& \V128.\VCONST~\bytes_{i128}^{-1}(\bytes_{f64}(z^2)) \\ &&|&
+     \text{i8x16.shuffle}~~(laneidx{:}\Tu8)^{16} &\Rightarrow& \I8X16.\SHUFFLE~laneidx^{16} \\ &&|&
+     \text{i8x16.swizzle} &\Rightarrow& \I8X16.\SWIZZLE\\
```
rossberg, line 488: Nit: move these last two instructions to the next block of rules.

document/core/exec/numerics.rst
ngzhian: I used this definition in `bytes_i8(n^16)`, without this definition it would be `(bytes_i8(n))^16)` which I originally saw as a bit confusing, since it could be 16 copies of `bytes_i8(n)`.
I.e., given `n^16`, there are two interpretations:
1. the value that `n` represents, repeated 16 times
2. 16 non-terminal `n`, where each `n` can be different.
I guess this two cases can be clear from the context (LHS).
I'll inline this, thanks! #347

document/core/text/instructions.rst
ngzhian: #347

PR #340, opened 2020-09-10
Execution semantics

ngzhian:
Preview at https://www.ngzhian.com/simd/core/exec/instructions.html#simd-instructions.


ngzhian:
@binji I'm still working on shuffle, swizzle, load splat and load extends, the rest of the ops are mostly done, definitely still needs cleanup. Hoping you can take a preliminary look.

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,494 @@ Where the underlying operators are non-deterministic, because they may return on
...
 
+.. index:: SIMD instruction
+   pair: execution; instruction
+   single: abstract syntax; instruction
+.. _exec-instr-simd:
+
+SIMD Instructions
+~~~~~~~~~~~~~~~~~~~~
+
+SIMD instructions are defined in terms of generic numeric operators applied lane-wise based on the :ref:`shape <syntax-simd-shape>`.
```
ngzhian, line 193: 2. This definition of simd operations rely on the tweak to numeric operations, that lifts numeric operations to sequences of input.

document/core/exec/numerics.rst
```diff
@@ -38,6 +38,23 @@ When several of these placeholders occur in a single clause, then they must be r
       \fcopysign_N(- p_1, + p_2) &=& + p_1 \\
       \end{array}
 
+Numeric operators are lifted to input sequences by applying the operator element-wise, returning a sequence of results. When there are multiple inputs, they must be of equal length.
```
ngzhian, line 41: 1. This tweak to numeric operations is useful for making the definition of simd operations slightly easier, since each simd operation. Each simd operation can simply pass a sequence of values to the numeric ops.

document/core/exec/numerics.rst
```diff
@@ -135,6 +152,143 @@ When a number is stored into :ref:`memory <syntax-mem>`, it is converted into a
...
+.. _aux-simdof:
+
+SIMD
+....
+
+SIMD values have the same underlying representation as an |i128|. They can also be interpreted as numeric values packed into a |V128| with a particular |shape|.
+
+.. math::
+   \begin{array}{lll@{\qquad}l}
+   \simdto_{t\K{x}N}(c) &=&
```
ngzhian, line 166: 3. This is the key aux function used for almost all simd operations, it converts a i128 into a sequence of numeric values. This is generalized over all shapes (doesn't care about i32x4 or f32x4, they are both converted to a sequence of length 4).

document/core/exec/numerics.rst
```diff
@@ -135,6 +152,143 @@ When a number is stored into :ref:`memory <syntax-mem>`, it is converted into a
...
+     c_0~\dots~c_N \\
+     && (\where
+       w = |t| / 8 \wedge
+       b^* = bytes_{\i128}(i) \\ &&
+       \wedge c_j = bytes_{t}^{-1}(b^*[j*w \slice ((j+1)*w)])
+       )
+   \end{array}
+
+..
+   bits_{i8}^{-1}(b^*[0 \slice 8])~
```
ngzhian, line 176: 4. This section is all my previous attempts to define the \simdto function, can be ignored. I kept them around for reference and will remove them before merge (if we are happy with the current definition above).

document/core/exec/numerics.rst
```diff
@@ -303,6 +457,18 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
    it holds that :math:`i_1 = i_2\cdot\idivs(i_1, i_2) + \irems(i_1, i_2)`.
 
 
+.. _op-inot:
```
ngzhian, line 460: 5. A couple of new numeric ops are defined that are pretty straightforward. We need this as SIMD has some numeric instructions, like or and andnot, that are not yet defined.

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,494 @@ Where the underlying operators are non-deterministic, because they may return on
...
+   invokes :math:`\ADD_{\K{i32x4}}(i_1, i_2)`, which maps to
+   :math:`\simdof_{\K{i32x4}}(\ADD_{\I32}(i_1^+, i_2^+))`,
+   where :math:`i_1^+` and :math:`i_2^+` are sequences resulting from invoking
+   :math:`\simdto_{\K{i32x4}}(i_1)` and :math:`\simdto_{\K{i32x4}}(i_2)`
+   respectively.
+
+
+..
+  .. math::
+     \begin{array}{lll@{\qquad}l}
```
ngzhian, line 217: 6. This commented section is how I defined simd operations *if* we did not lift numeric operations to work on input sequences. I think it is harder to read, many \dots, and more ambiguous.

ngzhian (commented):
I left some review notes (numbered based on suggested ordering to look at) to highlight bigger changes or extensions to existing definition.

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+:math:`\shape\K{.}\REPLACELANE~\laneidx`
+........................................
+
+1. Assert: due to :ref:`validation <valid-simd-replace-lane>`, :math:`\laneidx < \lanes(\shape)`.
+
+2. Let :math:`t_1` be the type :math:`\unpacked(\shape)`.
+
+3. Assert: due to :ref:`validation <valid-simd-replace-lane>`, a value of :ref:`value type <syntax-valtype>` :math:`t_1` is on the top of the stack.
+
+4. Pop the value :math:`t_1.\CONST~v_1` from the stack.
```
binji, line 464: why `v_1` instead of `c_1`?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+.. _exec-simd-vbinnop:
+
+:math:`\shape\K{.}\vbinop`
+..........................
+
+1. Assert: due to :ref:`validation <valid-vbinop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+
+2. Pop the value :math:`\V128.\CONST~c_2` from the stack.
+
+4. Pop the value :math:`\V128.\CONST~c_1` from the stack.
```
binji, line 521: typo `3.`

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+   (\V128\K{.}\CONST~c_1)~(\V128\K{.}\CONST~c_2)~\V128\K{.}\vbinop &\stepto& (\V128\K{.}\CONST~c)
+     & (\iff c \in \vbinop_{\shape}(c_1, c_2)) \\
+   (\V128\K{.}\CONST~c_1)~(\V128\K{.}\CONST~c_2)~\V128\K{.}\vbinop &\stepto& \TRAP
+     & (\iff \vbinop_{\shape}(c_1, c_2) = \{\})
+   \end{array}
+
+
+.. _exec-simd-vshiftop:
+
+:math:`t\K{x}N\K{.}\vshiftop`
```
binji, line 544: why not `shape` here?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+1. Assert: due to :ref:`validation <valid-vtestop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+
+2. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+3. Let :math:`n` be the integer :math:`\lanes(\shape)`.
+
+4. Let :math:`i_1^n` be the sequence :math:`\simdto_{\shape}(c_1)`
+
+5. Let :math:`i` be the result of computing :math:`\bool((i_1 != 0)^n)`.
```
binji, line 589: use `\neq` instead of `!=`

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+.. _exec-simd-anytrue:
+
+:math:`\shape\K{.}\ANYTRUE`
+...........................
+
+1. Assert: due to :ref:`validation <valid-vtestop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+
+2. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+3. Let :math:`i` be the result of computing :math:`\ine_{\K{128}}(c_1, 0)`.
```
binji, line 617: missing closing `)`

document/core/exec/instructions.rst
binji: Also I'm not sure how to understand the logic here -- we convert each value in the sequence to a true or false value, then apply bool to convert to a sequence of 0 and 1, right? When is it collapsed to a single integer?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+1. Assert: due to :ref:`validation <valid-vtestop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+
+2. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+3. Let :math:`i_1^N` be the sequence :math:`\simdto_{t\K{x}N}(c)`.
+
+4. Let :math:`B` be the :ref:`bit width <syntax-valtype>` :math:`|t|` of :ref:`value type <syntax-valtype>` :math:`t`.
+
+5. Let :math:`i_2^N` sequence as a result of computing :math:`\ilts_{B}(i_1^N, 0^N)`.
```
binji, line 641: s/sequence/be the sequence/

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+:math:`\shape\K{.}\SPLAT`
+.........................
+
+1. Let :math:`t` be the type :math:`\unpacked(\shape)`.
+
+2. Assert: due to :ref:`validation <valid-simd-splat>`, a value of :ref:`value type <syntax-valtype>` :math:`t` is on the top of the stack.
+
+3. Pop the value :math:`t.\CONST~c_1` from the stack.
+
+4. Let :math:`n` be the integer :math:`\lanes(\shape)`.
```
binji, line 399: below `N` is often used for the lane count, maybe use that instead for consistency?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+2. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+3. Let :math:`i_1^N` be the sequence :math:`\simdto_{t\K{x}N}(c)`.
+
+4. Let :math:`B` be the :ref:`bit width <syntax-valtype>` :math:`|t|` of :ref:`value type <syntax-valtype>` :math:`t`.
+
+5. Let :math:`i_2^N` sequence as a result of computing :math:`\ilts_{B}(i_1^N, 0^N)`.
+
+6. Let :math:`c` be the integer :math:`ibits_{32}^{-1}(i_2^N)`.
```
binji, line 643: link to ibits

binji, line 643: This feels a bit odd since N < 32, maybe pad with 0 bits?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+1. Assert: due to :ref:`validation <valid-vtestop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+
+2. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+3. Let :math:`i_1^N` be the sequence :math:`\simdto_{t_1\K{x}M}(c_1)[0 \slice N]`.
+
+4. Let :math:`c` be the result of computing :math:`\simdof_{t_2\K{x}N}((\extend^{\sx}_{t_1,t_2}(i_1))^N)`
+
+6. Push the value :math:`\V128.\CONST~c` onto the stack.
```
binji, line 701: typo `5.`

document/core/exec/numerics.rst
```diff
@@ -135,6 +152,143 @@ When a number is stored into :ref:`memory <syntax-mem>`, it is converted into a
...
+
+SIMD values have the same underlying representation as an |i128|. They can also be interpreted as numeric values packed into a |V128| with a particular |shape|.
+
+.. math::
+   \begin{array}{lll@{\qquad}l}
+   \simdto_{t\K{x}N}(c) &=&
+     c_0~\dots~c_N \\
+     && (\where
+       w = |t| / 8 \wedge
+       b^* = bytes_{\i128}(i) \\ &&
```
binji: clearer if you move this to the next line

binji: use `c` instead of `i`?

binji: link to bytes auxiliary function

document/core/exec/numerics.rst
```diff
@@ -626,6 +804,214 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
    \end{array}
 
 
+.. _op-ibitselect:
+
+:math:`\ibitselect_N(i_1, i_2, c)`
+..................................
+
+* Use the bits in :math:`c` to select corresponding bit from :math:`i_1` when 1 and :math:`i_2` when 0.
```
binji, line 812: `:math:` around `1` and `0`

document/core/exec/numerics.rst
```diff
@@ -626,6 +804,214 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
+
+
+.. _op-iabs:
+
+:math:`\iabs_N(i)`
+..................
+
+* Let :math:`j` be the :ref:`signed interpretation <aux-signed>` of :math:`i`.
+
+* If :math:`j` greater than or equal to :math:`0`.
```
binji, line 827: missing ", then return `i`"

document/core/exec/numerics.rst
binji, line 812: Maybe rephrase this? I can understand not wanting to write out the exact behavior, but I think it should still match the "Return ..." structure.

document/core/exec/numerics.rst
```diff
@@ -626,6 +804,214 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
+
+* If :math:`j` is :math:`0`, then return :math:`0`.
+
+* Else if :math:`j` is positive, then return :math:`-j`.
+
+* Else return the negation of :math:`j`, modulo :math:`2^N`.
+
+.. math::
+   \begin{array}{@{}lcll}
+   \ineg_N(0) &=& 0 \\
```
binji, line 853: missing other cases

document/core/exec/numerics.rst
```diff
@@ -626,6 +804,214 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
+   \iabs_N(i) &=& \signed_N^{-1}(-j) \mod 2^N & (\iff j = \signed_N(i) \wedge j < 0) \\
+   \end{array}
+
+
+.. _op-ineg:
+
+:math:`\ineg_N(i)`
+..................
+
+* Let :math:`j` be the :ref:`signed interpretation <aux-signed>` of :math:`i`.
```
binji, line 843: Maybe easiest to define in terms of `x = ~y + 1`, e.g. `inot` and `iadd`

document/core/exec/numerics.rst
```diff
@@ -626,6 +804,214 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
+* Return :math:`i_1` if :math:`\igts_N(i_1, i_2) = 1`, :math:`i_2` otherwise.
+
+.. math::
+   \begin{array}{@{}lcll}
+   \iminu_N(i_1, i_2) &=& i_1 & (\iff \igts_N(i_1, i_2) = 1)\\
+   \iminu_N(i_1, i_2) &=& i_2
+   \end{array}
+
+
+.. _op-iaddsat_u:
```
binji, line 913: should be `add_sat_`, right?

document/core/exec/numerics.rst
```diff
@@ -626,6 +804,214 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
+* If :math:`j` is less than :math:`-2^{N-1}`, return :math:`-2^{N-1}`.
+
+* If :math:`j` is greater than :math:`2^N-1`, return :math:`2^N-1`.
+
+* Return :math:`j` otherwise.
+
+.. math::
+   \begin{array}{lll@{\qquad}l}
+   \iaddsats_N(i_1, i_2) &=& -2^{N-1} & (\iff \signed(i_i) + \signed(i_2) < -2^{N-1})\\
+   \iaddsats_N(i_1, i_2) &=& 2^N-1 & (\iff \signed(i_i) + \signed(i_2) > 2^N-1)\\
```
binji, line 950: use `saturate_s` below?

binji:
Did a pass on it, I really like extending the numerics, makes it much nicer to read IMO

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,489 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+SIMD Instructions
+~~~~~~~~~~~~~~~~~~~~
+
+SIMD instructions are defined in terms of generic numeric operators applied lane-wise based on the :ref:`shape <syntax-simd-shape>`.
+
+.. math::
+   \begin{array}{lll@{\qquad}l}
+   \X{op}_{t\K{x}N}(n_1,\dots,n_k) &=&
+     \simdof_{t\K{x}N}(op_t(\simdto_{t\K{x}N}(n_1) ~\dots~ \simdto_{t\K{x}N}(n_k))
```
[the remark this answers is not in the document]
ngzhian: Just updated.

document/core/exec/instructions.rst
ngzhian: I wanted the `t` and the `N` to specify the subscript to `vshiftop`. `shape` will work here too, then I will need `\lanes(\shape)` and `\packed(\shape)` instead of `t` and `N`.

document/core/exec/instructions.rst
ngzhian: Ah I'm missing something in the definition of `bool` to say that it works on a sequence on integers as well (AND over all of them).

document/core/exec/numerics.rst
ngzhian: Yea agree that this should be rephrased, tried something, but it still looks awkward, especially since the implementation below doesn't match the description.

document/core/exec/instructions.rst
ngzhian: I wanted to differentiate them a little more, since `c_1` is always a `v128`, but `v_1` is `i32` or `i64`. I'll change it to `c`, it's more consistent.

document/core/exec/instructions.rst
binji: Yeah, looks fine to me either way, I just thought you were using `shape` below

ngzhian: Yea that was a mistake, i fixed them to use `txN`.

rossberg (commented):
Some preliminary comments, haven't read all of exec/instructions yet.

document/core/exec/numerics.rst
ngzhian: we already have `shape` defined, it's a non-terminal that represents "i32x4" and friends.
Would calling this function `shape` as well be confusing?
Maybe we can call it `to_shape` or something.

document/core/exec/numerics.rst
```diff
@@ -173,6 +327,13 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
    \bool(C) &=& 0 & (\otherwise) \\
    \end{array}
 
+We lift :math:`\bool` to sequences of predicates by conjunction over the inputs.
+
+.. math::
+   \begin{array}{lll@{\qquad}l}
+   \bool(C_1~C_2^*) &=& \bool(C_1) \wedge \bool(C_2^*)
```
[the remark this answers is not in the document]
ngzhian: Do we need to define how `\bigwedge` works?

document/core/exec/numerics.rst
```diff
@@ -626,6 +811,195 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
+
+* Let :math:`d_i` be the bit :math:`i` of :math:`i_1` if bit :math:`i` of :math:`c` is :math:`1`
+
+* Else let :math:`d_i` be the bit :math:`i` of of :math:`i_2`.
+
+* Return the number represented by all the concatenation of all bits :math:`d_i`.
+
+.. math::
+   \begin{array}{@{}lcll}
+   \ibitselect_N(i_1, i_2, N) &=& \ior_N(\iand_N(i_1, c), \iand_N(i_2, \inot_N(c)))
```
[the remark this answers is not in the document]
ngzhian: Make sense, I chose to fix the prose to match the current formal definition. And fixed the N too.

document/core/exec/numerics.rst
```diff
@@ -1621,3 +1995,74 @@ Conversions
    \begin{array}{lll@{\qquad}l}
    \reinterpret_{t_1,t_2}(c) &=& \bits_{t_2}^{-1}(\bits_{t_1}(c)) \\
    \end{array}
+
+
+.. _op-saturate_u:
+
+:math:`\saturateu_N(i)`
```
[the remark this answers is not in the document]
ngzhian: Yup, fixed.

document/core/exec/numerics.rst
```diff
@@ -135,6 +152,143 @@ When a number is stored into :ref:`memory <syntax-mem>`, it is converted into a
...
+       bits_{f32}^{-1}(d_2^{32}) ~
+       bits_{f32}^{-1}(d_3^{32}) \\ &&
+       \qquad (\iff \ibits_{128}(i) = d_3^{32}~d_2^{32}~d_1^{32}~d_0^{32}) \\
+     \simdto_{f64x2}(i) &=&
+       bits_{f64}^{-1}(d_0^{64}) ~
+       bits_{f64}^{-1}(d_1^{64}) \\ &&
+       (\iff \ibits_{128}(i) = d_1^{64}~d_0^{64}) \\
+     \end{array}
+
+These functions are bijections, so they are invertible. For convenience, we give a name to the inverse function:
```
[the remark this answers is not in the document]
ngzhian: Sg, removed \simdof and also changed all usages to use the inverse.

document/core/exec/numerics.rst
rossberg: Hm, yeah, it might be a bit odd. In math you don't usually have "to" prefixes. How about "shaped", somewhat analogous to "unpacked"?

document/core/exec/numerics.rst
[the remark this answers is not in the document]
rossberg: We could call it out as more auxiliary notation in syntax/conventions, like so:
```
* :math:`\bigwedge s^\ast` and :math:`\bigvee s^\ast` denote the conjunction and disjunction of a sequence of conditions, respectively; if the sequence is empty, then the result is vacuously true, respectively false.
```
OTOH, it's more or less established math notation and fairly intuitive, so I suppose it would be understandable without extra explanation. WDYT?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,540 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+2. Pop the value :math:`\V128.\CONST~c_2` from the stack.
+
+3. Let :math:`s^{16}` be the sequence :math:`\simdto_{i8x16}(c_2)`.
+
+4. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+5. Let :math:`i^{16}` be the sequence :math:`\simdto_{i8x16}(c_1)`.
+
+6. Let :math:`v^{255}` be the concatenation of the two sequences :math:`i^{16}~0^{239}`
```
binji, line 302: should be `256` and `240`, right?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,540 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+7. Let :math:`d_i` be the result of :math:`d_i = v^{255}[ s^{16}[i] ]`.
```
binji, line 304: Can this be `d^16`, and then it doesn't need to be expanded explicitly below?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,540 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+
+.. _exec-simd-shuffle:
+
+:math:`\K{i8x16.}\SHUFFLE~\laneidx^{16}`
+........................................
+
+1. Assert: due to :ref:`validation <valid-simd-shuffle>`, two values of :ref:`value type <syntax-valtype>` |V128| are on the top of the stack.
+
+1. Assert: due to :ref:`validation <valid-simd-shuffle>`, that all :math:`\laneidx_i < 32`.
```
binji, line 332: fix step numbers

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,540 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+3. Let :math:`i_1^{16}` be the sequence :math:`\simdto_{i8x16}(c_2)`.
+
+4. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+3. Let :math:`i_2^{16}` be the sequence :math:`\simdto_{i8x16}(c_2)`.
+
+3. Let :math:`i_3^{32}` be the concatenation of the two sequences :math:`i_1^{16}~i_2^{16}`.
+
+4. Let :math:`d_i` be :math:`i_3^{32}[\laneidx_i]`.
```
binji, line 344: not sure the convention, but it seems like we'd want `d^16` and `laneidx^16` here too

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,540 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+1. Assert: due to :ref:`validation <valid-vtestop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+
+2. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+3. Let :math:`i_1^N` be the sequence :math:`\simdto_{t_1\K{x}M}(c_1)[N \slice N]`.
+
+4. Let :math:`c` be the result of computing :math:`\simdto^{-1}_{t_2\K{x}N}((\extend^{\sx}_{t_1,t_2}(i_1))^N)`
+
+6. Push the value :math:`\V128.\CONST~c` onto the stack.
```
binji, line 704: typo: 5.

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,540 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+2. Pop the value :math:`\I32.\CONST~s` from the stack.
+
+3. Assert: due to :ref:`validation <valid-vshiftop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+
+4. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+5. Let :math:`i_1^N` be the sequence :math:`\simdto_{t\K{x}N}(c_1)`.
+
+7. Let :math:`c` be :math:`\simdto^{-1}_{t\K{x}N}(\vshiftop_{t}(i_1^N, s^N))`.
```
binji, line 534: renumber

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,540 @@ Where the underlying operators are non-deterministic, because they may return on
...
+.. _exec-simd-widen:
+
+:math:`t_2\K{x}N\K{.}\WIDEN\_\K{low}\_t_1\K{x}M\_\sx`
+.....................................................
+
+1. Assert: due to :ref:`validation <valid-vtestop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+
+2. Pop the value :math:`\V128.\CONST~c_1` from the stack.
+
+3. Let :math:`i_1^N` be the sequence :math:`\simdto_{t_1\K{x}M}(c_1)[0 \slice N]`.
```
binji, line 674: Maybe just me, but these steps make me uncomfortable unless I know that N = M * 2. Maybe we can assert here? (same for the load extend below).

document/core/exec/instructions.rst
```diff
@@ -464,6 +998,98 @@ Memory Instructions
...
+
+12. If :math:`\K{splat}` is part of the instruction, then:
+
+    a. Let :math:`n` be the integer for which :math:`\bytes_{\iN}(n) = b^\ast`.
+
+    b. Let :math:`L` be the integer :math:`128 / N`.
+
+    c. Let :math:`c` be the result of computing :math:`\simdto^{-1}_{\iN\K{x}L}(n^L)`.
+
+12. Else:
```
binji, line 1041: typo: 13.

document/core/exec/instructions.rst
```diff
@@ -464,6 +998,98 @@ Memory Instructions
...
+
+7. Pop the value :math:`\I32.\CONST~i` from the stack.
+
+8. Let :math:`\X{ea}` be the integer :math:`i + \memarg.\OFFSET`.
+
+9. If :math:`N` is not part of the instruction, then:
+
+   a. Let :math:`N` be :math:`M * L`.
+
+10. If :math:`\X{ea} + N/8` is larger than the length of :math:`\X{mem}.\MIDATA`, then:
```
binji, line 1027: This means that, even though the load-extend instructions read 8 bytes, they'll fail if they can't read 16 bytes. Is that intentional?

EDIT: Ah, n/m. This works because N for extend instructions is actually half the full size.

document/core/exec/instructions.rst
```diff
@@ -464,6 +998,98 @@ Memory Instructions
    \end{array}
 
 
+.. _exec-load-extend:
+.. _exec-load-splat:
+
+:math:`\V128\K{.}\LOAD{M}\K{x}L\_\sx~\memarg` and :math:`\V128\K{.}\LOAD{N}\K{\_splat}~\memarg`
```
binji, line 1004: It feels like these should be split into separate definitions, IMO

document/core/exec/instructions.rst
```diff
@@ -464,6 +998,98 @@ Memory Instructions
...
+
+    a. Let :math:`n` be the integer for which :math:`\bytes_{\iN}(n) = b^\ast`.
+
+    b. Let :math:`L` be the integer :math:`128 / N`.
+
+    c. Let :math:`c` be the result of computing :math:`\simdto^{-1}_{\iN\K{x}L}(n^L)`.
+
+12. Else:
+
+    a. Let :math:`m_i` be the integer for which :math:`\bytes_{\iM}(m_i) = b^\ast[i*M/8 \slice M/8]`.
```
binji, line 1043: Why can't this just be `b*`?

document/core/exec/numerics.rst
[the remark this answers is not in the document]
ngzhian: I agree that it's established notation and fairly intuitive, especially for someone reading formal notation. I'll use `\bigwedge` directly, thanks!

document/core/exec/instructions.rst
ngzhian: Thank you! off-by-one, confused length and indices.

document/core/exec/instructions.rst
ngzhian: Yea good idea, reworded it using `\dots`, that folds that sequence in nicely.

document/core/exec/instructions.rst
ngzhian: Reworded it like swizzle, so removed the need for `d_i` here.

document/core/exec/instructions.rst
ngzhian: I'm not sure how to assert here, because we always assert due to validation, and this isn't really a validation as N = M * 2 is guaranteed by the accepted syntax.

document/core/exec/instructions.rst
ngzhian: Yea good point, I'll split them up.

document/core/exec/instructions.rst
ngzhian: `b*` is the entire 8 bytes that is loaded, we want to extend it lane-by-lane (`m_i`), hence only taking a slice of size `M/8`.

document/core/exec/instructions.rst
ngzhian: Yea sorry for the confusion, I took your suggestion to split load extend and load splat up, that should make it clearer!

document/core/exec/numerics.rst
ngzhian: Yea shaped works, `shaped_{i64x2}(c)`, though I have to think it for half a second to see what it does - the grammar doesn't look like something I'm used to (this is more declarative than verb-y)

Couple more suggestions we might want to consider (in no order of preference)
-  `elements_{i64x2}(c)`
-  `scalarize_{i64x2}(c)`
-  `split_{i64x2}(c)`
-  `sequence_{i64x2}(c)`
-  `lanes_{i64x2}(c)` (this is a collision with an existing definition, which is an auxiliary function that returns the number of lanes, but we can rename that to num_lanes)

document/core/exec/instructions.rst
binji: Ah, that makes sense. Thanks!

document/core/exec/numerics.rst
ngzhian: @rossberg what do you think about the list of suggestion, does any one look better than the rest?

ngzhian: @rossberg any thoughts? Do you still prefer `shaped`?

rossberg: Ah, sorry, I somehow had missed these. `lanes` actually sounds best to me. How about we rename the existing `lanes` function to `dim`?

ngzhian: Sg, renamed `lanes` to `dim`, then renamed `simdto` to `lanes`, in two separate commit, hopefully easier to review.

binji:
It's a little tricky to review this, since there are a lot of changes and the force-pushes mean I have to look at everything. Do you have an overview of the changes since the last time?

ngzhian:
Yup I think you can start looking at this commit onwards https://github.com/WebAssembly/simd/pull/340/commits/2725a788440a0ac419dffb7e42ef66d35294bf37 which addresses your review comments previously, the notable changes are:

- swizzle and shuffle got reworded
- fixes in sequence lengths in swizzle
- bullet point numbering fixes
- split out load splat and load extend
- reworded extract lane
- reword simd_to description

And following Andreas' comments:

- rename lanes to dim
- rename simd_to to lanes

binji (approved):
Sorry for the delay -- didn't recheck everything, but what I did looks good.

document/core/exec/instructions.rst
```diff
@@ -464,6 +985,126 @@ Memory Instructions
...
+
+9. Let :math:`N` be :math:`M \cdot L`.
+
+10. If :math:`\X{ea} + N/8` is larger than the length of :math:`\X{mem}.\MIDATA`, then:
+
+    a. Trap.
+
+11. Let :math:`b^\ast` be the byte sequence :math:`\X{mem}.\MIDATA[\X{ea} \slice N/8]`.
+
+12. Let :math:`m_i` be the integer for which :math:`\bytes_{\iM}(m_i) = b^\ast[i*M/8 \slice M/8]`.
```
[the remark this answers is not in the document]
ngzhian: Variable shadowing here. The `i` in `m_i` is used as an index to refer to each lane, since each lane is loaded then extended. It has nothing to do with the `i` above which is the memory index.
I changed the `i` in `m_i` to `k`.

document/core/exec/instructions.rst
```diff
@@ -464,6 +985,126 @@ Memory Instructions
...
+   \begin{array}{l}
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~i)~(\V128.\LOAD{M}\K{x}L\_\sx~\memarg) &\stepto&
+     S; F; (\V128.\CONST~c)
+   \end{array}
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & \X{ea} = i + \memarg.\OFFSET \\
+     \wedge & \X{ea} + (M*L) \leq |S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA| \\
+     \wedge & \bytes_{\iM}(m_i) = S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA[\X{ea} + i * (M/8) \slice M/8]) \\
```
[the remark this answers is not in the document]
ngzhian: Fixed in the same way.

ngzhian:
@rossberg can you please take another look at this? All of the outstanding comments have been resolved.

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,527 @@ Where the underlying operators are non-deterministic, because they may return on
...
+5. Let :math:`i_2^N` be the sequence as a result of computing :math:`\ilts_{B}(i_1^N, 0^N)`.
+
+6. Let :math:`c` be the integer :math:`\ibits_{32}^{-1}(i_2^N~0^{32-N})`.
+
+7. Push the value :math:`\I32.\CONST~c` onto the stack.
+
+.. math::
+   \begin{array}{lcl@{\qquad}l}
+   (\V128\K{.}\VCONST~c_1)~\shape\K{.}\BITMASK &\stepto& (\I32\K{.}\CONST~c)
+     & (\iff c = \ibits_{32}^{-1}(\ilts_{B}(\lanes_{t\K{x}N}(c), 0^N)))
```
rossberg, line 614:
```suggestion
     & (\iff c = \ibits_{32}^{-1}(\ilts_{|t|}(\lanes_{t\K{x}N}(c), 0^N)))
```

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,527 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+5. Let :math:`i_2^N` be the sequence as a result of computing :math:`\ilts_{B}(i_1^N, 0^N)`.
+
+6. Let :math:`c` be the integer :math:`\ibits_{32}^{-1}(i_2^N~0^{32-N})`.
+
+7. Push the value :math:`\I32.\CONST~c` onto the stack.
+
+.. math::
+   \begin{array}{lcl@{\qquad}l}
+   (\V128\K{.}\VCONST~c_1)~\shape\K{.}\BITMASK &\stepto& (\I32\K{.}\CONST~c)
```
rossberg, line 613:
```suggestion
   (\V128\K{.}\VCONST~c_1)~t\K{x}N\K{.}\BITMASK &\stepto& (\I32\K{.}\CONST~c)
```

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,527 @@ Where the underlying operators are non-deterministic, because they may return on
...
+     & (\iff c = \ibits_{32}^{-1}(\ilts_{B}(\lanes_{t\K{x}N}(c), 0^N)))
+     \\
+   \end{array}
+
+
+.. _exec-simd-narrow:
+
+:math:`t_2\K{x}N\K{.}\NARROW\_t_1\K{x}M\_\sx`
```
rossberg, line 621:
```suggestion
:math:`t_2\K{x}N\K{.}\NARROW\K{\_}t_1\K{x}M\K{\_}\sx`
```

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,527 @@ Where the underlying operators are non-deterministic, because they may return on
...
+
+2. Pop the value :math:`\V128.\VCONST~c_2` from the stack.
+
+3. Let :math:`d_2^M` be the result of computing :math:`\narrow^{\sx}_{t_1, t_2}(\lanes_{t_1\K{x}M}(c_2))`.
+
+4. Pop the value :math:`\V128.\VCONST~c_1` from the stack.
+
+5. Let :math:`d_1^M` be the result of computing :math:`\narrow^{\sx}_{t_1, t_2}(\lanes_{t_1\K{x}M}(c_1))`.
+
+6. Let :math:`c` be the result of :math:`\lanes^{-1}_{t_2\K{x}N}(d_1^M~d_2^M)`.
```
rossberg, line 634: Does this need an assertion that N = 2M, due to syntactic constraints?

Btw, I'm now confused about the inclusion of bare `\NARROW` in vbinop in the syntax conventions. I suppose that should be `\NARROW_\shape`? Similar question about `\WIDEN`.

document/core/exec/instructions.rst
```diff
@@ -464,6 +985,126 @@ Memory Instructions
    \end{array}
 
 
+.. _exec-load-extend:
+
+:math:`\V128\K{.}\LOAD{M}\K{x}L\_\sx~\memarg`
```
rossberg, line 990: Nit: can we use N instead of L here?

document/core/exec/instructions.rst
```diff
@@ -464,6 +985,126 @@ Memory Instructions
...
+   ~\\[-1ex]
+   \begin{array}{l}
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~i)~(\V128.\LOAD{M}\K{x}L\_\sx~\memarg) &\stepto&
+     S; F; (\V128.\CONST~c)
+   \end{array}
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & \X{ea} = i + \memarg.\OFFSET \\
+     \wedge & \X{ea} + (M*L) \leq |S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA| \\
```
rossberg, line 1037:
```suggestion
     \wedge & \X{ea} + M \cdot L / 8 \leq |S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA| \\
```

document/core/exec/instructions.rst
```diff
@@ -464,6 +985,126 @@ Memory Instructions
...
+     \wedge & \bytes_{\iM}(m_k) = S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA[\X{ea} + k * (M/8) \slice M/8]) \\
+     \wedge & J = M*2 \\
+     \wedge & c = \lanes^{-1}_{\X{i}J\K{x}L}(\extend^{\sx}_{M,J}(m_0) \dots \extend^{\sx}_{M,J}(m_{L-1}))
```
rossberg, line 1040:
```suggestion
     \wedge & c = \lanes^{-1}_{\X{i}J\K{x}L}(\extend^{\sx}_{M,M\cdot2}(m_0) \dots \extend^{\sx}_{M,M\cdot2}(m_{L-1}))
```

rossberg (commented):
Sorry, I missed this somehow! Various nits and a number of small bugs. I ran out of steam for the load/store instructions, so I may have to take another pass on those.

document/core/exec/numerics.rst
```diff
@@ -625,6 +696,197 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
+* Let :math:`j` be the :ref:`signed interpretation <aux-signed>` of :math:`i`.
+
+* If :math:`j` greater than or equal to :math:`0`, then return :math:`i`.
+
+* Else return the negation of `j`, modulo :math:`2^N`.
+
+.. math::
+   \begin{array}{@{}lcll}
+   \iabs_N(i) &=& i & (\iff \signed_N(i) \ge 0) \\
+   \iabs_N(i) &=& \signed_N^{-1}(-j) \mod 2^N & (\iff j = \signed_N(i) \wedge j < 0) \\
```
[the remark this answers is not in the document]
ngzhian: I think the new change should be:
`\iabs_N(i) &=& -\signed_N(i) \mod 2^N & (\otherwise) \\`

from `\signed_N(j)` to \signed_N(i)`, right?

document/core/exec/numerics.rst
```diff
@@ -1620,3 +1882,70 @@ Conversions
...
+   \satu_N(i) &=& i & (\otherwise)
+   \end{array}
+
+
+.. _op-sat_s:
+
+:math:`\sats_N(i)`
+.......................
+
+* Let :math:`j` be the :ref:`signed interpretation <aux-signed>` of :math:`i` of size :math:`N`.
```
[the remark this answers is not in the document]
ngzhian: This function is not always called with signed integers:
- `\isubsatu_N(i_1, i_2)` calls it with the result of `i_1 - i_2`.
- `\addsatu_N(i_1, i_2)` calls it with the result of `i_1 + i_2`.

I can change these two calls so that we can drop the reinterpretation step here.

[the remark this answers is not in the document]
ngzhian: Oh oops, this is sat_s, i was looking at sat_u, ignore my comment.

document/core/exec/numerics.rst
```diff
@@ -1620,3 +1882,70 @@ Conversions
...
+.......................
+
+* If :math:`i` is greater than :math:`2^N-1`, return :math:`2^N-1`.
+
+* Else return :math:`i`.
+
+.. math::
+   \begin{array}{lll@{\qquad}l}
+   \satu_N(i) &=& 2^N-1 & (\iff i > 2^N-1)\\
+   \satu_N(i) &=& i & (\otherwise)
```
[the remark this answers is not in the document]
ngzhian: I think to be consistent, I will change `subsat_u` to pass the arg as a signed interpretation, so everything is consistent (both `satu` and `sats` will both take integers as signed interpretation)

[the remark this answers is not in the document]
ngzhian: Then in that case we don't need a case for values that are negative in two's complement.

document/core/exec/numerics.rst
```diff
@@ -1620,3 +1882,70 @@ Conversions
...
+   \narrows_{M,N}(i) &=& \sats_N(\signed_M(i))
+   \end{array}
+
+
+.. _op-narrow_u:
+
+:math:`\narrowu_{M,N}(i)`
+.........................
+
+* Let :math:`j` be the :ref:`signed interpretation <aux-signed>` of :math:`i` of size :math:`M`.
```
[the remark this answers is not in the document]
ngzhian: Yea the definition of narrow (at least for the SIMD instruction) is that the value is interpreted as a signed value. See https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#integer-to-integer-narrowing "Regardless of the whether the operation is signed or unsigned, the input lanes are interpreted as signed integers."

document/core/exec/numerics.rst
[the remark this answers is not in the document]
ngzhian: Nvm, we need a case for i < 0, since it is signed.

document/core/exec/numerics.rst
```diff
@@ -1620,3 +1882,70 @@ Conversions
...
+   \begin{array}{lll@{\qquad}l}
+   \sats_N(i) &=& -2^{N-1} & (\iff \signed(i) < -2^{N-1})\\
+   \sats_N(i) &=& 2^N-1 & (\iff \signed(i) > 2^N-1)\\
+   \sats_N(i) &=& i & (\otherwise)
+   \end{array}
+
+
+.. _op-narrow_s:
+
+:math:`\narrows_{M,N}(i)`
```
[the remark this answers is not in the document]
ngzhian: It looks like this convention of `i<operation>` is for the numeric operators. For these conversion, `extend` is just `extend`, not `iextend`. Should we keep the name as `narrow`?

document/core/exec/numerics.rst
```diff
@@ -1620,3 +1882,70 @@ Conversions
    \begin{array}{lll@{\qquad}l}
    \reinterpret_{t_1,t_2}(c) &=& \bits_{t_2}^{-1}(\bits_{t_1}(c)) \\
    \end{array}
+
+
+.. _op-sat_u:
```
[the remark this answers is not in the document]
ngzhian: Done, moved the aux-sat_{,s} up to the top. Ptal.

document/core/exec/numerics.rst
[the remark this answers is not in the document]
rossberg: Yes, the arguments need to be allowed to be negative as well as larger than the N value range, so they cannot be Wasm values.

document/core/exec/numerics.rst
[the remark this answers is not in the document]
rossberg: Sunds plausible, but if this has the same status as plain `extend`, then don't you need an `inarrow` wrapper as well, same as for extend? Otherwise, how is it picked up by the operator conventions?

document/core/exec/numerics.rst
```diff
@@ -79,6 +79,30 @@ Conventions:
...
+  * Unsigned saturation, :math:`\satu_N(i)` clamps :math:`i` to between :math:`0` and :math:`2^N-1`:
+
+    .. math::
+       \begin{array}{lll@{\qquad}l}
+       \satu_N(i) &=& 2^N-1 & (\iff i > 2^N-1)\\
+       \satu_N(i) &=& 0 & (\iff i < 0) \\
+       \satu_N(i) &=& i & (\otherwise) \\
+       \end{array}
+
+  * Signed saturation, :math:`\sats_N(i)` clamps :math:`i` to between :math:`-2^{N-1}` and :math:`2^N-1`:
```
rossberg, line 96:
```suggestion
  * Signed saturation, :math:`\sats_N(i)` clamps :math:`i` to between :math:`-2^{N-1}` and :math:`2^{N-1}-1`:
```

document/core/exec/numerics.rst
```diff
@@ -79,6 +79,30 @@ Conventions:
...
+
+  .. math::
+     \begin{array}{lll@{\qquad}l}
+     \sats_N(i) &=& \signed_N^{-1}(-2^{N-1}) & (\iff i < -2^{N-1})\\
+     \sats_N(i) &=& \signed_N^{-1}(2^N-1) & (\iff i > 2^N-1)\\
```
rossberg, line 101:
```suggestion
     \sats_N(i) &=& \signed_N^{-1}(2^{N-1}-1) & (\iff i > 2^{N-1}-1)\\
```

document/core/exec/numerics.rst
```diff
@@ -809,11 +833,11 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
 
 * Let :math:`i` be the result of adding :math:`i_1` and :math:`i_2`.
 
-* Return :math:`\satu_N(i)`.
+* Return :math:`\satu_N(\signed_N(i))`.
```
rossberg, line 836:
```suggestion
* Return :math:`\satu_N(i)`.
```
This isn't correct here, since the result of i1+i2 is a plain number, not a two's complement representation.

document/core/exec/numerics.rst
```diff
@@ -809,11 +833,11 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
 
 .. math::
    \begin{array}{lll@{\qquad}l}
-   \iaddsatu_N(i_1, i_2) &=& \satu_N(i_1 + i_2)
+   \iaddsatu_N(i_1, i_2) &=& \satu_N(\signed_N(i_1 + i_2))
```
rossberg, line 840:
```suggestion
   \iaddsatu_N(i_1, i_2) &=& \satu_N(i_1 + i_2)
```

document/core/exec/numerics.rst
```diff
@@ -843,11 +867,11 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
 
 * Let :math:`i` be the result of subtracting :math:`i_2` from :math:`i_1`.
 
-* Return :math:`\satu_N(i)`.
+* Return :math:`\satu_N(\signed_N(i))`.
```
rossberg, line 870:
```suggestion
* Return :math:`\satu_N(i)`.
```

document/core/exec/numerics.rst
```diff
@@ -843,11 +867,11 @@ The integer result of predicates -- i.e., :ref:`tests <syntax-testop>` and :ref:
...
 
 .. math::
    \begin{array}{lll@{\qquad}l}
-   \isubsatu_N(i_1, i_2) &=& \satu_N(i_1 - i_2)
+   \isubsatu_N(i_1, i_2) &=& \satu_N(\signed_N(i_1 - i_2))
```
rossberg, line 874:
```suggestion
   \isubsatu_N(i_1, i_2) &=& \satu_N(i_1 - i_2)
```

document/core/exec/numerics.rst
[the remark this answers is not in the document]
rossberg: Ah, yes, thanks.

document/core/exec/numerics.rst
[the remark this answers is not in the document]
ngzhian: In [exec/instructions.rst](https://www.ngzhian.com/simd/core/exec/instructions.html#t-2-mathsf-x-n-mathsf-xref-syntax-instructions-syntax-instr-simd-mathsf-narrow-t-1-mathsf-x-m-xref-syntax-instructions-syntax-sx-mathit-sx) we don't use the operator conventions, we directly refer to `narrow` with the correct types from the instruction's name.

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,527 @@ Where the underlying operators are non-deterministic, because they may return on
...
+.. math::
+   \begin{array}{l}
+   \begin{array}{lcl@{\qquad}l}
+   (\V128\K{.}\VCONST~c_1)~\shape\K{.}\ALLTRUE &\stepto& (\I32\K{.}\CONST~i)
+   \end{array}
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & n = \dim(\shape) \\
+     \wedge & i_1^n = \lanes_{\shape}(c) \\
+     \wedge & i = \bigwedge(i_1 \neq 0)^n
```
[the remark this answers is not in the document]
ngzhian: I used `\ast` instead of `*`.

document/core/exec/instructions.rst
ngzhian: `N` is already used later on, defined to be `M * L`. This follows the other memory instructions which use `N` as the number of bits actually loaded/stored.

document/core/exec/instructions.rst
ngzhian: > Does this need an assertion that N = 2M, due to syntactic constraints?

If we only get here from valid instructions, then we don't need such an assertion. The valid instructions are already defined such that N = 2M.

> Btw, I'm now confused about the inclusion of bare \NARROW in vbinop in the syntax conventions.

What I intended is that these narrow instructions are validated the same as the other vbinop, but for execution, it "pattern-matches" to this more specific definition of narrow (`shape.vbinop` wouldn't work for narrow/widen). I'm not sure if this works though.

I guess narrow and widen probably shouldn't be included in the syntax conventions, and should be separated out in the validation?

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,527 @@ Where the underlying operators are non-deterministic, because they may return on
...
+     & (\iff t = \unpacked(\shape)
+       \wedge c = \lanes^{-1}_{\shape}(c_1^{\dim(\shape)}))
+     \\
+   \end{array}
+
+
+.. _exec-simd-extract-lane:
+.. _exec-simd-extract-lane-sx:
+
+:math:`t_1\K{x}N\K{.}\EXTRACTLANE\K{\_}\sx^?~\laneidx`
```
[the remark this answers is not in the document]
ngzhian: > Nit: rename \laneidx to just x.

Done.

> Also, for stylistic consistency

I used `\shape` in places where we don't need to know the underlying shape. But for instructions like `shift`, `narrow`, etc, it is useful to know the components of the shape, and where `unpacked` wouldn't work (we want to know that it is an `i8` and not a `i32`). I guess we could create another helper for this?

ngzhian:
@rossberg Thanks for the careful review. I have resolved all the minor issues. Left a couple of responses and questions for you. And apologies for not doing the recent changes at one go. I missed that there was a "Show X hidden conversations" button that had a couple of your review notes, and so I didn't fix those issues during my first go.

document/core/exec/instructions.rst
rossberg: > If we only get here from valid instructions, then we don't need such an assertion. The valid instructions are already defined such that N = 2M.

Yup, that's why I suggested saying "due to syntactic constraints". The nature of an assertion of course is that it's always true. But the constraint is not apparent from the pattern match above, so local to this rule, it is not clear otherwise.

> > Btw, I'm now confused about the inclusion of bare \NARROW in vbinop in the syntax conventions.
> 
> What I intended is that these narrow instructions are validated the same as the other vbinop, but for execution, it "pattern-matches" to this more specific definition of narrow (`shape.vbinop` wouldn't work for narrow/widen). I'm not sure if this works though.

You can still do so, but you have to list them in vunop/vbinop with their suffixes, i.e. `\NARROW\K{\_}\shape\K{\_}\sx` and `\WIDEN\K{\_low\_}\shape\K{\_}\sx | \WIDEN\K{\_high\_}\shape\K{\_}\sx`, respectively. The same for `\VTRUNC` and `\VCONVERT`, now that I notice it.

document/core/exec/instructions.rst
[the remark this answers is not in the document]
rossberg: I see. You wouldn't need a helper, you could just pattern match later saying something like "Let txN be the shape \shape." But maybe that's causing more harm than benefit, so I'm fine leaving it as is.

document/core/exec/instructions.rst
```diff
@@ -293,65 +293,64 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
...
+      \wedge & c' = \lanes^{-1}_{i8x16}(c^\ast[ i^\ast[0] ] \dots c^\ast[ i^\ast[15] ])
      \end{array}
    \end{array}
 
 
 .. _exec-simd-shuffle:
 
-:math:`\K{i8x16.}\SHUFFLE~\laneidx^{16}`
+:math:`\K{i8x16.}\SHUFFLE~x^\ast`
 ........................................
```
rossberg, line 325: Nit: adjust underline

document/core/exec/instructions.rst
```diff
@@ -293,65 +293,64 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
...
 
 1. Assert: due to :ref:`validation <valid-simd-shuffle>`, two values of :ref:`value type <syntax-valtype>` |V128| are on the top of the stack.
 
-2. Assert: due to :ref:`validation <valid-simd-shuffle>`, that all :math:`\laneidx_i < 32`.
+2. Assert: due to :ref:`validation <valid-simd-shuffle>`, that all :math:`x^\ast[i] < 32`.
```
rossberg, line 329:
```suggestion
2. Assert: due to :ref:`validation <valid-simd-shuffle>`, for all :math:`x_i` in :math:`x^\ast` it holds that :math:`x_i < 32`.
```

document/core/exec/instructions.rst
```diff
@@ -293,65 +293,64 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
...
-4. Let :math:`i_1^{16}` be the sequence :math:`\lanes_{i8x16}(c_2)`.
+4. Let :math:`i_2^\ast` be the sequence :math:`\lanes_{i8x16}(c_2)`.
 
 5. Pop the value :math:`\V128.\VCONST~c_1` from the stack.
 
-6. Let :math:`i_2^{16}` be the sequence :math:`\lanes_{i8x16}(c_2)`.
+6. Let :math:`i_1^\ast` be the sequence :math:`\lanes_{i8x16}(c_1)`.
 
-7. Let :math:`i_3^{32}` be the concatenation of the two sequences :math:`i_1^{16}~i_2^{16}`.
+7. Let :math:`i` be the concatenation of the two sequences :math:`i_1^\ast~i_2^\ast`.
```
rossberg, line 339:
```suggestion
7. Let :math:`i^\ast` be the concatenation of the two sequences :math:`i_1^\ast~i_2^\ast`.
```

document/core/exec/instructions.rst
```diff
@@ -293,65 +293,64 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
...
 
-8. Let :math:`c` be the result of :math:`\lanes^{-1}_{i8x16}(i_3^{32}[\laneidx_0] \dots i_3^{32}[\laneidx_{15}])`.
+8. Let :math:`c` be the result of :math:`\lanes^{-1}_{i8x16}(i[x^\ast[0]] \dots i[x^\ast[15]])`.
```
rossberg, line 341:
```suggestion
8. Let :math:`c` be the result of :math:`\lanes^{-1}_{i8x16}(i^\ast[x^\ast[0]] \dots i^\ast[x^\ast[15]])`.
```

document/core/exec/instructions.rst
```diff
@@ -293,65 +293,64 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
...
-   (\V128\K{.}\VCONST~c_1)~(\V128\K{.}\VCONST~c_2)~\V128\K{.}\SHUFFLE~\laneidx^{16} &\stepto& (\V128\K{.}\VCONST~c)
+   (\V128\K{.}\VCONST~c_1)~(\V128\K{.}\VCONST~c_2)~\V128\K{.}\SHUFFLE~x^\ast &\stepto& (\V128\K{.}\VCONST~c)
    \end{array}
    \\ \qquad
      \begin{array}[t]{@{}r@{~}l@{}}
-      (\iff & i_1^{16} = \lanes_{i8x16}(c_1) \\
-      \wedge & i_2^{16} = \lanes_{i8x16}(c_2) \\
-      \wedge & c = \lanes^{-1}_{i8x16}((i_1^{16}~i_2^{16})[\laneidx_0] \dots (i_1^{16}~i_2^{16})[\laneidx_{15}])
+      (\iff & i = \lanes_{i8x16}(c_1)~\lanes_{i8x16}(c_2) \\
+      \wedge & c = \lanes^{-1}_{i8x16}(i[x^\ast[0]] \dots i[x^\ast[15]])
```
rossberg, line 353:
```suggestion
      (\iff & i^\ast = \lanes_{i8x16}(c_1)~\lanes_{i8x16}(c_2) \\
      \wedge & c = \lanes^{-1}_{i8x16}(i^\ast[x^\ast[0]] \dots i^\ast[x^\ast[15]])
```

document/core/exec/instructions.rst
```diff
@@ -382,74 +381,70 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
    \end{array}
 
 
-.. _exec-simd-extract-lane:
-.. _exec-simd-extract-lane-sx:
+.. _exec-simd-extract_lane:
 
-:math:`t_1\K{x}N\K{.}\EXTRACTLANE\K{\_}\sx^?~\laneidx`
+:math:`t_1\K{x}N\K{.}\EXTRACTLANE\K{\_}\sx^?~x`
 ......................................................
```
rossberg, line 387: Nit: adjust underline

document/core/exec/instructions.rst
```diff
@@ -382,74 +381,70 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
...
      \end{array}
    \end{array}
 
 
-.. _exec-simd-replace-lane:
+.. _exec-simd-replace_lane:
 
-:math:`\shape\K{.}\REPLACELANE~\laneidx`
+:math:`\shape\K{.}\REPLACELANE~x`
 ........................................
```
rossberg, line 419: Nit: underline

document/core/exec/instructions.rst
rossberg: Yes, but N is an intermediate that does not actually occur in the formal rule, so it's a bit odd skipping it from that perspective.

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,519 @@ Where the underlying operators are non-deterministic, because they may return on
...
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & i^\ast = \lanes_{t\K{x}N}(c_1) \\
+     \wedge & c = \lanes^{-1}_{t\K{x}N}(\vshiftop_{t}(i^\ast, s^N)))
+     \end{array}
+   \end{array}
+
+
+.. _exec-vitestop:
+.. _exec-simd-alltrue:
```
rossberg, line 535: Nit: just noticed that this label is named inconsistently as well.
```suggestion
.. _exec-simd-all_true:
```

document/core/exec/instructions.rst
```diff
@@ -182,6 +182,519 @@ Where the underlying operators are non-deterministic, because they may return on
...
+   \end{array}
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & i_1^\ast = \lanes_{\shape}(c) \\
+     \wedge & i = \bool(\bigwedge(i_1 \neq 0)^\ast)
+     \end{array}
+   \end{array}
+
+
+.. _exec-simd-anytrue:
```
rossberg, line 564:
```suggestion
.. _exec-simd-any_true:
```

document/core/exec/instructions.rst
ngzhian: > Yup, that's why I suggested saying "due to syntactic constraints".

Added this assertion.

> You can still do so, but you have to list them in vunop/vbinop with their suffixes,

Got it, fixed.

document/core/exec/instructions.rst
```diff
@@ -613,19 +613,21 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
 :math:`t_2\K{x}N\K{.}\NARROW\K{\_}t_1\K{x}M\K{\_}\sx`
 .....................................................
 
-1. Assert: due to :ref:`validation <valid-vitestop>`, two values of :ref:`value type <syntax-valtype>` |V128| are on the top of the stack.
+1. Assert: due to :ref:`syntax <syntax-instr-simd>`, that :math:`N = 2M`.
```
rossberg, line 616:
```suggestion
1. Assert: due to :ref:`syntax <syntax-instr-simd>`, :math:`N = 2\cdot M`.
```

document/core/exec/instructions.rst
ngzhian: I see what you mean. I think we could simplify this, since `M*L` should always be 64, we can assert this (due to syntax) early on, and propagate 64 everywhere we use `N` or `M*L`.

rossberg: I would suggest simply inlining M\cdot L -- it is short enough and occurs only in two places I think.

document/core/exec/instructions.rst
```diff
@@ -464,6 +985,126 @@ Memory Instructions
...
+
+10. If :math:`\X{ea} + N/8` is larger than the length of :math:`\X{mem}.\MIDATA`, then:
+
+    a. Trap.
+
+11. Let :math:`b^\ast` be the byte sequence :math:`\X{mem}.\MIDATA[\X{ea} \slice N/8]`.
+
+12. Let :math:`m_k` be the integer for which :math:`\bytes_{\iM}(m_k) = b^\ast[k*M/8 \slice M/8]`.
+
+13. Let :math:`J` be the integer :math:`M*2`.
```
rossberg, line 1013:
```suggestion
13. Let :math:`J` be the integer :math:`M \cdot 2`.
```
Also, for no strong reason, I think renaming to W would look slightly nicer in the opcode below.

document/core/exec/instructions.rst
```diff
@@ -464,6 +979,125 @@ Memory Instructions
...
+   \begin{array}{lcl@{\qquad}l}
+   S; F; (\I32.\CONST~i)~(\V128.\LOAD{M}\K{x}L\_\sx~\memarg) &\stepto&
+     S; F; (\V128.\CONST~c)
+   \end{array}
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff & \X{ea} = i + \memarg.\OFFSET \\
+     \wedge & \X{ea} + M \cdot L / 8 \leq |S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA| \\
+     \wedge & \bytes_{\iM}(m_k) = S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA[\X{ea} + k * (M/8) \slice M/8]) \\
+     \wedge & c = \lanes^{-1}_{\X{i}M\cdot2\K{x}L}(\extend^{\sx}_{M,M\cdot2}(m_0) \dots \extend^{\sx}_{M,M\cdot2}(m_{L-1}))
```
rossberg, line 1033: Actually, I'm sorry, I changed my mind on inlining M*2 here, since it's too obscure in the opcode.
```suggestion
     \wedge & W = M\cdot2 \\
     \wedge & c = \lanes^{-1}_{\X{i}W\K{x}L}(\extend^{\sx}_{M,W}(m_0) \dots \extend^{\sx}_{M,W}(m_{L-1}))
```

ngzhian:
As a checklist, the remaining open comments are (listing here since some of the comments are hidden):

- [ ] argument to narrow_u being signed https://github.com/WebAssembly/simd/pull/340#discussion_r533849824
- [ ] narrow v.s. inarrow https://github.com/WebAssembly/simd/pull/340#discussion_r534696756

document/core/exec/instructions.rst
```diff
@@ -1000,41 +1000,40 @@ Memory Instructions
...
      S; F; (\V128.\CONST~c)
    \end{array}
    \\ \qquad
      \begin{array}[t]{@{}r@{~}l@{}}
      (\iff & \X{ea} = i + \memarg.\OFFSET \\
-     \wedge & \X{ea} + M \cdot L / 8 \leq |S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA| \\
-     \wedge & \bytes_{\iM}(m_k) = S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA[\X{ea} + k * (M/8) \slice M/8]) \\
-     \wedge & c = \lanes^{-1}_{\X{i}M\cdot2\K{x}L}(\extend^{\sx}_{M,M\cdot2}(m_0) \dots \extend^{\sx}_{M,M\cdot2}(m_{L-1}))
+     \wedge & \X{ea} + M \cdot N / 8 \leq |S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA| \\
+     \wedge & \bytes_{\iM}(m_k) = S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA[\X{ea} + k \cdot (M/8) \slice M/8]) \\
```
rossberg, line 1030:
```suggestion
     \wedge & \bytes_{\iM}(m_k) = S.\SMEMS[F.\AMODULE.\MIMEMS[0]].\MIDATA[\X{ea} + k \cdot M/8 \slice M/8]) \\
```

document/core/exec/numerics.rst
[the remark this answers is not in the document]
rossberg: Ah, I see. But that causes another issue that I missed before (sorry): there is a generic execution rule for shape.vbinop, and by the definition of vbinop, that should also apply to narrow. So there are two overlapping exec rules matching narrow. Similarly for vunop and widen.

So unlike what I said before, I think you'll need to remove those instructions from vunop and vbinop -- they don't fit the pattern, because unlike the others, they do not work lane-wise. (And consequently, they need separate typing rules.)

document/core/exec/numerics.rst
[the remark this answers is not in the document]
rossberg: Cool, thanks.

document/core/exec/numerics.rst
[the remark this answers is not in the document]
ngzhian: Done! See a5165cb, I created a new production `\vwiden` that just groups `widen_low` and `widen_high`, that way I don't have to repeat this in the validation.
`narrow` also has its own validation rule now.

rossberg (approved):
Great! Thanks for bearing with me.

rossberg:
Oops, I just noticed that there's a related issue with vtrunc and vconvert, whose execution does not map to that of a regular op either, because it involves two shapes. I think you'll need to introduce a vcvtop category and handle that separately. (Also, they are currently included in vbinop, which does not even seem correct for typing).

Feel free to fix that in a separate PR, if you prefer.

ngzhian:
>  I think you'll need to introduce a vcvtop category and handle that separately.

Good catch, thanks. Added a new production, `vcvtop`, which has its own validation and execution. 

The way I'm writing them now relies on this new note on execution/numerics: https://github.com/WebAssembly/simd/pull/340/files#diff-14d634364829defe70ba02a1ed0075266fce32b208b681eb5a812e96104e2bf9R41
My intention is that this also applies to the conversion operators, is that a correct assumption? Or should I clarify that section with an example showing how conversion operators work when given a list of inputs?

document/core/syntax/instructions.rst
```diff
@@ -376,7 +376,8 @@ Occasionally, it is convenient to group operators together according to the foll
      \viminmaxop ~|~ \visatbinop \\&&|&
      \SWIZZLE ~|~
      \VMUL ~|~
-     \AVGR\K{\_u} \\&&|&
+     \AVGR\K{\_u} \\
+   \production{conversion operator} & \vcvtop &::=&
      \VTRUNC\K{\_sat\_}\shape\K{\_}\sx ~|~
      \VCONVERT\K{\_}\shape\K{\_}\sx \\
```
rossberg: The way you are using vcvtop in the rule patterns, you now have to exclude the shape/sx suffixes here.

rossberg (approved):
Thanks. LGTM modulo comment.

Yes, the note should suffice. If you want to emphasise that it applies to heterogeneously typed ops like conversions as well, you could replace the simple fabs example that follows with convert.

document/core/syntax/instructions.rst
ngzhian: Oh yes, fixed.

PR #346, opened 2020-09-16
Create Blaneidx production, and encode lane index as u8

ngzhian:
Also fix up the definition in the syntax to be a u4 instead of a byte.

### DIFF
--- a/document/core/binary/instructions.rst
+++ b/document/core/binary/instructions.rst
@@ -411,7 +411,7 @@ SIMD Instructions
 ~~~~~~~~~~~~~~~~~~~~
 
 All variants of :ref:`SIMD instructions <syntax-instr-simd>` are represented by separate byte codes.
-The all have a one byte prefix, whereas the actual opcode is encoded by a variable-length :ref:`unsigned integer <binary-uint>`.
+They all have a one byte prefix, whereas the actual opcode is encoded by a variable-length :ref:`unsigned integer <binary-uint>`.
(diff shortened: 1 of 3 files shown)


binji (commented):
I don't see where this is defined as a `u4`...? But also, I don't think we want to change this to a `u4` anyway, without discussion, since that makes a larger lane index a decoding error instead of a validation error.

ngzhian:
Sorry, latest commit changed it to u8, but I didn't update the PR description, will change it now.

binji (approved):

document/core/binary/instructions.rst
```diff
@@ -442,34 +442,37 @@ The |VCONST| instruction is followed by 16 immediate bytes, which are converted
...
 .. _binary-vternop:
+.. _binary-laneidx:
 
-The |SHUFFLE| instruction is also followed by 16 immediate bytes:
+The |SHUFFLE| instruction is also followed by the encoding of 16 |laneidx| immediates.
 
 .. math::
    \begin{array}{llclll}
+   \production{lane index} & \Blaneidx &::=&
+     l{:}\Bu8 &\Rightarrow& l \\
```
rossberg, line 452: Note that Bu8 actually is a LEB128 representation. Here, you'll still need to use Bbyte.

ngzhian: Good catch, fixed in #357

PR #355, opened 2020-09-21
Update implementation status

ngzhian:
Add pmin/pmax and mark pmin/pmax and floating-point rounding
instructions as implemented on both V8 and SM (based on this
https://github.com/mozilla/gecko-dev/commit/807022887f155153d398f9f6f94ef59e214489a2).


lars-t-hansen (approved):
LGTM

PR #358, opened 2020-09-22
Cleanup operators to remove duplicated V128Op

ngzhian:


rossberg (commented):
You can simplify `V128 (V128Op.(V128 Xyz))` to `V128 V128Op.(V128 xyz)` in all these cases.

ngzhian:
Ah yes, fixed!

rossberg (approved):

PR #360, opened 2020-09-23
Add v128 to JS API

ngzhian:
Following the description given in
https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#javascript-api-and-simd-values
and also what I64 did before bigint integration.

js-api tests are not updated yet, that will come in a future patch.

### DIFF
--- a/document/js-api/index.bs
+++ b/document/js-api/index.bs
@@ -110,6 +110,7 @@ urlPrefix: https://webassembly.github.io/spec/core/; spec: WebAssembly; type: df
         text: i32.const
         text: f32.const
         text: f64.const
+        text: v128.const
     text: function index; url: syntax/modules.html#syntax-funcidx
     text: function instance; url: exec/runtime.html#function-instances
     text: store_init; url: appendix/embedding.html#embed-store-init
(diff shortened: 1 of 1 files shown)


document/js-api/index.bs
```diff
@@ -1039,6 +1054,7 @@ Note: Exported Functions do not have a \[[Construct]] method and thus it is not
 <div algorithm>
 The algorithm <dfn>ToJSValue</dfn>(|w|) coerces a [=WebAssembly value=] to a JavaScript value by performing the following steps:
 
+1. Assert: |w| is not of  the form [=v128.const=] <var ignore>v128</var>.
```
dtig, line 1057: Perhaps I'm not understanding this well, why is this assert required? Also, what happens for i64 values without BigInt support?

ngzhian: > why is this assert required?

This assert is to ensure that we never call `ToJSValue` with a v128.const, since we don't know how to convert it to JS. In all the callers of `ToJSValue` we should have checked for v128 and raise an exception prior.

> Also, what happens for i64 values without BigInt support?

I think BigInt support is required, not too sure.

dtig (approved):

PR #361, opened 2020-09-23
Add SIMD instructions to index of instructions

ngzhian:
Depends on #340.

Preview at https://www.ngzhian.com/simd/core/appendix/index-instructions.html note how rows with "execution, operator" in the last column are now taller. This is unfortunate due to how long some of the SIMD instructions are.


binji:
Did you use the [`gen-index-instructions.py`](https://github.com/WebAssembly/spec/blob/master/document/core/appendix/gen-index-instructions.py) script, or do it by hand?

binji (approved):
table lgtm, though I mostly did a spot check. It's a bit sad about the columns, but I think it's OK. Maybe we can tweak it a bit later.

ngzhian:
I forgot about gen-index-instructions.py... I'll fix it.

ngzhian:
First I'll need to merge upstream.

ngzhian:
Replaced by #366

Issue #221, opened 2020-05-05
Proposal to add fixed-point multiplication instructions


bjacob:
(Branched from Issue #175)

It would be useful to have fixed-point multiplication instructions, e.g. 32x32=32 and 16x16=16, similar to ARM SQRDMULH.

Some may think that the availability of a 32x32=64 integer multiplication (Issue #175) would remove the need for that, but that would be sub-optimal: staying within 32bit means doing 4 scalar operations per 128-bit vector operation, and most applications want to use the rounding flavor (SQRDMULH not SQDMULH) which would require a few more instructions to emulate if the instruction is missing, which in practice would result in applications making compromises between accuracy and performance.

(This is critical to integer-quantized neural network applications as performed by TensorFlow Lite using the ruy matrix multiplication library, see e.g. the usage of these instructions here, https://github.com/google/ruy/blob/57e64b4c8f32e813ce46eb495d13ee301826e498/ruy/kernel_arm64.cc#L517 )

dtig:
Thanks for filing this issue, as we're at [Phase 3](https://github.com/WebAssembly/meetings/blob/master/process/phases.md#3-implementation-phase-community--working-group) of the current SIMD proposal we've put a soft freeze on addition of new operations as discussed in #203. Some interesting questions to answer - the issue description highlights the usage of these instructions in neural network applications, are there other applications that would benefit from this addition? What would the corresponding codegen look like for the 32x32=32 case on Intel platforms?

nfrechette:
One use case is to use it with SSE4 (or is it AVX?) to perform a logical shift per SIMD lane. With SSE, logical shifts either take an immediate value and all lanes are shifted by the same amount or they take a shift amount as a u64 value and all lanes are shifted by the same amount. It is thus not possible to shift each lane by a different amount. One way to achieve this is with integer multiplication but it is only worth it when the intrinsic is available. It avoids the need to swizzle each lane, shift, and reconstruct. I'm not sure if it's faster with a multiplication but it uses a lot fewer instructions and registers and it inlines better. I intend to use this trick in my decompression code path where a vector3 is packed in a variable number of bits (each lane having the same number of bits). Due to bit alignment, when the value is loaded, it needs to be shifted. By using the bit offset, a lookup table can provide a shift value per lane.

bjacob:
> Thanks for filing this issue, as we're at Phase 3 of the current SIMD proposal 

[...]

Issue #343, opened 2020-09-10
Finalizing the instruction set


tlively:
It's been a while since we introduced the criteria in #203 for evaluating new proposed instructions. Since then, we have made significant progress on the spec tests, text, and interpreter, to the point where finalizing the instruction set will soon be the main phase 4 blocker.

We should continue evaluating currently proposed instructions using the previous criteria and we should continue to be open to bug  fixes and improvements to our existing instructions, but I propose that effective immediately (upon stakeholder agreement) we start labeling all new proposed instructions as post-mvp. Does that sound reasonable to everyone?

lars-t-hansen:
I think this is a good idea, but as part of this I think we should also come up with and make public an overall strategy for how incremental extensions will be made to the (fixed-width) instruction set.  Having such a strategy will make it easier to draw the line for the MVP.

tlively:
@lars-t-hansen, that's a good idea. My assumption was that we would lean heavily on the normal proposal process for follow-on extensions, but the more we sketch out a framework for how we will evaluate and group post-MVP instructions now, the more streamlined that those follow-on proposals will be. Did you have specific thoughts about what our strategy should be?

I think that one of the things that has worked out really well in this proposal has been having clear technical guidelines (e.g. portability, determinism) to focus development and prevent feature creep. Follow-on extensions to this proposal (i.e. not radically different proposals like scalable vectors or machine-specific instructions) will be easier to scope if they each focus on doing one thing well, either by focusing on a particular use case (e.g. neural networks) or a single instruction family (e.g. fma).

lars-t-hansen:
@tlively, I think grouping instructions thematically is going to be very helpful and I support that.  We need to be careful about how we define the groups so that we don't get into another (bulk memory, reference types, threads) quagmire, but maybe that's not too hard with these low-level instructions.

Part of our SIMD strategy will likely have to do with dealing with machine specific corner cases, though.  IIRC both arm64 and x64 have support for 8-bit operations but differ in some cases.  Also IIRC, the QFMA class of instructions had similar issues.  These are desirable instruction (groups) and we should probably be talking about principles for handling cross-architecture incompatibilities (beyond NaN cases) so that we can ship this type of functionality.

binji:
> These are desirable instruction (groups) and we should probably be talking about principles for handling cross-architecture incompatibilities (beyond NaN cases) so that we can ship this type of functionality.

[...]

PR #365, opened 2020-09-25
Saturating Rounding Q-format Multiplication

Maratyszcza:
Introduction
=========

Fixed-point algorithms often represent fractional numbers in [Q format](https://en.wikipedia.org/wiki/Q_(number_format)), where a fixed number of bits is allocated to fractional part. Addition and subtraction in Q format are no different than addition and subtraction on integers, but multiplication is substantially different. First, to produce multiplication result in the same Q format as the factors, we need to compute the full (i.e. with twice the number of bits of the factors) product of the factors, and then do arithmetic shift right by Q bits. Secondly, as the right shift loses some of the bits of the product, it is typically done with rounding to nearest Q-format number to minimize the rounding error. Thus, for Q15 format (i.e. with 15 fractional bits) the multiplication result is computed as `(a * b + 0x4000) >> 15` where `a` and `b` are integer representation of the input factors.

Despite sophisticated low-level definition, Q-format multiplication is commonly used in fixed-point algorithms, and widely supported in hardware. x86 since SSSE3 supports multiplication of 16-bit numbers in Q15 format, and ARM NEON supports multiplication of 16-bit numbers in Q15 format and 32-bit numbers in Q31 format. This proposal suggests to add 16-bit Q15-format multiplication to WebAssembly SIMD instruction set. The native x86 and ARM variants of this instruction differ in how they handle overflow: x86 version wraps around while ARM version saturates. The overflow happens only when both inputs are `INT16_MIN`, and can be corrected with a couple of instructions. For the purpose of bitwise compatibility this proposal standardize on the saturating variant as the more mathematically meaningful.

Q-format multiplication instruction is particularly important for fixed-point neural network inference, and was previously requested by @bjacob in #221.

Applications
=========

- [libvpx (video codec)](https://github.com/webmproject/libvpx/blob/a5d499e16570d00d5e1348b1c7977ced7af3670f/vpx_dsp/x86/inv_txfm_ssse3.h#L47-L52)
- [AV1 Codec Library (video codec)](https://github.com/mozilla/aom/blob/c8b38b0bfd36306f1886375d98512d792ffe9517/aom_dsp/x86/aom_subpixel_8t_intrin_ssse3.c#L797-L798)
- [OpenVINO (machine learning toolkit)](https://github.com/openvinotoolkit/openvino/blob/efad27d68ca88a107f9980b7dd8f07556a9eaf03/inference-engine/src/preprocessing/cpu_x86_sse42/ie_preprocess_gapi_kernels_sse42.cpp#L106-L109)
- [gemmlowp (fixed-point matrix multiplication)](https://github.com/google/gemmlowp/blob/023c190f9341198e2cbdbac39b184a78f6f90806/fixedpoint/fixedpoint_sse.h#L359-L368)
- [skia (2D graphics library)](https://github.com/google/skia/blob/52a4379f03f7cd4e1c67eb69a756abc5838a658f/src/core/SkFixed15.h#L67-L69)
- [WebRTC library](https://webrtc.googlesource.com/src/+/92ea95e34af5966555903026f45164afbd7e2088/modules/audio_coding/codecs/isac/fix/source/lattice_neon.c#55)

===========================

- **y = i16x8.q15mulr_sat_s(a, b)** is lowered to
  - `VPMULHRSW xmm_y, xmm_a, xmm_b`
  - `VPCMPEQW xmm_tmp, xmm_y, wasm_i16x8_splat(0x8000)`
  - `VPXOR xmm_y, xmm_y, xmm_tmp`

x86/x86-64 processors with SSSE3 instruction set

- **y = i16x8.q15mulr_sat_s(a, b)** is lowered to
  - `MOVDQA xmm_y, xmm_a`
  - `MOVDQA xmm_tmp, wasm_i16x8_splat(0x8000)`
  - `PMULHRSW xmm_y, xmm_b`
  - `PCMPEQW xmm_tmp, xmm_y`
  - `PXOR xmm_y, xmm_tmp`

- **y = i16x8.q15mulr_sat_s(a, b)** (`y` is **NOT** `a` and `y` is **NOT** `b`) is lowered to
  - `MOVDQA xmm_y, xmm_a`
  - `MOVDQA xmm_tmp, xmm_a`
  - `PMULLW xmm_y, xmm_b`
  - `PMULHW xmm_tmp, xmm_b`
  - `PSRLW xmm_y, 14`
  - `PADDW xmm_tmp, xmm_tmp`
  - `PAVGW xmm_y, wasm_i16x8_splat(0)`
  - `PADDW xmm_y, xmm_tmp`
  - `MOVDQA xmm_tmp, wasm_i16x8_splat(0x8000)`
  - `PCMPEQW xmm_tmp, xmm_y`
  - `PXOR xmm_y, xmm_tmp`

ARM64 processors

- **y = i16x8.q15mulr_sat_s(a, b)** is lowered to
  - `SQRDMULH Vy.8H, Va.8H, Vb.8H `

- **y = i16x8.q15mulr_sat_s(a, b)** is lowered to
  - `VQRDMULH.S16 Qy, Qa, Qb`

### DIFF
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -509,6 +509,19 @@ def S.sub_sat_u(a, b):
     return S.lanewise_binary(subsat, S.AsUnsigned(a), S.AsUnsigned(b))
 ```
 
+### Saturating integer Q-format rounding multiplication
+
+* `i16x8.q15mulr_sat_s(a: v128, b: v128) -> v128`
+
+Lane-wise saturating rounding multiplication in Q15 format:
+
+```python
+def S.q15mulr_sat_s(a, b):
+    def subq15mulr(x, y):
+        return S.SignedSaturate((x * y + 0x4000) >> 15)
+    return S.lanewise_binary(subsat, S.AsSigned(a), S.AsSigned(b))
+```
+
 ### Lane-wise integer minimum
 * `i8x16.min_s(a: v128, b: v128) -> v128`
 * `i8x16.min_u(a: v128, b: v128) -> v128`
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -161,7 +161,7 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
 | `i16x8.min_u`               |    `0x97`| -                  |
 | `i16x8.max_s`               |    `0x98`| -                  |
 | `i16x8.max_u`               |    `0x99`| -                  |
-| `i16x8.avgr_u`              |    `0x9b`|                    |
+| `i16x8.avgr_u`              |    `0x9b`| -                  |
 | `i32x4.abs`                 |    `0xa0`| -                  |
 | `i32x4.neg`                 |    `0xa1`| -                  |
 | `i32x4.any_true`            |    `0xa2`| -                  |
@@ -237,3 +237,4 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
 | `i64x2.extmul_high_i32x4_s` |   `0x119`| -                  |
 | `i64x2.extmul_low_i32x4_u`  |   `0x11a`| -                  |
 | `i64x2.extmul_high_i32x4_u` |   `0x11b`| -                  |
+| `i16x8.q15mulr_sat_s`       |     `TBD`| -                  |
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -130,6 +130,7 @@
 | `i16x8.max_s`               |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i16x8.max_u`               |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i16x8.avgr_u`              |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
+| `i16x8.q15mulr_sat_s`       |                           |                    |                    |                    |                    |
 | `i32x4.abs`                 |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.neg`                 |               `-msimd128` | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: |
 | `i32x4.any_true`            |               `-msimd128` | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: |


ngzhian:
What is the alternative without this instruction? The asymmetry doesn't look great here (5 v.s. 1 instruction).

Maratyszcza:
@ngzhian With AVX it is just 3 instructions. Emulation in the current WebAssembly SIMD specification is even more expensive:

```c
v128_t wasm_i16x8_q15mulr_sat(v128_t a, v128_t b) {
  v128_t lo = wasm_i32x4_mul(wasm_i32x4_widen_low_i16x8(a), wasm_i32x4_widen_low_i16x8(b));
  v128_t hi = wasm_i32x4_mul(wasm_i32x4_widen_high_i16x8(a), wasm_i32x4_widen_high_i16x8(b));
  const v128_t inc = wasm_i32x4_splat(0x4000);
  lo = wasm_i32x4_add(lo, inc);
  hi = wasm_i32x4_add(hi, inc);
  lo = wasm_i32x4_shr(lo, 15);
  hi = wasm_i32x4_shr(hi, 15);
  return wasm_i16x8_narrow_i32x4(lo, hi);
}
```

ngzhian:
Sure, I'm looking at V8's SIMD baseline though, which is SSE4.1, so that's 5.
Yea agree the emulation looks bad.

I'm not familiar with this Q format, do you expect other instructions from this "class" to be useful? This suggestion is different from the rest in that it is just a single instruction.

Think this is worth prototyping just to see the improvements, it also variety of use cases (across ML, codecs). I'll get started on this.

Maratyszcza:
Other instructions are just regular integer saturated addition/subtraction. Q31 multiplication is useful, but it is natively supported only on ARM, and emulation on x86 would be very expensive.

penzn:
The asymmetry is likely to be worse than 5 vs 1, as the `wasm_i16x8_splat(0x8000)` values would need to be emitted as well.

Maratyszcza:
@penzn `wasm_i16x8_splat(0x8000)` is just an in-memory constant literal. V8 currently doesn't use in-memory constants for `v128` literal values, but it is a temporary limitation, and thus shouldn't be a factor when considering instructions.

penzn:
I am not sure if this would be worth it even if the functionality did exist. In this case `mov` would be from memory, which is not exactly cheap, and would lead to pref drops when the value is not in the cache.

As for the current state of things, I don't think any engines maintain lists of `v128` constants (@lars-t-hansen correct me if I am wrong), which realistically means that for the time being the instruction count on x86 would be much worse than listed here.

I think we need to keep implementation reality in mind, at the very least because we depend on implementations in order to move the proposal forward.

lars-t-hansen:
> As for the current state of things, I don't think any engines maintain lists of `v128` constants (@lars-t-hansen correct me if I am wrong), which realistically means that for the time being the instruction count on x86 would be much worse than listed here.

Not 100% sure what you're thinking about here but our optimizing compiler will constant-fold `wasm_i16x8_splat(0x8000)` into a loadable constant, and may merge multiple occurences of it to produce a reusable constant in a register under suitable conditions.  I confess I've not put a lot of effort into such low-level optimizations of SIMD code yet, and it's unlikely that we'll get to the point where we're trying to determine whether it's best to put the constant in a register or to load it from memory for each use; the placement decision will likely be driven by register pressure and ad-hoc heuristics in the instruction selector (given current architecture).

ngzhian:
Prototyped on arm64 as of https://chromium-review.googlesource.com/c/v8/v8/+/2438990, should see it in canary tomorrow.

penzn:
Not to pick on this PR in particular, as this is a common trend lately, but how many of these projects are porting or have been ported to WebAssembly:

* [libvpx (video codec)](https://github.com/webmproject/libvpx/blob/a5d499e16570d00d5e1348b1c7977ced7af3670f/vpx_dsp/x86/inv_txfm_ssse3.h#L47-L52)
* [AV1 Codec Library (video codec)](https://github.com/mozilla/aom/blob/c8b38b0bfd36306f1886375d98512d792ffe9517/aom_dsp/x86/aom_subpixel_8t_intrin_ssse3.c#L797-L798)
* [OpenVINO (machine learning toolkit)](https://github.com/openvinotoolkit/openvino/blob/efad27d68ca88a107f9980b7dd8f07556a9eaf03/inference-engine/src/preprocessing/cpu_x86_sse42/ie_preprocess_gapi_kernels_sse42.cpp#L106-L109)
* [gemmlowp (fixed-point matrix multiplication)](https://github.com/google/gemmlowp/blob/023c190f9341198e2cbdbac39b184a78f6f90806/fixedpoint/fixedpoint_sse.h#L359-L368)
* [skia (2D graphics library)](https://github.com/google/skia/blob/52a4379f03f7cd4e1c67eb69a756abc5838a658f/src/core/SkFixed15.h#L67-L69)
* [WebRTC library](https://webrtc.googlesource.com/src/+/92ea95e34af5966555903026f45164afbd7e2088/modules/audio_coding/codecs/isac/fix/source/lattice_neon.c#55)

(this a copy of the list in PR description)

ngzhian:
Skia has a Wasm port (https://skia.org/user/modules/canvaskit) that seems to be actively maintained (the most serious one.)

The rest of the projects turn up some experimental ports (such as https://github.com/GoogleChromeLabs/webm-wasm), not sure if anything serious (happy to be corrected.)

Good question though Petr, we should focus on use cases that are being worked on.

tlively:
Various folks are [building](https://stackoverflow.com/questions/62116705/libvpx-to-wasm-facing-performance-issue) libvpx to Wasm and it looks like it is [part](https://github.com/webrtc/apprtc/tree/wartc/src/web_app/wasm/libvpx) of some official WebRTC code 

The [dav1d](https://brionv.com/log/2019/02/17/ogv-js-experimental-av1-decoding/) AVI codec builds to WebAssembly.

Here's a [blog post](https://www.edgeimpulse.com/blog/audio-based-shower-timer-with-a-phone-machine-learning-and-webassembly/) about someone using gemmlowp with Emscripten to make a shower timer, of all things 🚿

Didn't find anything in particular for OpenVINO, but this is still more than I expected to find.

Maratyszcza:
IMO the list of applications is not to show which apps will use the new instruction as soon as it is available, but rather to show that instruction is useful for different applications. Practically speaking, Emscripten includes emulation of SSE and NEON intrinsics, so most codebases written with these SIMD intrinsics can be ported with little effort.

penzn:
@ngzhian, @tlively, good list, thank you! Especially the shower timer 🙄 I am (predictably) curious if any of those use cases can be used to test Wasm SIMD vs Wasm in context of this PR. Seems like Skia can be a good example.

> The [dav1d](https://brionv.com/log/2019/02/17/ogv-js-experimental-av1-decoding/) AVI codec builds to WebAssembly.

That seems to be a different AV1 codec though than the one in the description.

> IMO the list of applications is not to show which apps will use the new instruction as soon as it is available, but rather to show that instruction is useful for different applications.

I don't think this is completely fair - if we are to delay the proposal longer, we need to understand how code in the wild would benefit from proposed operations. In a sense, consider this a re-phrasing of @binji's question in https://github.com/WebAssembly/simd/issues/343#issuecomment-696280956

> Practically speaking, Emscripten includes emulation of SSE and NEON intrinsics, so most codebases written with these SIMD intrinsics can be ported with little effort.

Only if we consider compiling those codes as the end goal. Hitting a particular operation does not guarantee performance, due to various architectural quirks (see discussion on `hadd` for example). I really think further changes to the proposal should be tested, because (a) it has been available to users for some time and (b) we are venturing into either controversial territory (`hadd`, `popcount`), or starting to re-work things in stable state for some time (`ne`, `eq`).

Again, this is not to pick on this particular operation, but to illustrate the point of having real-world tests.

tlively:
This instruction has landed in both LLVM and Binaryen and should be ready to benchmark in tip-of-tree Emscripten in a few hours. The builtin function to use is `__builtin_wasm_q15mulr_saturate_s_i8x16`.

bjacob:
"i8x16" ? or "i16x8" ?

tlively:
Oh haha I committed it with the wrong name. I'll push a quick fix right now. Thanks for the catch!

tlively:
The name has now been fixed to `__builtin_wasm_q15mulr_saturate_s_i16x8`.

sbc100:
> The name has now been fixed to `__builtin_wasm_q15mulr_saturate_s_i16x8`.

We are starting to get almost java-like in the length of our identifiers here!

bjacob:
nah, java would be AbstractBuiltinWasmQ15MulrSaturateSI16x8ProducerBuilderFactory

Maratyszcza:
I evaluated performance impact of the proposed instruction by porting [fixed-point headers in gemmlowp](https://github.com/google/gemmlowp/tree/master/fixedpoint) to WebAssembly SIMD and benchmarking [fixed-point (16-bit) sigmoid implementation from gemmlowp](https://github.com/google/gemmlowp/blob/2483d846ad865dd4190fe4a1a1ba2d9cfcea78e1/fixedpoint/fixedpoint.h#L883-L898) (this implementation is being used in TensorFlow Lite). Benchmark results are presented below:

Processor (Device) | Performance with WAsm SIMD + `i16x8.q15mulr_sat_s` | Performance with WAsm SIMD (baseline) | Speedup |
-- | -- | -- | --
Snapdragon 855 (LG G8 ThinQ) | 893 MB/s | 244 MB/s | 3.7X |
Snapdragon 670 (Pixel 3a) | 482 MB/s | 140 MB/s | 3.4X |
Exynos 8895 (Galaxy S8) | 681 MB/s | 171 MB/s | 4.0X |

ngzhian:
How large is the benchmark? Is it a small benchmark exercising just the sigmoid implementation (single function), or is it some end-to-end inference benchmark in TF Lite?

Maratyszcza:
The benchmark is for a single operator (fixed-point Sigmoid). Here's the function being benchmarked:

```c++
void Sigmoid(const int16_t* input_ptr, int16_t* output_ptr, size_t elements) {
  assert(elements % 16 == 0);

  // F0 uses 0 integer bits, range [-1, 1].
  // This is the return type of math functions such as tanh, logistic,
  // whose range is in [-1, 1].
  using F0 = gemmlowp::FixedPoint<gemmlowp::int16x8_v128_t, 0>;
  // F3 uses 3 integer bits, range [-8, 8], the input range expected here.
  using F3 = gemmlowp::FixedPoint<gemmlowp::int16x8_v128_t, 3>;

  do {
    F3 input0 =
        F3::FromRaw(gemmlowp::to_int16x8_v128_t(wasm_v128_load(input_ptr)));
    F3 input1 =
        F3::FromRaw(gemmlowp::to_int16x8_v128_t(wasm_v128_load(input_ptr + 8)));
    F0 output0 = gemmlowp::logistic(input0);
    F0 output1 = gemmlowp::logistic(input1);
    wasm_v128_store(output_ptr, output0.raw().v);
    wasm_v128_store(output_ptr + 8, output1.raw().v);

    elements -= 16;
    input_ptr += 16;
    output_ptr += 16;
  } while (elements != 0);
}
```

Maratyszcza:
The code gemmlowp changes are in the google/gemmlowp#202 PR.

penzn:
Would this change affect xnnpack as well? What would be the effect there?

> The gemmlowp code changes are in the [google/gemmlowp#202](https://github.com/google/gemmlowp/pull/202) PR.

How does the comparison work then? Looks like there is just one implementation.

Is [this](https://github.com/google/gemmlowp/blob/2483d846ad865dd4190fe4a1a1ba2d9cfcea78e1/fixedpoint/fixedpoint.h#L883-L898) the implementation on `logistic`:

```C++
// Returns logistic(x) = 1 / (1 + exp(-x)) for any x.
template <typename tRawType, int tIntegerBits>
FixedPoint<tRawType, 0> logistic(FixedPoint<tRawType, tIntegerBits> a) {
  typedef FixedPoint<tRawType, tIntegerBits> InputF;
  typedef FixedPoint<tRawType, 0> ResultF;
  tRawType mask_if_positive = MaskIfGreaterThan(a, InputF::Zero());
  tRawType mask_if_zero = MaskIfZero(a);
  InputF abs_input = SelectUsingMask(mask_if_positive, a, -a);
  ResultF result_if_positive = logistic_on_positive_values(abs_input);
  ResultF result_if_negative = ResultF::One() - result_if_positive;
  const ResultF one_half =
      GEMMLOWP_CHECKED_FIXEDPOINT_CONSTANT(ResultF, 1 << 30, 0.5);
  return SelectUsingMask(mask_if_zero, one_half,
                         SelectUsingMask(mask_if_positive, result_if_positive,
                                         result_if_negative));
}
```

Maratyszcza:
> Would this change affect xnnpack as well? What would be the effect there?

No, XNNPACK is unaffected. It might be in the future when it implements the same operator.
 
> How does the comparison work then? Looks like there is just one implementation.

Replace [this line](https://github.com/google/gemmlowp/blob/8c3530b787b308a7387f9f98ce39ab791bf1ab86/fixedpoint/fixedpoint_wasmsimd.h#L345) with `#if 1` to enable the `i16x8.q15mulr_sat_s` instruction.

> Is [this](https://github.com/google/gemmlowp/blob/2483d846ad865dd4190fe4a1a1ba2d9cfcea78e1/fixedpoint/fixedpoint.h#L883-L898) the implementation on `logistic`:

This is the entry point. Note that it calls other inline functions, e.g. `logistic_on_positive_values`.

omnisip:
Discussed in #402 (12/22/2020 Sync Meeting) -- https://docs.google.com/document/d/1Tnf-fvRcCVj_vv8CjAN_qUvj4Wa57JnPa8ghGi4_C-s/edit# -- Currently awaiting x64 prototypes (@ngzhian) to do x64 benchmarks.

Maratyszcza:
Here are the additional results on x86-64 systems:

Processor | Performance with WAsm SIMD + `i16x8.q15mulr_sat_s` | Performance with WAsm SIMD (baseline) | Speedup |
-- | -- | -- | --
Intel Xeon W-2135 | 1218 MB/s | 430 MB/s | 2.8X |
Intel Celeron N3060 | 276 MB/s | 108 MB/s | 2.6X |
AMD PRO A10-8700B | 679 MB/s | 268 MB/s | 2.5X |

proposals/simd/NewOpcodes.md
```diff
@@ -108,6 +108,7 @@
 | i8x16.max_u          | 0x79   | i16x8.max_u              | 0x99   | i32x4.max_u              | 0xb9   | ----        | 0xd9   |
 | ----------------     | 0x7a   | ----------------         | 0x9a   | i32x4.dot_i16x8_s        | 0xba   | ----        | 0xda   |
 | i8x16.avgr_u         | 0x7b   | i16x8.avgr_u             | 0x9b   | ---- avgr_u ----         | 0xbb   | ----        | 0xdb   |
+| ----                 | 0x7c   | i16x8.q15mulr_sat_s      | 0x9c   | ----                     | 0xbc   | ----        | 0xdc   |
```
tlively, line 111: Can you add TBD here, too, so that the two opcode files are consistent?

Maratyszcza: Reverted the modification of this file

tlively (approved):

dtig (approved):

proposals/simd/SIMD.md
```diff
@@ -509,6 +509,19 @@ def S.sub_sat_u(a, b):
...
+
+* `i16x8.q15mulr_sat_s(a: v128, b: v128) -> v128`
+
+Lane-wise saturating rounding multiplication in Q15 format:
+
+```python
+def S.q15mulr_sat_s(a, b):
+    def subq15mulr(x, y):
+        return S.SignedSaturate((x * y + 0x4000) >> 15)
+    return S.lanewise_binary(subsat, S.AsSigned(a), S.AsSigned(b))
```
zjiaz, line 522: Hi, is this `subsat` should be `subq15mulr`?

Maratyszcza: Right

dtig: Oops, didn't catch that during review - fixed in #424.

PR #366, opened 2020-09-28
Add SIMD instructions to index of instructions

ngzhian:
Update the table generation script and then run it to generate the index.
This depends on the execution semantics (for all the references.)


rossberg (approved):

ngzhian:
Thanks for the review, I'll merge this in since it's ready, but the links on the index will be broken since #340 is not in yet (which is fine since it doesn't break the build and no one is reading it anyway)

PR #368, opened 2020-10-01
i64x2.bitmask instruction

Maratyszcza:
Introduction
=========

This is proposal to add new variant of existing `bitmask` instruction. The new variant extracts the highest bit of the two 64-bit lanes in a SIMD vector into an 32-bit integer. This variant was left out of #201 without any discussion (maybe @zeux knows why), but would be useful both for orthogonality of the instruction set and for efficiency: x86 natively supports this instruction since SSE2, and on ARM is can be emulated more efficiently than other `bitmask` variants.

Applications
=========

- [RTM (RealTime Math) library](https://github.com/nfrechette/rtm/blob/60e61b4fa37b01817e931164a30b21150b260354/includes/rtm/mask4d.h#L133)
- [Havok physics library](https://github.com/resigma/havok-2013/blob/54bc2e0334c526469781fc1186a94f5d850eded4/Common/Base/Math/Vector/Sse/hkSseVector4Comparison_D.inl#L149-L150)
- [Hierarchical Equal Area isoLatitude Pixelation (HEALPix) library](https://sourceforge.net/p/healpix/code/HEAD/tree/trunk/src/cxx/cxxsupport/sse_utils_cxx.h#l369)

===========================

- **i64x2.bitmask**
  - `y = i64x2.bitmask(x)` is lowered to `VMOVMSKPD reg_y, xmm_x`

- **i64x2.bitmask**
  - `y = i64x2.bitmask(x)` is lowered to `MOVMSKPD reg_y, xmm_x`

ARM64 processors

- **i64x2.bitmask**
  - `y = i64x2.bitmask(x)` is lowered to:
    - `SQXTN Vtmp.2S, Vx.2D`
    - `USHR Vtmp.2S, Vtmp.2S, 31`
    - `USRA Dtmp, Dtmp, 31`
    - `FMOV Wy, Stmp`

- **i64x2.bitmask**
  - `y = i64x2.bitmask(x)` is lowered to:
    - `VQMOVN.S64 Dtmp, Qx`
    - `VSHR.U32 Dtmp, Dtmp, 31`
    - `VSRA.U64 Dtmp, Dtmp, 31`
    - `VMOV.32 Ry, Dtmp[0]`

### DIFF
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -183,6 +183,7 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
 | `i32x4.max_u`               |    `0xb9`| -                  |
 | `i32x4.dot_i16x8_s`         |    `0xba`| -                  |
 | `i64x2.neg`                 |    `0xc1`| -                  |
+| `i64x2.bitmask`             |    `0xc4`| -                  |
 | `i64x2.shl`                 |    `0xcb`| -                  |
 | `i64x2.shr_s`               |    `0xcc`| -                  |
 | `i64x2.shr_u`               |    `0xcd`| -                  |
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -151,6 +151,7 @@
 | `i32x4.max_u`               |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.dot_i16x8_s`         |                           | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i64x2.neg`                 |               `-msimd128` | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: |
+| `i64x2.bitmask`             |                           |                    |                    |                    |                    |
 | `i64x2.shl`                 |               `-msimd128` | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: |
 | `i64x2.shr_s`               |               `-msimd128` | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: |
 | `i64x2.shr_u`               |               `-msimd128` | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: |
--- a/proposals/simd/NewOpcodes.md
+++ b/proposals/simd/NewOpcodes.md
@@ -78,36 +78,36 @@
 | v128.xor       | 0x51   |
 | v128.bitselect | 0x52   |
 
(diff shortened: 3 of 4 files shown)


zeux:
I think this might be an accidental omission - I didn't actually know that SSE2 supports this natively, and I didn't have use cases for 64-bit masks. So it didn't occur to me to propose that because I thought SSE2 lowering would have to scalarize.

tlively:
This is implemented in LLVM (but not Binaryen) as `__builtin_wasm_bitmask_i64x2` and will be available in Emscripten in a few hours. I don't know if we need benchmarking for this instruction or not, though.

akirilov-arm:
I'd like to propose an alternative Arm64 mapping:
```
ushr Vtmp.2d, Vx.2d, #63
mov Xy, Vtmp.d[0]
mov Xtmp, Vtmp.d[1]
add Wy, Wy, Wtmp, lsl #1
```
The main advantage of this sequence is that the middle 2 instructions are independent, so they can execute in parallel. In fact, the essentially scalarized version might execute even faster (and require only 1 temporary register), but is 1 instruction longer:
```
mov Xy, Vx.d[0]
mov Xtmp, Vx.d[1]
lsr Xy, Xy, #63
lsr Xtmp, Xtmp, #63
add Wy, Wy, Wtmp, lsl #1
```

ngzhian:
Forgot to say that this is prototyped in v8 (x64) https://chromium.googlesource.com/v8/v8/+/ceee7cfe7260152fd90c66657b8476b9d3a8b915

ngzhian:
@akirilov-arm would you suggest a similar mapping for ARMv7 as well?

akirilov-arm:
@ngzhian Something like this (`tmp2` = `tmp * 2`, `tmp3` = `tmp * 2 + 1`):
```
vshrq.u64 Qtmp, Qx, #63
vmov.32 Ry, Dtmp2[0]
vmov.32 Rtmp, Dtmp3[0]
add Ry, Ry, Rtmp, lsl #1
```
Note that I haven't tested the sequence and I am also not sure about its performance characteristics - extra latency may crop up due to the SIMD & FP register overlapping rules in AArch32.

ngzhian:
Prototyped on arm64 as well

proposals/simd/BinarySIMD.md
```diff
@@ -30,6 +30,7 @@ In the description below, `ImmLaneIdx{I}` indicates the maximum value of the byt
 For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
 
 
+<<<<<<< HEAD
```
ngzhian, line 33: merge marker

Maratyszcza: Fixed

tlively (approved):

dtig (approved):

PR #370, opened 2020-10-05
Clarify out-of-range behaviour of i8x16.swizzle

Maratyszcza:
Per @omnisip, the current wording is confusing as it can be understood as using lane #0 for the output.


tlively (approved):
Thanks!

penzn (approved):

Issue #373, opened 2020-10-06
v128.load32_zero and v128.load64_zero instructions extensions to #237


omnisip:
# Introduction
@Maratyszcza has done a wonderful job describing the use cases and functionality of load64_zero and load32_zero in #237.  This proposal seeks to extend the functionality of load64_zero and load32_zero to be functionally complete with the underlying architecture by adding support for its sister variants with identical implementations.  This would add support from other 32-bit and 64-bit registers and from the low 32 and 64 bits of other vectors.  The proposed instructions are  load32_zero_r, load64_zero_r, load32_zero_v, and load64_zero_v respectively. Since these are sister instructions, the applications, use cases, and instructions are identical to the original proposal.  This ticket will serve as a placeholder for the upcoming PR and will be updated in tandem.

# Mapping to Common Instruction Sets
This section illustrates how the new WebAssembly instructions can be lowered on common instruction sets. However, these patterns are provided only for convenience, compliant WebAssembly implementations do not have to follow the same code generation patterns.

### v128.load64_zero_v

penzn:
Memory operations can only do either of the following:

- Load: pop a memory index off the stack, get the value from specified memory location and push it onto the stack
- Store: pop a value and an index off the stack, put the value into specified memory location

omnisip:
Is this a labeling issue?

penzn:
Setting a lane is done using [replace lane](https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#replace-lane-value) operations. Please try to understand the spec before proposing changes to it, and if you have questions - please ask, we'll be happy to answer.

omnisip:
@penzn This isn't the same as replace_lane.  Replace lane replaces one value and returns the updated vector.  This initializes a vector from a scalar or another vector zeroing the upper parts.

Maratyszcza:
Looks reasonable. I suggest @omnisip open a Pull Request with the proposed changes to the specification, because is it more actionable for V8/SpiderMonkey/LLVM devs.

penzn:
Do we expect this operation to be "on critical path", and if so, what kind of gains are we going to get by going from two wasm instructions (neither of which touches memory) to one?

omnisip:
@Maratyszcza Thanks for the feedback.  If anyone can help with the ARMv7 with Neon intrinsics before I'll generate the PR today.

[...]

PR #374, opened 2020-10-07
move{32,64}_zero_{r,v} instructions

omnisip:
# Introduction
@Maratyszcza has done a wonderful job describing the use cases and functionality of load64_zero and load32_zero in #237.  This proposal seeks to extend the functionality of load64_zero and load32_zero to be functionally complete with the underlying architecture by adding support for its sister variants with identical implementations.  This would add support from other 32-bit and 64-bit registers and from the low 32 and 64 bits of other vectors.  The proposed instructions are  move32_zero_r, move64_zero_r, move32_zero_v, and move64_zero_v respectively. Since these are sister instructions, the applications, use cases, and instructions are identical to the original proposal. 

#### v = v128.move32_zero_v(v128) is lowered to 
```
        vxorps  xmm1, xmm1, xmm1
        vblendps        xmm0, xmm1, xmm0, 1             # xmm0 = xmm0[0],xmm1[1,2,3]
```
#### v = v128.move32_zero_v(v128) is lowered to 
```
        xorps   xmm1, xmm1
        movss   xmm1, xmm0                      # xmm1 = xmm0[0],xmm1[1,2,3]
        movaps  xmm0, xmm1
```
#### v = v128.move32_zero_r(r32) is lowered to
```
        movi    v0.2d, #0000000000000000
        mov     v0.s[0], w0
```
#### v = v128.move32_zero_r(r64) is lowered to
```
        movi    v0.2d, #0000000000000000
        mov     v0.d[0], x0
```
#### v = v128.move32_zero_v(v128) is lowered to
```
        movi    v1.2d, #0000000000000000
        mov     v1.s[0], v0.s[0]
        mov     v0.16b, v1.16b
```
### v128.move64_zero_v(v128) is lowered to
```
        movi    v0.2d, #0000000000000000
        mov     v0.d[0], x0
```

### DIFF
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -222,3 +222,7 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
 | `i32x4.trunc_sat_f32x4_u`  |    `0xf9`| -                  |
 | `f32x4.convert_i32x4_s`    |    `0xfa`| -                  |
 | `f32x4.convert_i32x4_u`    |    `0xfb`| -                  |
+| `v128.move32_zero_r`       |    `0xfe`| -                  |
+| `v128.move32_zero_v`       |    `0xff`| -                  |
+| `v128.move64_zero_r`       |   `0x100`| -                  |
+| `v128.move64_zero_v`       |   `0x101`| -                  | 
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -190,6 +190,10 @@
 | `i32x4.trunc_sat_f32x4_u`  |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `f32x4.convert_i32x4_s`    |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `f32x4.convert_i32x4_u`    |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
+| `v128.move32_zero_r`       |                           |                                         |                    |                    |
+| `v128.move32_zero_v`       |                           |                                         |                    |                    |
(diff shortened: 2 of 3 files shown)


omnisip:
Copied from #373 for easy reading:

This proposal was originally put together for completeness along with the load32/load64_zero. Its functionality is equivalent to and 0xffffffffffffff for the lower 64 bits with movq and and 0xffffffff for the lower 32 bits with movd. Since there is an efficient alternative (albeit register using alternative), I don't think this should hold up any standardization effort. If someone has a need for this to be included before we finalize the instruction set, please comment here.

dtig:
@omnisip As there have been no further comments, would you object to this PR being closed?

Maratyszcza:
The lowering of these instructions looks good, and if there are important use cases it makes sense to expose the underlying native instructions to WAsm SIMD. However, I personally don't have a use-case for these instructions.

dtig:
Agree that the lowering on different architectures is good - I read this PR as a call for use cases, and as there have been none so far pinging this thread to see if @omnisip has reasons we should include this. We can also punt this to post-MVP, but I'd be inclined to close if there are no concrete use cases.

omnisip:
If we include these, we would probably only include move32_zero_r and move64_zero_r as top level instructions.   Move32_zero_r and move64_zero_r are special in that they make it possible to initialize a vector from a register without splatting the value.  This is something that doesn't exist in our current instruction set.

Maratyszcza:
@omnisip We need use-cases where this is useful, and they are not obvious: e.g. AFAIK NEON intrinsics on ARM don't expose this instruction.

omnisip:
Yeah.  That's what I've been running across in my research.  When I looked for the intrinsic versions of the moves on x64, I found that in most cases, they were immediately followed by a broadcast or a shuffle to replicate the same behavior as dup.  If ARM doesn't expose these, I don't see any pertinent reason to keep this PR open -- I just wanted to check loose ends before we closed this out.

dtig:
Closing as per #436.

PR #377, opened 2020-10-07
Move simd operations out into a new file

ngzhian:


interpreter/exec/eval.ml
```diff
@@ -282,27 +282,27 @@ let rec step (c : config) : config =
         with exn -> vs', [Trapping (numeric_error e.at exn) @@ e.at])
 
       | Ternary ternop, v3 :: v2 :: v1 :: vs' ->
-        (try Eval_numeric.eval_ternop ternop v1 v2 v3 :: vs', []
+        (try Eval_simd.eval_ternop ternop v1 v2 v3 :: vs', []
```
rossberg, line 285: Since this actually is a SIMD-specific, maybe rename to SimdTernary? If just named Ternary, it should be handled analogously to Binary.

rossberg (approved):
LGTM modulo comment

interpreter/exec/eval.ml
ngzhian: Yea good point, renamed to SimdTernary (otherwise we will need to create new ternop for all the other numeric types, which contain nothing...). Thanks!

PR #381, opened 2020-10-09
i64x2.eq instruction

Maratyszcza:
Introduction
=========

This is proposal to add 64-bit variant of existing `eq` instruction. ARM64 and x86 (since SSE4.1) natively support this instruction, and on ARMv7 NEON and SSE2 is can be efficiently emulated with 3-4 instructions.

Applications
=========

- [Boost.Atomic library](https://github.com/boostorg/atomic/blob/4b2edae1a7dcbd2be1c088ddc05acb0634e69a15/src/find_address_sse41.cpp#L55-L62)
- [SUPERCOP cryptography library](https://github.com/jedisct1/supercop/blob/a4203f1f35c3dac9b96a9ae297f13e22151428e0/crypto_dh/gls254/opt/smu.h#L236-L251)
- [Flang compiler runtime library](https://github.com/flang-compiler/flang/blob/a21454d4a0e27e31743864e389a66b1fd95fbac0/runtime/libpgmath/lib/common/pow/fma3/sdpow.cpp#L115)
- [BearSSL cryptography library](https://bearssl.org/gitweb/?p=BearSSL;a=blob;f=src/symcipher/aes_x86ni_ctrcbc.c;h=f57fead6858a1b7de132a6276d627c9c2d0e6713;hb=HEAD#l84)
- [Plato framework for distributed graph computations and machine learning](https://github.com/Tencent/plato/blob/28fd58ffcbfa8a9e1365a90371eccf010429b3f0/plato/util/intersection_impl.hpp#L215-L216)
- [BESS (Berkeley Extensible Software Switch) framework](https://github.com/NetSys/bess/blob/ae52fc5804290fc3116daf2aef52226fafcedf5d/core/packet_avx.h#L87-L89)
- [LightMatrix library for matrix computation](https://github.com/lindahua/light-matrix/blob/8ce65d14b6027ee9e081dd12460bc714e9131d11/light_mat/simd/internal/sse_fpclass_impl.h#L75)

===========================

- **i64x2.eq**
  - `y = i64x2.eq(a, b)` is lowered to `VPCMPEQQ xmm_y, xmm_a, xmm_b`

- **i64x2.eq**
  - `y = i64x2.eq(a, b)` is lowered to `MOVDQA xmm_y, xmm_a` + `PCMPEQQ xmm_y, xmm_b`

- **i64x2.eq**
  - `y = i64x2.eq(a, b)` is lowered to:
    - `MOVDQA xmm_y, xmm_a`
    - `PCMPEQD xmm_y, xmm_b`
    - `PSHUFD xmm_tmp, xmm_y, 0xB1`
    - `PAND xmm_y, xmm_tmp`

ARM64 processors
- **i64x2.eq**
  - `y = i64x2.eq(a, b)` is lowered to `CMEQ Vy.2D, Va.2D, Vb.2D`

- **i64x2.eq**
  - `y = i64x2.eq(a, b)` is lowered to:
    - `VCEQ.I32 Qy, Qa, Qb`
    - `VREV64.32 Qtmp, Qy`
    - `VAND Qy, Qtmp`

### DIFF
--- a/proposals/simd/BinarySIMD.md
+++ b/proposals/simd/BinarySIMD.md
@@ -249,3 +249,4 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
 | `v128.store16_lane`         |     `TBD`| m:memarg, i:ImmLaneIdx8  |
 | `v128.store32_lane`         |     `TBD`| m:memarg, i:ImmLaneIdx4  |
 | `v128.store64_lane`         |     `TBD`| m:memarg, i:ImmLaneIdx2  |
+| `i64x2.eq`                  |     `TBD`| -                        |
--- a/proposals/simd/ImplementationStatus.md
+++ b/proposals/simd/ImplementationStatus.md
@@ -148,6 +148,7 @@
 | `i32x4.max_s`               |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.max_u`               |               `-msimd128` | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
 | `i32x4.dot_i16x8_s`         |                           | :heavy_check_mark: |                    |                    | :heavy_check_mark: |
+| `i64x2.eq`                  |                           |                    |                    |                    |                    |
 | `i64x2.neg`                 |               `-msimd128` | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: |
 | `i64x2.bitmask`             |                           | :heavy_check_mark: |                    |                    |                    |
 | `i64x2.shl`                 |               `-msimd128` | :heavy_check_mark: |                    | :heavy_check_mark: | :heavy_check_mark: |
--- a/proposals/simd/SIMD.md
+++ b/proposals/simd/SIMD.md
@@ -717,6 +717,7 @@ in each lane are `0` for `false` and all ones for `true`.
 * `i8x16.eq(a: v128, b: v128) -> v128`
 * `i16x8.eq(a: v128, b: v128) -> v128`
 * `i32x4.eq(a: v128, b: v128) -> v128`
+* `i64x2.eq(a: v128, b: v128) -> v128`
 * `f32x4.eq(a: v128, b: v128) -> v128`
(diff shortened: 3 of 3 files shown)


ngzhian:
Any reason why we specifically want only i64x2.eq?
In #101 we decided that the set of i64x2 instructions we would keep did not include any of the comparisons.

Maratyszcza:
@ngzhian Of course, I'd rather have a full set of compare instructions, but ordered comparisons are hard to emulate in lieu of hardware support. On the other side, emulating 64-bit compare is trivial, and it is in our baseline ISAs (SSE4.1 and ARM64 NEON).

ngzhian:
It looks incomplete that we only have i64x2.eq, and no other i64x2 comparisons.

How useful will only adding this instruction be? Are there use cases where adding this instruction is sufficient to unlock?

Maratyszcza:
I don't have any use-cases in mind, just trying to orthogonalize the instruction set.

lemaitre:
I have no code to present, but a use case for that is when vectorizing code that mix doubles and integers: in order to limit the number of shuffles (going back and forth 32-bit elements), one would use 64-bit integers.
There, it would be nice to have a `i64x2.eq`.

tlively:
This has been prototyped in LLVM (but not Binaryen) as `__builtin_wasm_eq_i64x2`. It should be usable from tot Emscripten in a few hours as long as you don't use optimization flags at link time.

omnisip:
> Any reason why we specifically want only i64x2.eq?
> In #101 we decided that the set of i64x2 instructions we would keep did not include any of the comparisons.
Ditto on the same question.  When posted to [Stackoverflow regarding pcmpgtq](https://stackoverflow.com/questions/65166174/how-to-simulate-pcmpgtq-on-sse2-and-cmgt-on-armv7-with-neon/65175746#65175746), a response was provided that produced a high-quality result for both SSE2 as well as ARMv7+Neon.

Maratyszcza:
Added examples of applications

abrown:
I actually think this would be nice to add if it didn't have orthogonality implications. Could we just merge this without `i64x2.ne` and the `i64x2` comparisons?

dtig (approved):
Approved for merge as of #419.

PR #391, opened 2020-11-02
Implement encode, decode for load_zero

ngzhian:


ngzhian:
Missed that the build was failing because encode/decode not implemented for these new instructions.

rossberg (approved):

PR #394, opened 2020-11-04
Update SpiderMonkey implementation status

lars-t-hansen:
Dot and LoadZero landed a while back, should be in FF83.


ngzhian:
LGTM, feel free to merge once CI is done :)

PR #399, opened 2020-12-01
Update ChakraCore status

penzn:


tlively (approved):
Is the footnote at the bottom about ChakraCore still correct?

penzn:
Good catch - I missed that.

PR #406, opened 2020-12-17
Fix ext mul opcodes in BinarySIMD.md

ngzhian:


tlively (approved):
I really hope that’s not permanent, but probably good to have the docs reflect reality.

ngzhian:
> I really hope that’s not permanent

Wdym?

proposals/simd/BinarySIMD.md
```diff
@@ -225,15 +225,15 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
...
-| `i32x4.extmul_high_i16x8_s` |   `0x115`| -                  |
-| `i32x4.extmul_low_i16x8_u`  |   `0x116`| -                  |
-| `i32x4.extmul_high_i16x8_u` |   `0x117`| -                  |
-| `i64x2.extmul_low_i32x4_s`  |   `0x118`| -                  |
-| `i64x2.extmul_high_i32x4_s` |   `0x119`| -                  |
-| `i64x2.extmul_low_i32x4_u`  |   `0x11a`| -                  |
-| `i64x2.extmul_high_i32x4_u` |   `0x11b`| -                  |
+| `i16x8.extmul_low_i8x16_s`  |    `0x9a`| -                  |
+| `i16x8.extmul_high_i8x16_s` |    `0x9d`| -                  |
+| `i16x8.extmul_low_i8x16_u`  |    `0x9d`| -                  |
```
tlively, line 230: Oops, just noticed this repetition.

tlively:
It would be nice to have consistently contiguous opcodes for these instructions eventually. My understanding is that the current opcodes were necessary to prototype them quickly without increasing opcode size, so right now I would be sad if these were the final opcodes.

Maratyszcza:
It seems that no other WAsm engine implements these instructions yet, but it wouldn't be for long, so suggest you change the opcodes now or leave them as is. Later renumbering is always a pain.

proposals/simd/BinarySIMD.md
ngzhian: Good catch, fixed.

lars-t-hansen:
> It seems that no other WAsm engine implements these instructions yet, but it wouldn't be for long, so suggest you change the opcodes now or leave them as is. Later renumbering is always a pain.

Yes please.  Either renumber ASAP or freeze until we do a final renumbering (which IMO we should not, but I lost the discussion last time and I expect to lose it again...)

tlively:
The only instructions we've merged to the proposal that weren't accounted for in the last renumbering are the load zero instructions, the [floating point rounding instructions](https://github.com/WebAssembly/simd/commit/8e87db70fe237937fa10d735cc2a2aa8c851f476#diff-e9b9432641b95561a3a6e1559d172e1e10000e3ec6b4421668db6202722d3200R191-R198), and these extending multiply instructions. The load zero instructions and the extending multiply instructions were just added at the end of the opcode space, so I don't think there's a pressing reason to change them. The floating point rounding instructions were placed in the space reserved for i64x2 operations including max_s, max_u, dot_i32x4_s, and avgr_u. As long as we don't plan on introducing any of those i64x2 instructions, those can probably stay where they are as well.

I think the best move would be to update the extending multiply instruction implementations to match the current spec proposal (i.e. have them use ops 0x110-0x11b).  For any future instructions that we merge to the proposal, we should just add them at the end rather than putting them in holes reserved for other instructions. I think the thing blocking this plan is that V8 does not support opcodes above 0xff right now. @ngzhian how hard would that be to support?

ngzhian:
> I'm hoping that within a short period of time we make a decision as to which operations will make it into the final proposal and do the best we can with numbering. before merging the ones currently being prototyped.

I agree with Deepti. We have a number of outstanding proposals, especially i64x2 ones. I think we should figure out if we want those or not, then worry about renumbering.
- if we don't include them, we can likely fit everything into 0xff opcodes, so this won't be a problem
- if we include them, we will exceed 0xff, and V8 will have to support it

> how hard would that be to support?
This is a cross cutting change, it wouldn't be hard, but will be many lines changed.

tlively:
My understanding of the proposed plan is to first figure out all the additional instructions we intend to add to the instruction set, then to figure out how best number them. No instructions that have already been merged at this point (except possibly the extending multiply instructions) will have their numbers changed. @ngzhian @dtig is that along the lines of what you were thinking?

dtig:
@tlively Yes, that is what I was thinking. As we are close to the end here, I like the idea of evaluating all the outstanding instructions and make a final decision - it makes the boundary between prototype/officially supported opcodes clear as well. As @ngzhian mentioned, it's not hard to support but we would rather not if it's not required.

ngzhian:
Deprecating this, we will fix all opcodes after #452 .

PR #409, opened 2020-12-21
Fix typo

00ff0000red:
Actually, it's not even a typo, it's just a random extra word that made no sense based on its placement.


ngzhian:
LGTM, thanks!

PR #434, opened 2021-01-28
Run simd test cases with make partest

ngzhian:
We list all simd test cases in simd/* in TESTFILES, then tweak the
substitution in quiettest to strip quiettest/ from the front of the
target, rather than match on just the file name (which loses the simd/)
prefix.

The temp output file name is unchanged, it uses the base file name (no
simd/ prefix).

### DIFF
--- a/interpreter/Makefile
+++ b/interpreter/Makefile
@@ -113,7 +113,7 @@ $(WINMAKE):	clean
 # Executing test suite
 
 TESTDIR =	../test/core
-TESTFILES =	$(shell cd $(TESTDIR); ls *.wast)
+TESTFILES =	$(shell cd $(TESTDIR); ls *.wast; ls simd/*.wast)
 TESTS =		$(TESTFILES:%.wast=%)
 
 .PHONY:		test debugtest partest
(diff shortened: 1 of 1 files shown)


interpreter/Makefile
```diff
@@ -113,7 +113,7 @@ $(WINMAKE):	clean
 # Executing test suite
 
 TESTDIR =	../test/core
-TESTFILES =	$(shell cd $(TESTDIR); ls *.wast)
+TESTFILES =	$(shell cd $(TESTDIR); ls *.wast; ls simd/*.wast)
```
rossberg, line 116:
```suggestion
TESTFILES =	$(shell cd $(TESTDIR); ls *.wast; ls */*.wast)
```
(so that it works generically for potential future subdirs)

rossberg (approved):

ngzhian:
Just realized something incorrect about this, if you run `make partest` twice locally, the first run generates `test/core/_output`, and this global will run the tests inside that temporary folder as well.
Uploaded #449 to fix this.

PR #435, opened 2021-01-29
[interpreter] Implement store lane instructions

ngzhian:
v128.store8_lane
v128.store16_lane
v128.store32_lane
v128.store64_lane

Introduce a new ast type, SimdStoreLane.

The exact binary opcodes for these instructions are not yet fixed, I've
used the ones currently implement in V8 and LLVM/Binaryen, we can change
those later.

Also added a new test generation script, and the generated test files.


interpreter/text/lexer.mll
```diff
@@ -314,6 +314,14 @@ rule token = parse
   { SIMD_LOAD_LANE (fun a o i -> (v128_load32_lane (opt a 2)) o i) }
   | "v128.load64_lane"
   { SIMD_LOAD_LANE (fun a o i -> (v128_load64_lane (opt a 3)) o i) }
+  | "v128.store8_lane"
+  { SIMD_STORE_LANE (fun a o i -> (v128_store8_lane (opt a 0)) o i) }
```
rossberg, line 318: Nit: just noticed that the indentation of this and all previous semantic actions is off.

interpreter/valid/valid.ml
```diff
@@ -323,6 +323,14 @@ let rec check_instr (c : context) (e : instr) (s : infer_stack_type) : op_type =
     check_memop c memop (fun _ -> None) e.at;
     [I32Type; memop.ty] --> []
 
+  | SimdStoreLane (memop, i) ->
+    check_memop c memop (fun o -> o) e.at;
+    (match memop.sz with
+    | Some pack_size ->
+      require (i < 16 / packed_size pack_size) e.at "invalid lane index";
+      [I32Type; V128Type] -->  []
+    | _ -> assert false)
```
rossberg, line 332: Maybe simplify (and similarly SimdLoadLane):
```suggestion
    let sz = Lib.Option.get memop.sz Pack8 in
    require (i < 16 / packed_size sz) e.at "invalid lane index";
    [I32Type; V128Type] -->  []
```

interpreter/runtime/memory.mli
```diff
@@ -43,3 +43,6 @@ val load_simd_lane :
 val store_packed :
   pack_size -> memory -> address -> offset -> value -> unit
     (* raises Type, Bounds *)
+val store_simd_lane :
```
rossberg, line 46: I wonder if it wouldn't make more sense to remove these lane functions from the Memory module and refactor slightly: turn the splicing part into functions in the Simd module and perform the memory access separately in Eval via the existing load/store_packed functions.

ngzhian: store_simd_lane is: extract_lane + store_value/store_packed.
extract_lane (and correspondingly replace_lane for load_simd_lane) is already defined in Simd module.
We could remove these lane functions from Memory, it would probably look like inlining what is already in store_simd_lane, since the bulk of the work is deciding which extract_lane functions to call for various pack_size.
Or instead of inlining, we define a small helper in Simd, to perform this mapping from pack_size -> extract_lane functions.

rossberg: Yes, the latter is more or less what I had in mind. That could go into the Simd module.

ngzhian: Happy a bit of difficulty defining the helper.
The extract lane functions have different signatures that aren't compatible:
`i32x4.extract_lane` returns a int32, `i64x2.extract_lane` returns a int64.
And I can't wrap those in a `value` due to circular dependency (`Values` depends on `Simd` already).
The best I can do right now is to inline the logic inside of eval.ml.

rossberg: Then I'd inline it in eval.ml. It's only used once anyway, isn't?

ngzhian: Yup, only used once, I inlined them into SimdStoreLane, ptal.

rossberg: For consistency, can you do the same for simd_load_lane?

ngzhian: Done!

rossberg (approved):
Thanks!

ngzhian:
Rebasing to fix merge conflicts.

PR #438, opened 2021-02-02
[interpreter] Implement SIMD extended multiply instructions

ngzhian:
These were accepted into the proposal in #376.

There are 12 instructions in total:

- i16x8.extmul_{low,high}_i8x16_{s,u}
- i32x4.extmul_{low,high}_i16x8_{s,u}
- i64x2.extmul_{low,high}_i32x4_{s,u}

The implementation is straightforward, widen (using existing
operations), then a multiply with the wider shape.

The binary opcodes are not decided yet, they currently follow the ones
used in V8, when those are finalized, we can change it to match.

Added a test generation script that reuses some logic in the generator
for arithmetic instructions. Since these instructions have different
src and dst shapes, I tweaked the base class to allow for having
different shapes.


interpreter/text/lexer.mll
```diff
@@ -578,6 +578,19 @@ rule token = parse
   | "i32x4.dot_i16x8_s"
   { BINARY i32x4_dot_i16x8_s }
 
+  | "i16x8.extmul_low_i8x16_"(sign as s)
+  { BINARY (ext s i16x8_extmul_low_i8x16_s i16x8_extmul_low_i8x16_u) }
```
rossberg, line 582: Nit: indentation

rossberg (approved):

PR #451, opened 2021-02-05
[interpreter] Implement i8x16.popcnt

ngzhian:
Small fix to the test generation script, the order of applying operators
was incorrect.

Includes a fix to popcnt implementation that is upstream
(https://github.com/WebAssembly/spec/pull/1286).


rossberg (approved):

ngzhian:
Added decode and encode, force pushed because I rebased. Will merge after CI is green.

PR #452, opened 2021-02-06
Final opcodes

tlively:
A draft numbering of the most recently added instructions. I tried to keep related instructions together, so there are still holes in the opcode space. I also could have more aggressively filled holes at the expense of destroying much of the internal structure of the opcode space. There are 236 instructions, so in principle we could fit them all below `0x100`, but it would be ugly. Comments welcome :)


proposals/simd/NewOpcodes.md
```diff
@@ -100,14 +112,26 @@
 | i8x16.sub            | 0x71   | i16x8.sub                | 0x91   | i32x4.sub                | 0xb1   | i64x2.sub                | 0xd1   |
 | i8x16.sub_sat_s      | 0x72   | i16x8.sub_sat_s          | 0x92   | ---- sub_sat ----        | 0xb2   | -------------            | 0xd2   |
 | i8x16.sub_sat_u      | 0x73   | i16x8.sub_sat_u          | 0x93   | ---- sub_sat ----        | 0xb3   | -------------            | 0xd3   |
-| -------------        | 0x74   | -------------            | 0x94   | -------------            | 0xb4   | -------------            | 0xd4   |
+|                      | 0x74   | -------------            | 0x94   | -------------            | 0xb4   | -------------            | 0xd4   |
```
lars-t-hansen, line 115: Spurious change from dashes to blank?

tlively: Indeed, fixed!

Maratyszcza:
Why are `0xe2` and `0xee` reserved for `round`?

Maratyszcza:
IMO we should recycle at least the opcodes where the original reserved instruction is meaningless. E.g. `0x67-0x6a` opcodes are meaningless for widen instructions. Also, AFAICT it is missing `v128.all_true`.

proposals/simd/NewOpcodes.md
```diff
@@ -100,14 +112,26 @@
...
-| i8x16.max_u          | 0x79   | i16x8.max_u              | 0x99   | i32x4.max_u              | 0xb9   | -------------            | 0xd9   |
-| ----------------     | 0x7a   | ----------------         | 0x9a   | i32x4.dot_i16x8_s        | 0xba   | -------------            | 0xda   |
-| i8x16.avgr_u         | 0x7b   | i16x8.avgr_u             | 0x9b   | ---- avgr_u ----         | 0xbb   | -------------            | 0xdb   |
+| i8x16.max_s          | 0x78   | i16x8.max_s              | 0x98   | i32x4.max_s              | 0xb8   | -------------            | -      |
+| i8x16.max_u          | 0x79   | i16x8.max_u              | 0x99   | i32x4.max_u              | 0xb9   | -------------            | -      |
+| ----------------     | 0x7a   | ----------------         | 0x9a   | i32x4.dot_i16x8_s        | 0xba   | -------------            | -      |
+| i8x16.avgr_u         | 0x7b   | i16x8.avgr_u             | 0x9b   | ---- avgr_u ----         | 0xbb   | -------------            | -      |
+| ----------------     | 0x7c   | i16x8.q15mulr_sat_s      | 0x9c   | ----------------         | 0xbc   | -------------            | -      |
```
ngzhian, line 123: this could go into the same row as i8x16.popcnt, since they are the only once-off operation of its lane shape.
Then, if you choose some different opcodes for f32x4.ceil to f64x2.nearest, you could fit all the ext mul inside
0x7c-0x7f, 0x9c-0x9f, 0xbc-0xbf, 0xdc-0xdf.
e.g. https://github.com/ngzhian/simd/blob/another-opcode-map/proposals/simd/NewOpcodes.md

tlively:
@Maratyszcza I moved `v128.any_true` to 0x53 because it's now a one-off instruction, but if that renumbering would be disruptive, I could move it back to where it was.

proposals/simd/BinarySIMD.md
```diff
@@ -228,43 +228,43 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
...
+| `v128.store32_lane`             |    `0x5a`| m:memarg, i:ImmLaneIdx4  |
+| `v128.store64_lane`             |    `0x5b`| m:memarg, i:ImmLaneIdx2  |
+| `i64x2.eq`                      |   `0x102`| -                        |
+| `i64x2.ne`                      |   `0x103`| -                        |
+| `i64x2.lt_s`                    |   `0x104`| -                        |
+| `i64x2.gt_s`                    |   `0x105`| -                        |
+| `i64x2.le_s`                    |   `0x106`| -                        |
+| `i64x2.ge_s`                    |   `0x107`| -                        |
+| `i64x2.all_true`                |    `0xc3`| -                        |
+| `f64x2.convert_low_i32x4_s`     |    `0xfc`| -                        |
```
yurydelendik, line 260: `fc` and `fd` are taken by `v128.load32_zero` and `v128.load64_zero`

tlively: Thanks for catching that!

tlively:
Ok, I pushed a new version based on @ngzhian's suggestion and compacting more instructions into previous opcode holes. This version only uses opcodes up to 0xff, which will simplify V8 at least. The only block of instructions that really got scattered were the `f64x2` rounding instructions, which are scattered among the integer operations.

proposals/simd/NewOpcodes.md
```diff
@@ -12,8 +12,6 @@
 | v128.load32_splat  | 0x09   |
 | v128.load64_splat  | 0x0a   |
 | v128.store         | 0x0b   |
```
ngzhian, line 14: Looks like we have 0x0c-0x0f (4 opcodes) free now, we could move `f64x2.{ceil,floor,trunc,nearest}` here so they are not as scattered?

proposals/simd/NewOpcodes.md
```diff
@@ -124,9 +145,13 @@
...
-| i32x4.trunc_sat_f32x4_u | 0xf9   |
-| f32x4.convert_i32x4_s   | 0xfa   |
-| f32x4.convert_i32x4_u   | 0xfb   |
+| Conversion Op                | opcode |
+| ---------------------------- | ------ |
+| i32x4.trunc_sat_f32x4_s      | 0xf8   |
+| i32x4.trunc_sat_f32x4_u      | 0xf9   |
+| f32x4.convert_i32x4_s        | 0xfa   |
+| f32x4.convert_i32x4_u        | 0xfb   |
+| f64x2.convert_low_i32x4_s    | 0xfc   |
```
ngzhian, line 154: Should we flip f64x2.convert with i32x4.trunc_sat, to match the ordering we have above? trunc_sat first, then convert. No big deal though.

ngzhian:
Amazing bin packing work Thomas, thanks!

proposals/simd/NewOpcodes.md
tlively: No, 0x0c-0x0f are occupied byt const, shuffle, swizzle, and a splat. Were you looking somewhere else?

ngzhian: Whoops, misread, forgot to expand the collapsed diffs.

proposals/simd/BinarySIMD.md
```diff
@@ -226,45 +226,45 @@ For example, `ImmLaneIdx16` is a byte with values in the range 0-15 (inclusive).
...
+| `i16x8.extmul_high_i8x16_u`     |    `0x9f`| -                        |
+| `i32x4.extmul_low_i16x8_s`      |    `0xbc`| -                        |
+| `i32x4.extmul_high_i16x8_s`     |    `0xbd`| -                        |
+| `i32x4.extmul_low_i16x8_u`      |    `0xbe`| -                        |
+| `i32x4.extmul_high_i16x8_u`     |    `0xbf`| -                        |
+| `i64x2.extmul_low_i32x4_s`      |    `0xdc`| -                        |
+| `i64x2.extmul_high_i32x4_s`     |    `0xdd`| -                        |
+| `i64x2.extmul_low_i32x4_u`      |    `0xde`| -                        |
+| `i64x2.extmul_high_i32x4_u`     |    `0xdf`| -                        |
+| `i16x8.q15mulr_sat_s`           |    `0x9c`| -                        |
```
yurydelendik, line 243: I believe it is 0x82 (i16x8.extmul_low_i8x16_s uses 0x9c)

tlively:
> Why are `0xe2` and `0xee` reserved for `round`?

TBH I don't remember. I must have thought at the time of the last renumbering that there would be just one floating point rounding instruction for each of f32x4 and f64x2.

proposals/simd/BinarySIMD.md
tlively: Thanks, fixed.

zeux:
256 opcodes must be enough for anyone 🚀

proposals/simd/NewOpcodes.md
```diff
@@ -77,37 +75,60 @@
...
+| f64x2.trunc                   | 0x7a   |                           | 0x9a   | i32x4.dot_i16x8_s         | 0xba   | i64x2.le_s                | 0xda   |
+| i8x16.avgr_u                  | 0x7b   | i16x8.avgr_u              | 0x9b   | ---- avgr_u ----          | 0xbb   | i64x2.ge_s                | 0xdb   |
+| i16x8.extadd_pairwise_i8x16_s | 0x7c   | i16x8.extmul_low_i8x16_s  | 0x9c   | i32x4.extmul_low_i16x8_s  | 0xbc   | i64x2.extmul_low_i32x4_s  | 0xdc   |
+| i16x8.extadd_pairwise_i8x16_u | 0x7d   | i16x8.extmul_high_i8x16_s | 0x9d   | i32x4.extmul_high_i16x8_s | 0xbd   | i64x2.extmul_high_i32x4_s | 0xdd   |
+| i32x4.extadd_pairwise_i16x8_s | 0x7e   | i16x8.extmul_low_i8x16_u  | 0x9e   | i32x4.extmul_low_i16x8_u  | 0xbe   | i64x2.extmul_low_i32x4_u  | 0xde   |
+| i32x4.extadd_pairwise_i16x8_u | 0x7f   | i16x8.extmul_high_i8x16_u | 0x9f   | i32x4.extmul_high_i16x8_u | 0xbf   | i64x2.extmul_high_i32x4_u | 0xdf   |
 
 | f32x4 Op        | opcode | f64x2 Op        | opcode |
 | --------------- | ------ | --------------- | ------ |
```
ngzhian, line 134: (I can't seem to add a comment to the line I want, so commenting here instead).
It is a bit arbitrary, but we could put f32x4.ceil and f64x2.ceil here, then we can fit f64x2.{floor,trunc,nearest} into `0xd2-0xd4`, that might make the LLVM opcode definitions a bit simpler? Since the 64x2 opcodes will all be a constant number of opcodes away from 32x4.

ngzhian (approved):

proposals/simd/NewOpcodes.md
tlively: Where exactly were you thinking about putting the `ceil`s? When I was splitting up these operations, I was trying to optimize for filling holes below 0x80 to save code size. For a group of just a few related operations like this, the difference in simplicity in the LLVM definitions is very minor.

ngzhian: inside 0xe2, 0xee. Oh okay, then optimizing for < 0x80 sg, thanks!

tlively:
Merging because there were no further comments on this at today's meeting.

PR #455, opened 2021-02-09
[spectext] Add i64x2 eq,ne,lt_s,gt_s,le_s,ge_s.

ngzhian:
Previously it was included \ishape.\virelop, but it is incorrect, as we
don't have i64x2 unsigned comparisons.

i64x2 unsigned comparisons were added in #412.

### DIFF
--- a/document/core/syntax/instructions.rst
+++ b/document/core/syntax/instructions.rst
@@ -221,7 +221,15 @@ SIMD instructions provide basic operations over :ref:`values <syntax-value>` of 
      \K{i64x2.}\EXTRACTLANE~\laneidx \\&&|&
      \fshape\K{.}\EXTRACTLANE~\laneidx \\&&|&
      \shape\K{.}\REPLACELANE~\laneidx \\&&|&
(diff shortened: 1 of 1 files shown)


rossberg (approved):
Oh dear, the arbitrariness in the instruction matrix is starting to drive me insane...

rossberg:
Hm, I wonder if some of the irregularity wouldn't be more readable with grammar side conditions that blacklist the exceptions, instead of listing all allowed cases.

ngzhian:
> Oh dear, the arbitrariness in the instruction matrix is starting to drive me insane...

Yea, the lowerings for i64x2 unsigned comparisons is not good enough to meet acceptance criteria, and is only better on hardware that isn't widely available yet.

> Hm, I wonder if some of the irregularity wouldn't be more readable with grammar side conditions that blacklist the exceptions, instead of listing all allowed cases.

I can try to come up with some restrictions on the grammar, not sure how I would phrase it. Let me finish adding all the instructions to syntax then we can explore this, I'll file a TODO on myself.

PR #383, opened 2020-10-13
Double-precision conversion instructions

Maratyszcza:
Introduction
=========

As discussed in #348, WebAssembly SIMD specification is functionally incomplete w.r.t double-precision computations as it doesn't cover many basic C/C++ operations, particularly conversions between `double` type and other types. Currently any code that use these operations would be de-vectorized (see [examples](https://github.com/emscripten-core/emscripten/blob/40d5db0cd0de1ac49ed959e6554f26ab52601155/system/include/SSE/emmintrin.h#L365-L369) in `emmintrin.h` header in Emscripten). Double-precision computations are extensively used in scientific computing, and providing a complete set of double-precision is a must to bring scientific computing applications to the Web. In the foreseeable future WebAssembly SIMD will be the only avenue for high-performance scientific computing in the Web, as neither WebGL nor WebGPU support double-precision computations.

This PR introduce the minimal set of conversion operations to fill this gap.

Applications
=========

- Conversions from double to int are used when storing data in quantized representation and in table-based algorithms (need to real indices to int index for table lookup).
- Conversions from int to double are used when loading data in quantized representation and in random number generation (the raw output of RNGs is typically int, but numerically algorithms need real values).
- Conversions between double and float are used in mixed-precision algorithms, where less sensitive parts of the application are computed in single precision and the more sensitive parts - in double precision. E.g. iterative algorithms may first run in single precision to convergence, and then starting with the single-precision solution run double precision iterations.

Use-cases in OpenCV
=========

- [`f32x4.demote_f64x2_zero`](https://github.com/opencv/opencv/blob/8f653ba8decc76f896e4a5e578302c8443eaacc7/modules/core/include/opencv2/core/hal/intrin_wasm.hpp#L2455-L2465)
- [`f64x2.promote_low_f32x4`](https://github.com/opencv/opencv/blob/8f653ba8decc76f896e4a5e578302c8443eaacc7/modules/core/include/opencv2/core/hal/intrin_wasm.hpp#L2510-L2518)
- [`f64x2.convert_low_i32x4_s`](https://github.com/opencv/opencv/blob/8f653ba8decc76f896e4a5e578302c8443eaacc7/modules/core/include/opencv2/core/hal/intrin_wasm.hpp#L2480-L2493)

Mapping to Common Instruction Sets
===========================

This section illustrates how the new WebAssembly instructions can be lowered on common instruction sets. However, these patterns are provided only for convenience, compliant WebAssembly implementations do not have to follow the same code generation patterns.

[...]

PR #466, opened 2021-02-12
[interpreter] Implement i32x4.trunc_sat_f64x2_{s,u}_zero

ngzhian:
This converts 2 f64 to 2 i32, then zeroes the top 2 lanes.

These 2 instructions were merged as part of #383.

This change also refactors some test cases from simd_conversions out
into a script that generates both i32x4.trunc_sat_i32x4_{s,u} and
i32x4.trunc_sat_f64x2_{s,u}_zero.


interpreter/exec/simd.ml
```diff
@@ -444,6 +446,10 @@ struct
     let trunc_sat_f32x4_s = convert I32_convert.trunc_sat_f32_s
     let trunc_sat_f32x4_u = convert I32_convert.trunc_sat_f32_u
 
+    let convert_zero f v = Rep.of_i32x4 (I32.zero :: I32.zero :: (List.map f (Rep.to_f64x2 v)))
```
rossberg, line 449:
```suggestion
    let convert_zero f v =
      Rep.(of_i32x4 I32.(zero :: zero :: (List.map f (to_f64x2 v)))
```

rossberg (approved):

PR #478, opened 2021-02-19
[spectext] Rename integer widen instructions to integer extend

### LINKED ISSUE #467: Rename integer-to-integer widening instructions to "extend"?
Referring to this section https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#integer-to-integer-widening, was there a reason why we chose "widen"? That isn't a terminology used in the current Wasm spec, it uses "extend" (e.g. see https://webassembly.github.io/spec/core/appendix/index-instructions.html).

Can we change the names to extend? (Everything else stays the same, only `s/widen/extend/`.

ngzhian:
Changes to interpreter will come in a follow-up patch.

Fixed #467.


tlively (approved):
Thanks!

rossberg (approved):

PR #476, opened 2021-02-19
[spectext] Add extended pairwise add instructions

ngzhian:
This was merged in #380.

Drive-by formatting fix for header.

PR #482, opened 2021-02-23
[spectext] Add double precision conversion

ngzhian:
Instructions were added in #383.

Consolidate conversion operations (vcvtop) more, merging int-int
widening operations.

Drive-by fix extmul definition in syntax (shouldn't include the shape).


rossberg (approved):
Nice!

ngzhian:
Rebased and force push to fix merge conflict. index-instructions.rst has some extadd instructions from #476 because I didn't properly add the generated index before merging.

PR #486, opened 2021-02-26
Update interpreter and text with finalized opcodes

ngzhian:
Opcodes were finalized in #452, this updates the interpreter and spec
text to use these final opcodes.

I noticed some opcodes were missing from binary/instructions (floating-point rounding) and add them in.


rossberg (approved):
(Rubberstamp)

PR #489, opened 2021-03-05
Fix a typo in the f64x2.{pmin,pmax} binary spec

alexcrichton:
It looks like these have a typo in copying them over from
`BinarySIMD.md`


ngzhian:
Thanks!

PR #491, opened 2021-03-10
[interpreter] Small clean up in eval_simd unop

ngzhian:


ngzhian:
Abandoning since after the merge from upstream this will not apply cleanly, will rebase and upload another change.

PR #493, opened 2021-03-10
Comparison instructions should return -1 in lanes

### LINKED ISSUE #441: Specification for comparison instructions should return -1 (all ones) and not 1
Comparisons should set a lane to 0 if false or all ones if true (https://github.com/WebAssembly/simd/blob/master/proposals/simd/SIMD.md#comparisons).

For `{i8x16,i16x8,i32x4}.{eq,ne...}`, the specification seems to rely on http://localhost:9000/exec/numerics.html#xref-exec-numerics-op-ieq-mathrm-ieq-n-i-1-i-2 which sets a lane to `1` or `0`.

E.g.:
i8x16.eq gets matched by `shape.vbinop`, which uses the rule `vbinop_shape`, which is `eq_i8x16` which applies `eq_i8` lanewise.

Not sure how to solve this yet, filing a bug to track this. We might need a special rule to match comparisons and do a "signed extension" step after the comparisons. Or define a new set of comparisons that return `-1`.

Note that only the text is wrong, the interpreter implementation doing the right thing.

ngzhian:
SIMD comparisons return -1 (all bits set) in the lanes. Currently as
specified, they return 1 (since we use the numeric comparison). We fix
this by extracting these instructions (virelop and vfrelop) into vrelop,
and in the execution, sign extend them from i1 to iX (which treats the
1-bit integer "1" as -1).

Drive-by fix for exec-vbinop, some typos in the execution semantics.

Fixed #441.

### DIFF
--- a/document/core/exec/instructions.rst
+++ b/document/core/exec/instructions.rst
@@ -494,7 +494,7 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
 :math:`\shape\K{.}\vbinop`
 ..........................
 
-1. Assert: due to :ref:`validation <valid-vbinop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+1. Assert: due to :ref:`validation <valid-vbinop>`, two values of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
 
 2. Pop the value :math:`\V128.\VCONST~c_2` from the stack.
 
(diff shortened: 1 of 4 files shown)


document/core/exec/instructions.rst
```diff
@@ -494,7 +494,7 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
 :math:`\shape\K{.}\vbinop`
 ..........................
 
-1. Assert: due to :ref:`validation <valid-vbinop>`, a value of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
+1. Assert: due to :ref:`validation <valid-vbinop>`, two values of :ref:`value type <syntax-valtype>` |V128| is on the top of the stack.
```
rossberg, line 497:
```suggestion
1. Assert: due to :ref:`validation <valid-vbinop>`, two values of :ref:`value type <syntax-valtype>` |V128| are on the top of the stack.
```

document/core/exec/instructions.rst
```diff
@@ -512,13 +512,44 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
...
+7. Push the value :math:`\V128.\VCONST~c` to the stack.
+
+.. math::
+   \begin{array}{l}
+   \begin{array}{lcl@{\qquad}l}
+   (\V128\K{.}\VCONST~c_1)~(\V128\K{.}\VCONST~c_2)~t\K{x}N\K{.}\vrelop &\stepto& (\V128\K{.}\VCONST~c)
+   \end{array}
+   \\ \qquad
+     \begin{array}[t]{@{}r@{~}l@{}}
+     (\iff c = \lanes^{-1}_{t\K{x}N}(\extend^x_{i1,t}(\vrelop_t(\lanes_{t\K{x}N}(c_1), \lanes_{t\K{x}N}(c_2)))))
```
rossberg, line 548: You probably meant \i1, but i fact I think that \extend only takes widths, not types, so you don't need it. Also, you probably meant to say ^s:
```suggestion
     (\iff c = \lanes^{-1}_{t\K{x}N}(\extend^{\K{s}}_{1,|t|}(\vrelop_t(\lanes_{t\K{x}N}(c_1), \lanes_{t\K{x}N}(c_2)))))
```

document/core/exec/instructions.rst
```diff
@@ -512,13 +512,44 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
...
+
+2. Pop the value :math:`\V128.\VCONST~c_2` from the stack.
+
+3. Pop the value :math:`\V128.\VCONST~c_1` from the stack.
+
+4. Let :math:`i^\ast` be the sequence :math:`\lanes_{t\K{x}N}(c_1)`.
+
+5. Let :math:`j^\ast` be the sequence :math:`\lanes_{t\K{x}N}(c_2)`.
+
+6. Let :math:`c` be the result of computing :math:`\lanes^{-1}_{t\K{x}N}(\extend^s_{i1,t}(\vrelop_t(i^\ast, j^\ast)))`.
```
rossberg, line 537: See below.

document/core/exec/instructions.rst
ngzhian: Yup, \i1 and ^s, good catch. \extend takes type, there is an automatic conversion from type to widths (https://webassembly.github.io/spec/core/exec/instructions.html#numeric-instructions).

document/core/exec/instructions.rst
```diff
@@ -534,7 +534,7 @@ SIMD instructions are defined in terms of generic numeric operators applied lane
 
 5. Let :math:`j^\ast` be the sequence :math:`\lanes_{t\K{x}N}(c_2)`.
 
-6. Let :math:`c` be the result of computing :math:`\lanes^{-1}_{t\K{x}N}(\extend^s_{i1,t}(\vrelop_t(i^\ast, j^\ast)))`.
+6. Let :math:`c` be the result of computing :math:`\lanes^{-1}_{t\K{x}N}(\extend^s_{\i1,t}(\vrelop_t(i^\ast, j^\ast)))`.
```
rossberg, line 537: Nit: there are \extendu and \extends macros that typeset the s/u correctly.

I'd still suggest to got to the bit-width version directly. The "automatic conversion" is just intended as a way to define the actual instructions of that name. Here it's just an unnecessary step. (Accordingly, the aforementioned macros point to the base definition.)

ngzhian: Oh yes I forgot about those. Fixed according to your suggestion, and removed \i1 since it's not longer used anywhere, thanks!

rossberg (approved):

PR #495, opened 2021-03-12
Merge from upstream spec

ngzhian:
Quite a number of changes from upstream because ref-types was merged.
Most of the fixes have to do with num_type v.s. value_type, wrapping up V128 into Num, or pattern matching a Num V128.

Moved some code from Eval_numeric into Values to avoid circular imports:

- eval_numeric imports eval_simd
- both needs the I32Num and TypeError

For the spec text, I put SIMD 128 sections after the reference types section, e.g. in validation.


rossberg (approved):
Mostly rubber-stamping. After having spent an entire day merging into various down-stream proposal repos myself, I can imagine that it wasn't much fun. :)

I did note various exceptions and oddities produced by lumping v128 under num_type, however. I wonder if it wouldn't be cleaner (in both spec and interpreter) if Simd was treated as a third kind of value type (perhaps introducing a new name for the union of the two). From first glance, there seem to be way more places where their handling differs than where it is the same. The impossible cases in the encoder are a good indicator that the current factorisation isn't ideal (ideally, the AST should not have any malformed values, though we already have a couple of small exceptions).

But we can revisit that later.

rossberg:
Btw, I'll be on vacation for the rest of the month, so expect some delay in code reviews.

ngzhian:
> I wonder if it wouldn't be cleaner (in both spec and interpreter) if Simd was treated as a third kind of value type (perhaps introducing a new name for the union of the two).

Hm, not too sure if what the end result will look like if we do that. I will file a TODO on myself to think about this a bit more and maybe do some experiments, thanks for the suggestion! Merging as it is for now.

PR #497, opened 2021-03-18
Small cleanup to num_pat to remove redundant Source.phrase

ngzhian:
Took the chance to clean up the matching logic for SimdResult too.

### DIFF
--- a/interpreter/script/js.ml
+++ b/interpreter/script/js.ml
@@ -288,7 +288,7 @@ let assert_return ress ts at =
       | ArithmeticNan -> canonical_nan_of (* can be any NaN that's one everywhere the canonical NaN is one *)
     in
     match res.it with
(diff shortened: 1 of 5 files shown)


rossberg (commented):
Hm, would it make sense to first make this change upstream, since it is mostly unrelated to SIMD?

ngzhian:
I think this wouldn't work upstream, because the [upstream definition]:(https://github.com/WebAssembly/spec/blob/master/interpreter/script/script.ml#L25)

```
type result = result' Source.phrase
and result' =
  | LitResult of literal
  | NanResult of nanop
  | RefResult of Types.ref_type
```

`RefResult` needs the Source.phrase outside.

This PR is more to clean up the kludge I introduced when changing result to include `SimdResult`.

rossberg (approved):
I see, makes sense.

PR #501, opened 2021-03-26
Improve simd_i64x2_cmp.wast

### LINKED ISSUE #487: Improve simd_i64x2_cmp.wast
@ngzhian, I believe the [simd_i64x2_cmp.wast](https://github.com/WebAssembly/simd/blame/master/test/core/simd/simd_i64x2_cmp.wast) tests could be improved if we varied the expected results of the `ge_s` and `le_s` tests. Most of those tests expect `[-1, -1]` so when I made a mistake re: flipped operands the tests did not fail for me. When I added tests like the following then I observed the expected failure (note the expected result is not all 1s):

```
(assert_return (invoke "ge_s" (v128.const i64x2 -1 -1)
                              (v128.const i64x2 0 -1))
                              (v128.const i64x2 0 -1))
```

Should we add a few of these types of tests (for implementor sanity)? And have you been adding these by hand or using some test generator (I think there was a Python one somewhere)?

abrown:
Fixes #487 by flipping the operands of the `ge_s` and `le_s` tests. This change makes immediately clear when an x86 implementation incorrectly uses the XMM registers; without these tests, running the SIMD suite could pass with an incorrect implementation.


abrown:
cc: @ngzhian

ngzhian:
LGTM, thanks!

PR #510, opened 2021-08-02
Refactor some SIMD code

rossberg:
I apologise for the large PR! It grew organically and is a bit difficult to disentangle retroactively. Most of the changes are fairly mechanical, though. I hope it's manageable.

This PR refactors some pieces of the interpreter to separate scalar from vector types and ops more thoroughly and make them as regular as possible. That gives rise to fewer "junk" AST configurations (saving cases of `assert false`), delimits the irregularities of SIMD to SIMD-specific instructions, and will allow easier subsetting in the future. None of the changes should affect the behaviour of the interpreter.

Specifically:

* Separate V128 into simd_type.

* Separate SIMD ops and instructions into their own op types and instructions.

* Separate SIMD `v128.` ops from shape-aware ops.

Other tweaks:

* Refactor representation of packed SIMD loads/stores to closer match structure of scalar loads.

* Refactor representations and handling of assertion result patterns to be more regular.

* Refactor printing code for SIMD ops to avoid repitition.

* Reorder some code, especially group SIMD instructions together and in same order everywhere.

* Update README with SIMD constructs.

* Fix Makefile to handle test subdirs correctly.


interpreter/runtime/instance.ml
```diff
@@ -43,6 +43,13 @@ let () =
     | FuncRef _ -> "func"
     | r -> string_of_ref' r
 
+let () =
```
ngzhian, line 46: Is this an intended change?

ngzhian (approved):
LGTM, thanks for refactoring this. Small comment about something I see in instance.ml, it looks like unrelated to SIMD.

interpreter/runtime/instance.ml
rossberg: Ah, yes: it is related to the refactoring of result matching. While doing that, I noticed that the way it compared references was actually bogus, and a proper function is needed for that. I could have fixed that separately, but while I was here...

rossberg (commented):
Thanks for the quick review!

interpreter/runtime/instance.ml
ngzhian: Got it, LGTM!

PR #511, opened 2021-08-03
Fix TODOs

rossberg:
Kill the remaining todos and fixmes:

* Inline Simd functor, since it hardcoded various assumptions about the width anyway, e.g. in form of the known shapes. All types that are specific to V128 now live in that module.

* Plus a few structural simplifications to the module after that.

* Use I8 and I16 modules for string conversion.

* While at it, rename Eval_numeric to Eval_num and inline Numeric_error into Int module.

This also fixes a bug in Int.Make().to_int_u which I ran into through one of the changes. Will fix it upstream as well.


ngzhian:
Fixes, renames, refactoring LGTM, thanks!

PR #512, opened 2021-08-03
Rename Int/Float modules to Ixx/Fxx

rossberg:
This is for hygiene, since OCaml now has stdlib modules of the same names.


ngzhian (approved):
Looks great, thanks!

PR #522, opened 2021-09-28
Add Changelog

rossberg:
Add changelog for SIMD extensions, plus a couple of minor fixes/tweaks.


document/core/appendix/changes.rst
```diff
@@ -104,17 +104,52 @@ Added instructions that modify ranges of memory or table entries [#proposal-reft
 * Active data and element segments boundaries are no longer checked at compile time but may trap instead
 
 
+.. index:: instructions, SIMD, value type, vector type
+
+Vector instructions
+...................
+
+Added vector type and instructions that manipulate multiple numeric values in parallel (also known as _SIMD_, single instruction multiple data) [#proposal-vectype]_
```
ngzhian, line 112: what is `_SIMD_` supposed to type set as? I see it on the HTML as `_SIMD_`, is it supposed to underline/italics?

document/core/appendix/changes.rst
```diff
@@ -104,17 +104,52 @@ Added instructions that modify ranges of memory or table entries [#proposal-reft
...
+
+* New :ref:`value type <syntax-valtype>`: |V128|
+
+* New :ref:`memory instructions <syntax-instr-memory>`: :math:`\K{v128.}\LOAD{}\!N\!\K{x}\!M\!\K{\_}\sx`, :math:`\K{v128.}\LOAD{}N\K{\_zero}`, :math:`\K{v128.}\LOAD{}N\K{\_splat}`, :math:`\K{v128.}\LOAD{}N\K{\_lane}`, :math:`\K{v128.}\STORE{}N\K{\_lane}`
```
ngzhian, line 116: missing `v128.load` and `v128.store`.

document/core/appendix/changes.rst
```diff
@@ -104,17 +104,52 @@ Added instructions that modify ranges of memory or table entries [#proposal-reft
...
+
+* New lane :ref:`vector instructions <syntax-instr-vec>`: :math:`\K{i}\!N\!\K{x}\!M\!\K{.extract\_lane\_}\sx^?`, :math:`\K{i}\!N\!\K{x}\!M\!\K{.replace\_lane}`, :math:`\K{f}\!N\!\K{x}\!M\!\K{.extract\_lane}`, :math:`\K{f}\!N\!\K{x}\!M\!\K{.replace\_lane}`
+
+* New conversion :ref:`vector instructions <syntax-instr-vec>`: :math:`\K{i}\!N\!\K{x}\!M\!\K{.extend\_}\half\K{\_i}\!N'\!\K{x}\!M'\!\K{\_}\sx`, :math:`\K{i32x4.trunc\_sat\_f32x4\_}\sx`, :math:`\K{i32x4.trunc\_sat\_f64x2\_}\sx\K{\_zero}`, :math:`\K{f32x4.convert\_i32x4\_}\sx`, :math:`\K{f32x4.demote\_f64x2\_zero}`, :math:`\K{f64x2.convert\_low\_i32x4\_}\sx`, :math:`\K{f64x2.promote\_low\_f32x4}`
+
+* New reordering :ref:`vector instructions <syntax-instr-vec>`: :math:`\K{i8x16.shuffle}`, :math:`\K{i8x16.swizzle}`, :math:`\K{i8x16.narrow\_i16x8\_}\sx`, :math:`\K{i16x8.narrow\_i32x4\_}\sx`
```
ngzhian, line 134: not sure if narrow should be part of reordering, i think it should go into conversion (above), since it's the opposite to extend.

ngzhian (changes requested):

document/core/appendix/changes.rst
rossberg: Fixed.

document/core/appendix/changes.rst
rossberg: Added.

document/core/appendix/changes.rst
rossberg: Yeah, I wasn't sure where else to put it. It doesn't fit with the conversion operators because unlike those it is binary, like the others here. And like these, it really just moves bytes/words around, doesn't it?

Is there a better place?

ngzhian: I see, yea depends on how you want to classify it. Conversion to me is for shape 1 -> shape 2, so extend and narrows fit into conversion for me. It is not just moving bytes/words around, since we will saturate values if they overflow.

Not sure if there is a clearer classification here, and maybe we don't need to be too strict about it. We could move this (and even shuffle and swizzle) to "binary vector instructions". (Although shuffles take 16 immediates, which is different from the other binary vector instructions.)

dtig: I would also expect the narrow/extend to be in the same set of instructions. Perhaps there's space for another category of `lane width modification` instructions that are not really conversions?

document/core/appendix/changes.rst
```diff
@@ -104,17 +104,52 @@ Added instructions that modify ranges of memory or table entries [#proposal-reft
 * Active data and element segments boundaries are no longer checked at compile time but may trap instead
 
 
+.. index:: instructions, SIMD, value type, vector type
+
+Vector instructions
+...................
+
+Added vector type and instructions that manipulate multiple numeric values in parallel (also known as *SIMD*, single instruction multiple data) [#proposal-vectype]_
```
dtig, line 112: Would it make sense to add text here that's representative of the lane operations that extract and replace single lanes instead of multiple values?

document/core/appendix/changes.rst
rossberg: Fair enough. I moved extend/narrow to their own bullet and reordered bullets slightly. (I wish extend & narrow actually were inverses, but it's so much more ad-hoc than that...)

document/core/appendix/changes.rst
rossberg: Hm, there is no such breakdown in the intro sentence of the other changelog sections either. The details are in the bullet list, this is just for the general direction.

dtig: Ack.

dtig (approved):

PR #528, opened 2021-11-15
Make i8 and i16 work correctly on their own

ngzhian:
Previously, i8 and i16 was careless w.r.t. the top bits,
because they were only used by v128, which uses Bytes.get/set that
appropriate masks the top bits when storing into the byte array.

With this change, i8 and i16 are self contained small integers
implemented using int32. They are always stored signed-extended, e.g.
INT8_MIN (-128) is stored as 0xffffff80.

This requires adding sign-extension operation (Rep.sx) in a couple of
places, that will make sure to extend the sign bit to the rest of the
int32.

Also add a small, non-exhaustive test for these small integers in a new
file test/smallint.ml. This is the first ml test we are adding (all
previous tests are wasm/wast tests that we run using the interpreter),
so there are some modifications to the Makefile to build and run this
test (make smallinttest, make test will also run it).


interpreter/exec/ixx.ml
```diff
@@ -39,6 +39,9 @@ sig
   val to_int64: t -> int64
   val to_string : t -> string
   val to_hex_string : t -> string
+  (* If bit (bitwidth - 1) is set, sx will sign-extend t to maintain the
+   * invariant that small ints are stored sign-extended inside a wider int. *)
+  val sx : t -> t
```
rossberg, line 44: Is this actually needed as a parameter? I think you can implement this generically inside the module:
```
sx i = of_i64 ((to_i64 i << (64 - bitwidth)) >> (64 - bitwidth))
```

rossberg (commented):
Very cool, thanks a lot!

interpreter/exec/ixx.ml
ngzhian: You're right, not needed, but i can pass Fun.id in the i32 and i64 case which makes it a tiny bit more performant?

interpreter/exec/ixx.ml
```diff
@@ -152,11 +149,17 @@ struct
   let abs = Rep.abs
   let neg = Rep.neg
 
+  (* If bit (bitwidth - 1) is set, sx will sign-extend t to maintain the
+   * invariant that small ints are stored sign-extended inside a wider int. *)
+  let sx x =
+    let i = 64 - Rep.bitwidth in
+    Rep.of_int64 Int64.(shift_right( shift_left (Rep.to_int64 x) i) i)
```
rossberg, line 156:
```suggestion
    Rep.of_int64 Int64.(shift_right (shift_left (Rep.to_int64 x) i) i)
```

rossberg (approved):

